Issue #19, opened 2021-09-19
Mutiple axes homing in same cycle with one axis squaring stressfree mode crashes FluidNC


nxperera:
When two axes home in same cycle (e.g. X and Y both has  /homing/cycle = 2  ) if one axis is ganged(has two motors) with independent limit switches on each motor and squaring enabled,  FluidNC homing crashes  and CNC becomes unresponsive . Serial output  following 
```
Backtrace: 0x401ae337:0x3ffb1d70 0x400d8f4d:0x3ffb1d90 0x400d904b:0x3ffb1dc0 0x400dc7ff:0x3ffb1df0 0x400e3832:0x3ffb1e10 0x400e38ad:0x3ffb1e30 0x401ae749:0x3ffb1e50 0x400e4226:0x3ffb1e70 0x400e4543:0x3ffb1ef0 0x400e456c:0x3ffb1f20 0x400e551c:0x3ffb1f40 0x400dbe32:0x3ffb1f60 0x40107e04:0x3ffb1fb0 0x4008be02:0x3ffb1fd0
  #0  0x401ae337:0x3ffb1d70 in Machine::Axis::pulloffOffset() at FluidNC/src/Machine/Axis.cpp:83
  #1  0x400d8f4d:0x3ffb1d90 in Machine::Homing::run_one_cycle(unsigned short) at FluidNC/src/Machine/Homing.cpp:304 
```
My configuration example 
```
axes:
     x:
      homing:
           cycle: 2
    motor0:
      limit_pos: gpio.33
      hard_limits: false
      pulloff: 2.000

 y:
    homing:
      cycle: 2

    motor0:
      limit_pos: gpio.34
      hard_limits: false
      pulloff: 2.000

    motor1:
      limit_pos: gpio.32
      hard_limits: false
      pulloff: 2.000

```

nxperera:
This happens because X and Y both are in same homing cycle and Y is squired with independent limit switches which make FluidNC to use squaredStressfree() routine which eventually calls  following  pulloffOffset() method while homing x axis where  _motors[1] is null and that cause the crash.
 
```
float Axis::pulloffOffset() { return  _motors[1]->_pulloff - _motors[0]->_pulloff } 
```

PR #21, opened 2021-09-21
BugFix:  Mutiple axes homing in same cycle with one axis squaring stressfree mode crashes FluidNC #19

nxperera:
Added guard in Axis::pulloffOffset() method 
In Homing::run_one_cycle(..) method  added a check if the axis  hasDualMotor before calling nudge move.


bdring:
Thanks!

I changed it to merge into devt

PR #22, opened 2021-09-23
Move radio settings from the config file to $ settings

bdring:
- This is done to make it easier to setup
- Bluetooth fixed
- Added STA and STA_AP modes
- Release system changed so only wifi and bt compiles are done.


luc-github:
what is the user case to let user to en
able  AP_STA ?

bdring:
Some people are afraid to use AP mode.

luc-github:
ap_sta is also ap.mode,

luc-github:
and it is also disabled in codely enven you set  it manualy https://github.com/bdring/FluidNC/blob/0b4fbcd3fb80ff17cbae7ded40b6eb6a4e555cf9/FluidNC/src/WebUI/WifiServices.cpp#L113
so if people afraid of AP mode better move the fail mode to no wifi mode

bdring:
@luc-github  The goal is to have 3 options
.
 - **STA_AP** This is likely to be most popular and is equivalent to the old STA mode
 -  **AP** Same as old AP mode
 - **STA** This has been request by a couple of people because they do not want an AP at any time. 

```
[MSG:INFO: Connecting to STA SSID:Barts-WLAN]
[MSG:INFO: Connecting.]
[MSG:INFO: Connecting..]
[MSG:INFO: Connecting...]
[MSG:INFO: Connecting....]
[MSG:INFO: Connecting.]
[MSG:INFO: Connecting..]
[MSG:INFO: Connecting...]
[MSG:INFO: Connecting....]
[MSG:INFO: Connecting.]
[MSG:INFO: Connecting..]
[MSG:INFO: WiFi off]

Grbl 3.1 [FluidNC v3.1.2 (RadioSettings-5bee7f4) (wifi) '$' for help]
[MSG:INFO: Check limits]
[MSG:INFO: '$H'|'$X' to unlock
```

luc-github:
ok so the STA_AP mode is not actually the AP_STA mode - the working is confusing - I understand now

bdring:
Yes, the wording was chosen because it tries STA first.

PR #24, opened 2021-09-24
Backlash correction proposal

nxperera:
This change will create intermediate (secret) plan_buffer_line() call when a backlash correction is required for a given  motion and this correction motion is supposed to be hidden from the system as much as possible (this is achieved by skipping the motor_steps increment/decrement at  Stepper::pulse_func() method for backlash steps). I have considered the risk of messing up  Bresenham line algorithm by this skipping steps. Even if that's the case it is localized only to the backlash motion. 
In order to define a backlash correction use /Axes/<x,y,z,a,..>/backlash=<correction_value> in config.yaml . Apart from system motions backlash correction is applied to all the cnc motions when defined it in the configuration.  This implementation tracks the motion changes while probing and homing. Backlash correction is applied to probe moves.

### DIFF
--- a/FluidNC/src/Backlash.cpp
+++ b/FluidNC/src/Backlash.cpp
@@ -85,12 +85,8 @@ bool backlash_Compensate_befor_target(float* target, plan_line_data_t *pl_data){
 
     if(backlash_comp_needed){
         log_debug( "BS_COMP (" << backlash_comp_target[0] << " ," << backlash_comp_target[1] << " ," << backlash_comp_target[2]<< " )!!");
-        //mc_line(backlash_comp_target, pl_data_backlash);
         plan_buffer_line(backlash_comp_target, pl_data_backlash);
     }
-    // else {
-    //     log_debug("No Backlash compensation..!!");
-    // }
     return true;
 }
 
@@ -110,7 +106,7 @@ void backlash_Reset_for_homing(bool approach, uint8_t homing_mask) {
     for (int axis = 0; axis < config->_axes->_numberAxis; axis++) {
         if(config->_axes->_axis[axis]->_homing != nullptr ){
             if (bitnum_is_true(homing_mask, axis)) {
-                float t_pos = steps_to_mpos( motor_steps[axis], axis ); /// axis_settings[axis]->steps_per_mm->get();
+                float t_pos = steps_to_mpos( motor_steps[axis], axis );
                 if(t_pos != backlash_data[axis].prev_target){
                     if(approach){
                         backlash_data[axis].prev_direction   = !config->_axes->_axis[axis]->_homing->_positiveDirection ? MotionDirection::Negative : MotionDirection::Positive;
@@ -124,7 +120,6 @@ void backlash_Reset_for_homing(bool approach, uint8_t homing_mask) {
             }
         }
     }
-    //grbl_msg_sendf ( CLIENT_ALL, MsgLevel::Info, "BKSL Initi to (%3.3f, %3.3f, %3.3f) after homing!! ", backlash_data[0].prev_target, backlash_data[1].prev_target, backlash_data[2].prev_target );
 }
 
 //When ever the mechine origin is reset (homing) bcklash also need to reset
--- a/FluidNC/src/Backlash.cpp
+++ b/FluidNC/src/Backlash.cpp
@@ -16,21 +16,25 @@ void backlash_ini() {
         memset(&backlash_data[i], 0, sizeof(backlash_data_t));  // Clear planner struct
         backlash_data[i].backlash_enable    = true;
         backlash_data[i].prev_direction     = MotionDirection::Nutral; //DIR_NEUTRAL;
-        backlash_data[i].prev_target        = 0.000;
+        backlash_data[i].prev_target        = 0.0f;
     }
 }
 
+//This method check if the intended motion to target require any backlash correction
+//if needed it will create intermedeate (secret*) plan_buffer_line() call for the required correction motion.
+//this corrrection motion is supposed to be hidden from the system as much as possible.
+//This method should be called by plan_buffer_line() for all the none backlash motions, 
+//so that all the motion direction changes and previos positions of the machine are tracked. 
 bool backlash_Compensate_befor_target(float* target, plan_line_data_t *pl_data){
-    // Initialize planner data struct forbacklash motion blocks.
+    // Initialize planner data struct for backlash motion blocks.
     plan_line_data_t  plan_data_backlash;
     plan_line_data_t* pl_data_backlash = &plan_data_backlash;
     memcpy(pl_data_backlash, pl_data, sizeof(plan_line_data_t)); 
 
     bool backlash_comp_needed = false;
 
-    //pl_data_backlash->is_backlash_motion = true;  
     //the backlash correction is hidden from any future GRBL or gcode calculations, 
-    //in effect it is almost like a hardware correction which is not visible to software. 
+    //in effect it should look like a hardware correction which is not visible to software. 
     pl_data_backlash->motion.backlashMotion = 1;
 
     if(pl_data_backlash->motion.systemMotion == 1){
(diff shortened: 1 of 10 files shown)


atlaste:
First off, thanks for your contribution; it's not every day that people attempt to contribute to the planner buffer, so that's interesting. I looked through the backlash correction proposal. Unfortunately, I'm not sure if it's going to work like this. A few things that come to mind:

- The steps in mpos and mm should be in sync. So if you correct in one direction for backlash, it should probably compensate that in a next move somehow. By doing small hidden moves, I would think there could be an error that gradually increases. However, mechanically, I would argue that backlash is a true constant that relates to the slack in the gearing. 
- It's also interesting what would happen with a lot of tiny moves, like -say- milling a pcb. 
- I'm also not sure about the ramp-up, ramp-down behavior and what would happen when the planner buffer is reset. 
- Doing a reverse motion after a normal motion, has to inverse the direction pin of the stepper, which is usually a magnitude slower than the pulses themselves. Instead of compensating it - can't we just shorten the moves a tiny bit according to backlash? 
- I can't remember where kinematics are being done, and I don't know how it relates to backlash compensation. That is something I would have to check though. 

There's also minor details like clang-format, and that PR's should always target the 'Devt' branch.

pagabee:
If this helps, the backlash correction is implemented in this GRBL fork:
https://github.com/Schildkroet/GRBL-Advanced

nsiatras:
I often look to see if the backlash compensation feature has been Implemented in FluidNC. 
But the FluidNC project development says 

"Our current position...

We do not support backlash compensation at this time. It requires that the firmware know which side of travel the backlash is at all times. Lightweight machines that are prone to chatter tend to scramble the backlash. Backlash compensation can make things worse. Physical means of limiting backlash like preloading tend to work better. We prefer to spend our time on other features.

If you know of any research papers addressing backlash on lightweight machines, please send them our way."

After a google search for "lightweight cnc machines and backlash compensation" I found nothing. This makes perfect sense as there is no reason to have a lightweight CNC. CNC machines are machines that are built to perform precision movements. No sane person would build a CNC machine with the intention of being "light".

Anyway... The 2 photos I upload were taken by me. As you will see Backlash compensation is something that works. Siemens, Fanuc, Heidenhain and all CNC controller manufacturers worldwide have the same opinion, as they all have Backlash Compensation in their software.

Backlash compensation is a great feature that is still been missing from FluidNC. Hope FluidNC devs can see the photo.

image

PS. My tests are done using GRBL_ESP32. I implemented backlash compensation to GRBL_ESP32 but unfortunately I cannot make it work with Arcs. It only works with G0 and G1 commands.

kutukvpavel:
@nxperera Did you test this code on an actual machine? Can it be used if simply merged?

Overall, I second that backlash compensation is an important feature for FluidNC. While I agree that backlash compensation is more or less useless for, let's say, a "conventional" lightweight CNC router that uses ballscrews, this is not the only application of FluidNC. In my experience, retrofits of old mills and lathes are common, and those machines use leadscrews that introduce up to 1mm of repeatable backlash, which renders the whole retrofit with FluidNC useless.

Another approach to this issue would be to implement backlash compensation in a G-Code postprocessor. However, this approach requires the postprocessor to keep track of the position, which is not always trivial to implement (for arcs an helixes, for example), and essentially repeats controller's functionality.

MitchBradley:
The developers do not want to support backlash compensation considering all of the caveats about configuring and using it correctly, and situations where it does not work correctly.  Feel free to maintain a fork and support it.

nsiatras:
> Another approach to this issue would be to implement backlash compensation in a G-Code postprocessor. However, this approach requires the postprocessor to keep track of the position, which is not always trivial to implement (for arcs an helixes, for example), and essentially repeats controller's functionality.

This solution is wrong and should not be advised. 
CNC machines have tool change procedures, probing, canned cycles and other procedures that cannot be tracked in post processing.

Backlash compensation is related to the machine and the motion controller.
Therefore **backlash compensation should be performed on the motion controller and on the motion controller only.**

MitchBradley:
> Backlash compensation is related to the machine and the motion controller. Therefore **backlash compensation should be performed on the motion controller and on the motion controller only.**

The motion controller does not have information about milling strategies such as climb vs conventional milling, so it cannot account for tool side loading, which affects actual position on sloppy machines.  The most effective compensation strategy depends on many assumptions and details.  Conundrums like this are the reason why the FluidNC developers decline to support backlash compensation.  As I said, if anyone else is willing to take it on, feel free to maintain and support a fork.

nsiatras:
@MitchBradley 

**Backlash compensation minimizes the small gaps or slack between mating components**, such as gears, lead screws, or ball screws. These gaps cause a delay in movement when reversing direction, leading to inaccuracies in positioning. Backlash compensation is not related to milling strategies.

Backlash compensation can be achieved with the following methods

1. **Software Compensation** (Motion Controller)
2. **Mechanical Compensation** (Preloaded Ballscrews, Anti-Backlash Gears etc.)
3. **Electronic Compensation** (Servo Systems with Linear encoders)

Here we discuss Method 1 - Software Compensation. 

In software compensation, the motion controller's software keeps track of the backlash amount and offsets movements accordingly. **When the system reverses direction, the software commands an extra movement equal to the measured backlash to ensure accurate positioning.**

If you take a look at the photos I uploaded 3 years ago, you can see that software compensation, when used with a machine that has been built to milling machine specifications, gives good results. Sloppy and lightweight machines are not intended to be milling machines. Perhaps someone could build a lightweight machine and use it as a laser or plasma cutter. Even in this case, the software compensation would be useful with machines using rack and pinion.

PS. I 've implemented the Backlash Compensation algorithm in RabbitGRBL. 
It is not difficult to implement it on FluidNC

**Code:** https://github.com/SourceRabbit/RabbitGRBL/tree/main/RabbitGRBL/src/Backlash
**Demo Video:** https://www.youtube.com/watch?v=-t-U46T7R6s

ryan1smith:
Hey @nsiatras thanks for the post and links. I've discovered I have about 20-30 microns backlash and 5-20 um absolution position variation using DFU1605 C5 ball screws and angular contact bearings in the fixed block. I also have closed-loop steppers with rotary encoders (1000 PPR only, 1.25 um/count on 5mm/rev screws) and 1 um linear glass scales. I also have a camera and the glass scales seem to be accurate to at least a few microns from the standards I've tested with. 

I also think it may be advantageous to preload any ball screw or belt in one direction so backlash compensation isn't needed electronically but only absolute position compensation is. I tightened the DFUs and that never resolved the backlash. I didn't look for the root cause but seemed like it was from rotational clearance, maybe from clearance between the balls. It seems backlash compensation would be very difficult to implement electronically. Any thoughts?

In some name brand CNCs they used to use a compensation table and I'm most interested in eliminating backlash mechanically and absolute position through the controller. Were you able to implement any type of backlash and/or absolute position compensation in FluidNC? If so do you mind sharing a little more detail on how? So far I haven't done any research into the guts of position control or planning in GRBL/FluidNC/etc. but this would certainly be useful.

nsiatras:
@ryan1smith  I 've implemented the Backlash Compensation algorithm in RabbitGRBL. You can find the backlash compensation manager on the following link https://github.com/SourceRabbit/RabbitGRBL/tree/main/RabbitGRBL/src/Backlash

Find references of BacklashManager::CompensateBacklash, BacklashManager::ResetTargets and BacklashManager::SynchPositionWhileUsingProbe methods inside the project and you will find out the "path"

ryan1smith:
Thanks for the quick reply @nsiatras. I think I'm looking for more of an absolute position compensation but this looks reasonable for backlash if I decide to go that way.

ReeceRobinson:
I agree this feature can be useful. I hope enough users in the community can raise this to a level of interest so we can make our builds that much better.

kutukvpavel:
I have created a fork and ported the backlash compensation algorithm from RabbitGRBL to FluidNC (also I implemented selective motor activation during jogging). Like @nsiatras said it wasn't hard. The core FluidNC functionality with compensation disabled seems stable, but I have yet to activate the compensation in my config file and mill some circles.

bdring:
Let us know the results of your testing?

Is backlash always a positive float value? The config item handler should constrain the value to valid and reasonable values.

bdring:
@kutukvpavel 

I just tested your fork. With backlash at 0 it appears to work normally (not fully tested). 

With backlash on X at 1.00, I cannot home the X axis. If I unlock without homing, `G91 G0 X0` does not move. The machine makes some repeating noise on the X axis, but it does not appear to move.

bdring:
It appears that BL comp is initially using the G1 feed rate. If you do a G0 before a feed rate is changed from 0, it uses F1. 

```
$X
[MSG:INFO: Caution: Unlocked]
ok
<Idle|MPos:0.000,0.000,0.000|FS:0,0>
g91 G0 X10
ok
<Run|MPos:0.000,0.000,0.000|FS:1,0>
<Run|MPos:0.000,0.000,0.000|FS:1,0>
```

If you set a feed rate it appears to use the max rate.

```
F100
ok
X100
ok
<Run|MPos:20.735,0.000,0.000|FS:2100,0>
<Run|MPos:30.228,0.000,0.000|FS:2880,0>
<Run|MPos:41.366,0.000,0.000|FS:3000,0>
```

kutukvpavel:
Thanks @bdring !

Well, looks like its not going to be so easy after all. Probably will take me a while to figure this out.

You have tested the main branch, right?

PR #27, opened 2021-09-26
Fix WebUI access to NVS Settings

MitchBradley:
and reordered the NVS settings so the most likely ones
to change are at the top.


bdring:
@MitchBradley  This is out of sync with devt. The wifi_mode enum does not have the new options like, "Off".  $WiFi/Enable is still here.

PR #32, opened 2021-09-28
Haunyang testing

atlaste:
Changes to VFD spindles, specifically Huanyang.

### DIFF
--- a/FluidNC/src/Spindles/HuanyangSpindle.cpp
+++ b/FluidNC/src/Spindles/HuanyangSpindle.cpp
@@ -222,8 +222,9 @@ namespace Spindles {
             return [](const uint8_t* response, Spindles::VFD* vfd) -> bool {
                 uint16_t value = (response[4] << 8) | response[5];
 #ifdef DEBUG_VFD
-                log_debug("VFD: Max frequency = " << value);
+                log_debug("VFD: Max frequency = " << value / 100 << "Hz " << value / 100 * 60 << "RPM");
 #endif
+                log_info("VFD: Max speed:" << (value / 100 * 60) << "rpm");
 
                 // Set current RPM value? Somewhere?
                 auto huanyang           = static_cast<Huanyang*>(vfd);
@@ -239,8 +240,9 @@ namespace Spindles {
                 uint16_t value = (response[4] << 8) | response[5];
 
 #ifdef DEBUG_VFD
-                debug_all("VFD: Min frequency set to " << value);
+                log_debug("VFD: Min frequency = " << value / 100 << "Hz " << value / 100 * 60 << "RPM");
 #endif
+                log_info("VFD: Min speed:" << (value / 100 * 60) << "rpm");
 
                 // Set current RPM value? Somewhere?
                 auto huanyang           = static_cast<Huanyang*>(vfd);
--- a/FluidNC/src/Spindles/HuanyangSpindle.cpp
+++ b/FluidNC/src/Spindles/HuanyangSpindle.cpp
@@ -203,6 +203,7 @@ namespace Spindles {
         data.msg[4] = dev_speed & 0xFF;
     }
 
+    // This gets data from the VFS. It does not set any values
     VFD::response_parser Huanyang::initialization_sequence(int index, ModbusCommand& data) {
         // NOTE: data length is excluding the CRC16 checksum.
         data.tx_length = 6;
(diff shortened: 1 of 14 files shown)


FluidNC/src/Spindles/HuanyangSpindle.cpp
```diff
@@ -222,8 +223,9 @@ namespace Spindles {
             return [](const uint8_t* response, Spindles::VFD* vfd) -> bool {
                 uint16_t value = (response[4] << 8) | response[5];
 #ifdef DEBUG_VFD
-                log_debug("VFD: Max frequency = " << value);
+                log_debug("VFD: Max frequency = " << value / 100 << "Hz " << value / 100 * 60 << "RPM");
```
atlaste, line 226: Careful, this is integer math. `value/100*60` does `(value/100)*60`, which is probably incorrect. It should be `(value*60)/100`.

I also make it a habit to put the math between '(...)', because '<<' is an operator and therefore follows the precedence rules. In this case:

log_debug("VFD: Max frequency = " << (value / 100) << "Hz " << (value * 60 / 100) << "RPM");

FluidNC/src/Spindles/HuanyangSpindle.cpp
```diff
@@ -239,8 +241,9 @@ namespace Spindles {
                 uint16_t value = (response[4] << 8) | response[5];
 
 #ifdef DEBUG_VFD
-                debug_all("VFD: Min frequency set to " << value);
+                log_debug("VFD: Min frequency = " << value / 100 << "Hz " << value / 100 * 60 << "RPM");
```
atlaste, line 244: Same as above

FluidNC/src/Spindles/HuanyangSpindle.cpp
```diff
@@ -222,8 +223,9 @@ namespace Spindles {
             return [](const uint8_t* response, Spindles::VFD* vfd) -> bool {
                 uint16_t value = (response[4] << 8) | response[5];
 #ifdef DEBUG_VFD
-                log_debug("VFD: Max frequency = " << value);
+                log_debug("VFD: Max frequency = " << value / 100 << "Hz " << value / 100 * 60 << "RPM");
 #endif
+                log_info("VFD: Max speed:" << (value / 100 * 60) << "rpm");
```
atlaste, line 228: Same as above

FluidNC/src/Spindles/VFDSpindle.cpp
```diff
@@ -347,7 +334,7 @@ namespace Spindles {
                 setSpeed(dev_speed);
             }
         }
-        if (!supports_actual_speed()) {
+        if (!use_delay_settings()) {
```
atlaste, line 337: The '!' is incorrect; should be `if (use_delay_settings())`

example_configs/6_pack_external_huanyang.yaml
```diff
@@ -0,0 +1,173 @@
+board: 6 Pack
+name: 6 Pack External XYZ Haunyang
+stepping:
+  engine: I2S_STREAM
+  idle_ms: 250
+  pulse_us: 2
+  dir_delay_us: 1
```
atlaste, line 7: Just spotted this. Normally dir / disable delays are higher than pulse. I'd rather just leave them out, than to specify them incorrectly.

example_configs/6_pack_external_huanyang.yaml
```diff
@@ -0,0 +1,173 @@
...
+limits_two_switches_on_axis: false
+disable_laser_during_hold: true
+use_line_numbers: false
+
+Huanyang:
+  uart:
+    txd_pin: gpio.26
+    rxd_pin: gpio.16
+    rts_pin: gpio.4
+    baud: 9600
```
atlaste, line 167: 9600 baud is the default? Wow, that's really low... most VFD manuals I've seen specify a default that's way higher.

I'd also personally prefer a higher value to a lower value. 9600 baud is roughly 30 messages a second. The ESP32 threading capabilities are limited and not preemptive. I'm not too worried about setting the ena/disable here, but more about the safety polling that happens very frequently -- I wouldn't want that to stall other things. To be fair, it shouldn't, and the config is fine as-is, and a lower baud rate is probably better for EMI... it's a bit of a trade-off.

IIRC the default for the H2A is 19200 or 38400 baud.

atlaste (commented):
Just a bit of nitpicking and minor details.

PR #58, opened 2021-10-20
Clean up kinematics interface

### LINKED ISSUE #57: Clean up Kinematics -- mc_line() vs cartesian_to_motors()
**Please describe the feature you would like implemented**

mc_line() is special in that it operates in a post-kinematic environment, while the other mc_ operations are in a pre-kinematic environment.

This means that code that calls, e.g., mc_arc() cannot also call mc_line(), but must instead call cartesian_to_motors().

Let's rename cartesian_to_motors() as mc_line() and rename mc_line() to something more like mc_move_motors().

This would fit with mc_move_motors() receiving the motors computation from cartesian_to_motors().

**Why do you think this would improve FluidNC?**

It was confusing to understand the special status of mc_line() vs mc_arc when reviewing the kinematic code, to understand which operations were operating in which coordinate spaces.

**What do you need the feature for?**

To make kinematic custom code easier to understand.

The separation between cartesian and motor planning would become clearer.

**Do you know of other users who need this feature?**

A number of custom kinematics would be made simpler to understand.

pentacular:
This makes a clearer separation between cartesian and motor space operations.

Previously mc_line() would operate in motor space, and cartesian_to_motors() would be the operation to call to plan a linear motion in cartesian space -- in contrast to mc_arc, etc, which operate in cartesian space.

With this change mc_linear() plans a linear motion in cartesian space, and mc_move_motors() plans a linear motion in motor space, with cartesian_to_motors providing a translation from mc_linear() to mc_move_motors().

Closes #57

### DIFF
--- a/FluidNC/src/System.cpp
+++ b/FluidNC/src/System.cpp
@@ -8,7 +8,7 @@
 
 #include "System.h"
 #include "Report.h"                 // report_ovr_counter
-#include "MotionControl.h"          // mc_linear
+#include "MotionControl.h"          // motors_to_cartesian
 #include "Config.h"                 // MAX_N_AXIS
 #include "Machine/MachineConfig.h"  // config
 


pentacular:
Closes #57

PR #82, opened 2021-10-26
Less boot messages

bdring:
Removing some boot messages for things that are defaulting. These are typically setup to do no IO. If if defaults to some IO, you will see it later. Less messages will help reduce the clutter and noise with support posts.

Added one meta: key. This can be used by the author to store a single string, like a comment. It will not print, but it will be saved with $CD.

### DIFF
--- a/FluidNC/src/Machine/MachineConfig.cpp
+++ b/FluidNC/src/Machine/MachineConfig.cpp
@@ -63,12 +63,10 @@ namespace Machine {
         }
 
         if (_coolant == nullptr) {
-            //log_info("Coolant: using defaults");
             _coolant = new CoolantControl();
         }
 
         if (_probe == nullptr) {
-            //log_info("Probe: using defaults");
             _probe = new Probe();
         }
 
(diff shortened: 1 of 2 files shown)


FluidNC/src/Machine/MachineConfig.cpp
```diff
@@ -62,12 +63,12 @@ namespace Machine {
         }
 
         if (_coolant == nullptr) {
-            log_info("Coolant: using defaults");
+            //log_info("Coolant: using defaults");
```
MitchBradley: Just delete them.  There is no point in keeping around dead code that is trivial to reconstruct.  It just creates clutter.

FluidNC/src/Machine/MachineConfig.h
```diff
@@ -82,9 +82,10 @@ namespace Machine {
 
         String _board = "None";
         String _name  = "None";
-
+        String _meta  = "";
 #if 1
-        static MachineConfig*& instance() {
+            static MachineConfig*&
```
MitchBradley: Looks like a clang-format botch here.

PR #85, opened 2021-10-27
I2SI implementation ... 

phorton1:
Hey there, 

Here's a hardware implementation of an I2SI bus and pins.

Take it for what it's worth.    Or leave it.   Just wanna make sure you get to see it.

- uses ESP32s other I2S peripheral (number 1)
- issues real interrupts to clients who call attachInterrupt()
- allows for task based non-ISR "shiftIn" approach with "use_shift_in" config variable
- tested with 74HC165 and limit and probe pins

p.s. don't know why you can't automatically merge this.  I went back, checked out devt and re-applied my changes.  

Oh well :-)   It's the thought that counts.

Peace and happiness!

Patrick

### DIFF
--- a/FluidNC/src/Main.cpp
+++ b/FluidNC/src/Main.cpp
@@ -69,13 +69,13 @@ void setup() {
             }
             if (config->_spi) {
                 config->_spi->init();
-
-                if (config->_sdCard != nullptr) {
-                    config->_sdCard->init();
-                }
+            }
+            if (config->_sdCard != nullptr) {
+                config->_sdCard->init();
             }
             if (config->_i2si) {
                 config->_i2si->init();
+            }
 
             Stepper::init();  // Configure stepper pins and interrupt timers
 


FluidNC/src/Main.cpp
```diff
@@ -74,6 +74,8 @@ void setup() {
                     config->_sdCard->init();
                 }
             }
+            if (config->_i2si) {
+                config->_i2si->init();
```
MitchBradley, line 79: Where is the closing brace for this block?  If it is the one on line 91, the indentation is wrong.

FluidNC/src/Pins/I2SIPinDetail.cpp
```diff
@@ -0,0 +1,118 @@
...
+        _readWriteMask(0)
+    {
+        bool claimed = Machine::I2SIBus::getPinsUsed() & (1 << index);
+        Assert(index < Machine::I2SIBus::s_max_pins-1, "Pin number is greater than max %d", Machine::I2SIBus::s_max_pins-1);
+        Assert(!claimed, "Pin is already used.");
+
+        // User defined pin capabilities
+
+        for (auto opt : options)
+        {
```
MitchBradley, line 28: The formatting does not match our coding style standard - see https://github.com/bdring/FluidNC/blob/main/CodingStyle.md

FluidNC/src/Machine/I2SIBus.cpp
```diff
@@ -0,0 +1,185 @@
...
+                return;
+            }
+
+            _bck.setAttr(Pin::Attr::Output);
+            _ws.setAttr(Pin::Attr::Output);
+            _data.setAttr(Pin::Attr::Input);
+
+            if (_use_shift_in)
+            {
+                xTaskCreatePinnedToCore(
```
atlaste, line 85: I'd still rather tie this to the stepper ISR than make it a task. For repeatibility, I think that would work better, because the inputs would become (step) motion dependent.

PR #93, opened 2021-11-03
Added $MS for motor status. Fixed bug in MachineConfig for user outputs.

atlaste:
As the title describes. Seems to be more bugs tho...


FluidNC/src/Machine/MachineConfig.cpp
```diff
@@ -79,6 +79,7 @@ namespace Machine {
 
         if (_userOutputs == nullptr) {
             _userOutputs = new UserOutputs();
+            _userOutputs->init();
```
MitchBradley, line 82: I think this is unnecessary since I moved the array population to the UserOutputs constructor.

FluidNC/src/Motors/NullMotor.h
```diff
@@ -2,15 +2,43 @@
 
 #include "MotorDriver.h"
 
+#include <cstdint>
+
 namespace MotorDrivers {
     class Nullmotor : public MotorDriver {
+        int32_t direction = 0;
+        int32_t steps     = 0;
+        int     state     = 0;
```
MitchBradley, line 11: Is state actually used anywhere?  It gets set and unset but other than that I don't see anywhere that the information goes.

FluidNC/src/Motors/NullMotor.h
```diff
@@ -2,15 +2,43 @@
...
+
     public:
         Nullmotor() = default;
 
         bool set_homing_mode(bool isHoming) { return false; }
 
         bool isReal() override { return false; }
 
+        void step() override {
+            Assert(!disabled, "Cannot step motor while disabled.");
```
MitchBradley, line 22: I think it is unwise to Assert in something that is usually called from an ISR

FluidNC/src/Motors/NullMotor.h
```diff
@@ -2,15 +2,43 @@
...
+                steps += direction;
+            }
+        }
+
+        void unstep() override {
+            state = 0;
+        }
+
+        void set_direction(bool dir) override {
+            Assert(!disabled, "Cannot set direction while disabled.");
```
MitchBradley, line 34: Again, Assert from ISR

MitchBradley (commented):
If the comments are addressed, I don't have any further issues with this, but overall, it has the feel of a debugging tool that was intended for a special purpose use.  Do we want to keep it around?

atlaste:
I only see your comments now, I really have to fix my github preferences... not sure why I didn't get a proper notification. Anyways,

> If the comments are addressed, I don't have any further issues with this, but overall, it has the feel of a debugging tool that was intended for a special purpose use. Do we want to keep it around?

I have to clarify I think. This is indeed intended to be a debugging tool, specifically to test the behavior of the stepper engine with motor movements. I considered adding a 'debug_motor', but since I noticed that I'm usually developing on null_motor, it made more sense to just add it here. It does 2 things:

1. Count the number of steps from the origin.
2. Test if enable behavior is correct.

Since these are quite basic functionalities of the stepper engine, I made this PR. A better place would be to make a proper unit test from this... but due to all the dependencies to other components I don't see that happening anytime soon.

FluidNC/src/Motors/NullMotor.h
atlaste: In general, I agree that asserts are "unwise" in an ISR. However, the rationale here was that it should "crash early" if things go wrong. After all, if the ena/dir is incorrect, it will just crash - but then again, if it's wrong and a machine moves there are bigger issues... It was a bit of a trade-off here.

FluidNC/src/Machine/MachineConfig.cpp
atlaste: OK.

FluidNC/src/Motors/NullMotor.h
atlaste: Might be a left-over. I have to check. Let's first decide what to do with the PR.

PR #124, opened 2021-11-17
Read switches after init

bdring:
This will give accurate reporting before any have triggered an ISR.

### DIFF
--- a/FluidNC/src/Machine/LimitPin.cpp
+++ b/FluidNC/src/Machine/LimitPin.cpp
@@ -45,23 +45,6 @@ namespace Machine {
     }
 
     void IRAM_ATTR LimitPin::handleISR() {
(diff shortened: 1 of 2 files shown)


FluidNC/src/Machine/LimitPin.cpp
```diff
@@ -45,23 +45,24 @@ namespace Machine {
...
-            }
-        } else {
-            if (_posLimits != nullptr) {
-                clear_bits(*_posLimits, _bitmask);
-            }
-            if (_negLimits != nullptr) {
-                clear_bits(*_negLimits, _bitmask);
-            }
-        }
+        //bool pinState = _pin.read();
```
MitchBradley: Please don't leave large hunks of commented-out code lying around.

PR #131, opened 2021-11-19
Fix tests

atlaste:
Made the unit test project work again. Fixed a few issues in the code.


atlaste:
Changes:

- n_axis is not a constant, so cannot be used for float arrays. MAX_N_AXIS is the correct thing.

atlaste:
I'm going to take this one step further and ensure that `pio test -e native` works. That allows us all to use unit tests AND should enable them in Travis CI.

DO NOT PULL!

PR #134, opened 2021-11-23
Move Vtable to RAM

MitchBradley:
This patch fixes crashes when loading WebUI during motion.  To repro the problem, start a long move like `G0 x2000` and then refresh the WebUI main page while the move is still running.

The root cause of the problem is that ISR code cannot run reliably from FLASH because FLASH accesses from the other core will prevent proper reload of the FLASH cache.  ISR code must be marked with IRAM_ATTR, but that is not enough - it must also avoid switch statements - whose jump tables are placed in FLASH - and also C++ virtual methods have FLASH-resident data structures.

The fix is twofold.  First, we rewrite a few switch statements inside IRAM_ATTR code to use if chains instead of switch().  Second, we change the linker script to put selected vtables in DRAM instead of FLASH.  The technique for re-locating the vtables is described in FluidNC/ld/esp32/README.md , and all of the files necessary for that are isolated in that directory, apart from a single reference in platformio.ini

### DIFF
--- a/FluidNC/ld/esp32/README.md
+++ b/FluidNC/ld/esp32/README.md
@@ -1,0 +1,121 @@
+# Vtable Linker Fixups for FluidNC
+
+## The Problem
+
+FluidNC uses C++ abstract classes with virtual methods for several
+things including Spindles and Motors.  Unfortunately, virtual methods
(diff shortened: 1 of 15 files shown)


MitchBradley:
Superseded by #143

PR #135, opened 2021-11-23
Simplified UserOutputs into a single level

MitchBradley:
The two-level structure was causing problems with the timing of
construction, configuration setting, and initialization.


bdring:
Tested this with a few config files. It worked and caught the intentional errors.

👍

PR #142, opened 2021-11-27
Allow direct input from websockets

MitchBradley:


deemoss:
Hi Mitch, I have been struggling to send simple Gcode with Python and Websockets.

I have the following code that executes OK, but I get no motion from steppers (the steppers work fine from WebUI). 

When I run this:
```
from websocket import create_connection
import time

ws = create_connection("ws://xyxrf:81")
print(ws.recv())
gcode = "G1X1Y1F1000"
print("Sending " + gcode)
ws.send(gcode)
print("Sent " + gcode)
print("Receiving...")
result = ws.recv()
print("Received '%s'" % result)
ws.close()
time.sleep(5)
```

... I get in terminal:

```
CURRENT_ID:0
Sending G1X1Y1F1000
Sent G1X1Y1F1000
Receiving...
Received 'ACTIVE_ID:0' 
```

Also, Wireshark confirms that the gcode goes through:

[https://www.dropbox.com/s/cuijqbfld1gtj77/wireshark.png?dl=0](url)

Why why could this be?

Cheers

MitchBradley:
What package do I need to install to try it?  There are multiple websocket libs. and your message has two different spellings for the name.  I tried, with Python 3.7.3.
```
pip install websocket
python3
from websocket import create_connection
```
 and got `ImportError: cannot import name 'create_connection' from 'websocket'`

Then I did
```
pip uninstall websocket
pip install Websockets
python3
from websocket import create_connection
```
and got `ModuleNotFoundError: No module named 'websocket'`

You may have noticed that I get irritated when my time is wasted due to incomplete or inaccurate information.  If you want free consulting, you need to respect my time.

bdring:
Ya, I tried every combination of imports I could think of without luck.

Do WS request require a line end? Does python print give you that?

MitchBradley:
I found the websocket-client package.  Bart is right; you need to add a newline like `gcode = "G1X1Y1F1000\n"`

deemoss:
Yes! The newline... it works! Thanks!

PR #148, opened 2021-11-30
H100 VFD control

atlaste:
Implementation of the popular H100 RS485 VFD protocol. The MD document describes the details.

https://nl.aliexpress.com/item/1005003108743467.html are the inverters. These are sold by a lot of vendors at the moment.

Tested by @bcrazycramer , see the different Discord chats for all the details if you like...

### DIFF
--- a/FluidNC/src/Spindles/H100Spindle.cpp
+++ b/FluidNC/src/Spindles/H100Spindle.cpp
@@ -1,0 +1,174 @@
+// Copyright (c) 2020 -	Stefan de Bruijn
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+/*
+    H100Spindle.cpp
+
+    This is for a H100 VFD based spindle via RS485 Modbus.
+
+                         WARNING!!!!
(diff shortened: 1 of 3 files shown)


atlaste:
WIKI docs:
---
This is the standard H100 VFD (all power levels). Part numbers typically look like H100-xxx.

The VFD registers need to be setup prior to use. FluidNC will not change any of the registers. Read your VDF documentation on how to do that. 

The most relevant sections are:
- F011 (min frequency) 
- F005 (max frequency)

The min and max speeds set in the VFD will be displayed in the startup messages. They might be spread out through the messages, because they are coming from a separate task.

[MSG:INFO: VFD: Max speed:24000rpm]
[MSG:INFO: VFD: Min speed:6000rpm]

A minimum speed is typical with VFD spindles because they lack power and can overheat at lower speeds. For speed_map, feedback and the RS485 adapter settings, see the Huanyang section for more details.

If you don't specify the speed_map, the firmware will automatically put in the default values based on the frequencies that are set in the VFD. Only specify a speed_map if you use a gearbox or other contraption.

PR #159, opened 2021-12-06
Update HOWTO-INSTALL.txt

Longus:
A few directional sign posts for navigating thru the rathole.

### DIFF
--- a/install_scripts/macos/HOWTO-INSTALL.txt
+++ b/install_scripts/macos/HOWTO-INSTALL.txt
@@ -1,14 +1,29 @@
+211205_151757: from [https://github.com/bdring/FluidNC/edit/ReleaseEngineering/install_scripts/macos/HOWTO-INSTALL.txt]
+
+## Requirements
+
+This install requires:
+
+- A Serial USB Port (such as the Silicon Labs Driver)
+- Python3
+
+For help with these, see https://github.com/Longus/FluidNC/wiki/Set-Up-FluidNC-Using-macOS
+
 ## Installing FluidNC on your ESP32
 
 Unpack this FluidNC release Zip file.
 
 Plug your ESP32 into a USB port.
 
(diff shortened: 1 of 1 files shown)


install_scripts/macos/HOWTO-INSTALL.txt
```diff
@@ -1,14 +1,29 @@
...
+
+## Requirements
+
+This install requires:
+
+- A Serial USB Port (such as the Silicon Labs Driver)
+- Python3
+
+For help with these, see https://github.com/Longus/FluidNC/wiki/Set-Up-FluidNC-Using-macOS
+
```
MitchBradley, line 11: Re the following text in that page:

> hold down the 'io0' button on the ESP32 (next to the usb port),

In principle, with a "devkit c" version of the ESP32 - and most other modules - holding down IO0 is unnecessary, because esptool automatically cycles that signal via modem control signals on the serial port.  I have had a very small number of situations where it was necessary to do the manual thing.  It "almost always" Just Works.

> and in 'Terminal.app' run /Users/your_username/Desktop/fluidnc-vX.X.X-macos/install-wifi.sh (I couldn't get it to run unless I used the full path)

Did you try typing "./install-wifi.sh" (note the ./) ?  Unix systems in general require you to type ./ in order to execute a script or executable in the current directory.  This is a security feature, to prevent an attack whereby somebody drops a poisoned version of some system command in a directory.  The unsuspecting user then runs some "safe" system command like, say, "ls", but there is a fake version of "ls" in the current directory that runs instead, and does something evil.  Requiring you to type ./<whatever> sort of blocks this, because nobody is likely to type ./ls to run the system ls. .

Typing "sh install-wifi.sh" should work too, per the instructions.  Whether to prefer "sh install-wifi.sh" vs "./install-wifi.sh" is a matter of personal preference; either works.

PR #181, opened 2021-12-16
Escape grbl magic

MitchBradley:


MitchBradley:
GRBL treats !, ?, and ~ as realtime characters so if you try to
put them in string values via the serial port UI, bad things
happen.  The solution is to escape those character using URI
encoding - %nn .  The escape codes are
! - %21
? - %3F
~ - %75

and the escape character % must itself be escaped:

% - %25

On input, any character at all can be so represented by its hex
value, but on output, only the characters above will be replaced
by their escape sequences.

The escaping applies only to interaction via serial-style channels,
i.e. the channels that interpret Grbl realtime characters,  It does
not affect the WebUI settings page.  In that page, you can still
include the special characters in plaintext.

MitchBradley:
I thought of another way to handle this problem.  When in line editing mode, the realtime characters could be handled as such only at the beginning of a line.  Inside an edited line, i.e. when echoing is one and other characters have already started the line, those characters could be treated as ordinary printable characters.

This violation of the Grbl protocol might be acceptable in light of the fact that existing senders do not turn on line editing mode, so they will not see the different.  Furthermore, senders typically do not inject real time characters into the middle of a line, instead sending pre-composed lines as a unit, and separately sending realtime characters.

This could be done in conjunction with this patch, both handling the escape sequences for the benefit of the consoles in existing senders, and also having the passthrough behavior for line-editing-savvy senders.

PR #220, opened 2021-12-30
Adding partition options

bdring:


app3M_spiffs1M_8MB.csv
```diff
@@ -0,0 +1,6 @@
+# Name,   Type, SubType, Offset,  Size, Flags
```
MitchBradley: I'm wondering if maybe these files should be in a subdirectory.  For the ld scripts that were used to fix the flash/IRQ problem, I created the subdirectory ./FluidNC/ld/esp32 and put the stuff there.  Perhaps we should do the same for partition maps, since they are very platform dependent.  ./FluidNC/layout/esp32/ ?

PR #232, opened 2022-01-06
strncasecmp function fix

bberkhin:
function strncasecmp does not work correctly, so unit tests fail

### DIFF
--- a/X86TestSupport/TestSupport/WString.cpp
+++ b/X86TestSupport/TestSupport/WString.cpp
@@ -96,38 +96,11 @@ StringAppender& operator+(const StringAppender& lhs, double num) {
 #ifdef _MSC_VER
 
 int strcasecmp(const char* lhs, const char* rhs) {
-	return ::_stricmp(lhs, rhs);
-	/*
-	int result = 0;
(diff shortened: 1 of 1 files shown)


atlaste:
Whoa you just saved me a lot of work, because this was exactly the intention of how I started this x86 support branch (and I didn't knew that you can use conio like this)! I'm actually surprised that it works with candle with just these changes.

I did an extensive review, and nagged about the things that I could find. One thing that is a bit of an issue, is that you targetted the 'main' branch instead of the 'devt' branch in the PR (which means we cannot take it as-is without breaking git), but tbh I don't think it will be a lot of trouble cherry-picking the code and re-target it.

I would very much like to have these remaining things fixed so we can get this in the main branch. If you have any questions or remarks, feel free to do so, or just join us on Discord.

bberkhin:
great that you liked my work )
I will re-target all changes to devt branch today or tomorrow and will send a new pull requsts. 
Coud you send me an invitation do Discord. My id: "Boris Berkhin#3986"

readeral:
@bberkhin https://discord.gg/aKAT4Dfb

Issue #249, opened 2022-01-15
Full step not working


VZP356:
Hello, I am bench testing a https://www.tindie.com/products/33366583/6-pack-universal-cnc-controller/ for a CNC router with tmc5160, and 5mm pitch ballscrews.

I would like to have full step, due to the torque loss with microstepping.
in the Tablet tab, i have tried G91 G1 Z5 F500 . 

This with: 
microsteps: 32 , steps_per_mm: 1280.000  does a full revolution
microsteps: 16 , steps_per_mm: 640.000  does a full revolution
microsteps: 2 , steps_per_mm: 80.000 does a full revolution
microsteps: 1 , steps_per_mm: 40.000 does only a fraction of a revolution ( i woul say 1/64th)

File changes over 32 in the Microstepping field, do not load or work 
Wiki: microsteps: This sets the microstepping level. Valid values are 1,2,4,8,16,32,64,128 & 256

What version of the firmware are you using?
3.3.0

How should I setup for full step? 
Update:
microsteps: 0 , steps_per_mm: 40.000 does a full revolution

MitchBradley:
That 0-instead-of-1 thing is an artifact of the TMCStepper library that we use for the low-level access to TMC devices.  It takes the following arguments: 256, 128, .., 8, 4, 2, 0, skipping 1.  I don't know why its author chose that odd representation.  We should probably work around it.

PR #250, opened 2022-01-15
TMCStepper library microstep strangeness

MitchBradley:
This lets you says microsteps: 1  for Trinamic (0 also works).  For some obscure reason, the TMCStepper uses the value 0 to mean 1.  This code compiles but has not been tested on Trinamic hardware.


MitchBradley:
Fixes #249

PR #271, opened 2022-02-04
Dynamixel

bdring:
Fixed: The dynamixel2 motors were never finished, tested or documented.


mbecker496:
I am looking for some help using the dynamixel servos with fluidnc. I am new to all this and any help would be great. My confusion comes in mostly with wiring the servos. All the boards I find are set up for stepper motors not servo motors, does this matter? How would you wire a RS485 servo to the board? Thanks in advanced 
Mike

Issue #240, opened 2022-01-10
Support ARM linux in install scripts.


easytarget:
My controller is permanently connected to a Raspberry PI, and I wanted to install and upgrade from that since it was already connected. 

So the specific, low priority, 'feature request' is for 32-bit armv7 install support in the release tooling.
- adding arm6 support would also cover the Pi1, Pi2 and v1 Pi Zero, but is probably less relevant
- 64bit arm support is not really an issue at present, but an official 64bit RasPI OS will happen eventually

With the install scripting you obviously have allowed for adding other architectures going forward, Until then I wanted to document how I got it working somewhere more public than Discord. 

----------

First, install esptool for the arm architecture; the quickest commandline way to get this is via python/pypi
```$ pip install esptool```

Then edit `tools.sh` in the unpacked zip folder and replace line 3 with:
```EsptoolPath=esptool.py```

Then follow instructions as normal.

Caveats:
This was done using a fully updated Bullseye release with Python3 as the default.
- afik this wil also work just fine on the older Stretch and Buster releases, and should also work for Python2. But I have no systems to test this on so ymmv.
- There is a version of esptool available via apt; `sudo apt install esptool` followed by `EsptoolPath=esptool` should work too; and may even be better advice. But I have not tested that, the pypi version of esptool is generally much newer.

MitchBradley:
We will accept a PR to add Raspberry Pi support to the release scripting if you author it and agree to support if/when people have problems.

easytarget:
As you identify, supporting various install scripts with binary blobs for differing OS's is not fun, and not something I'd be prepared to take on.

However; as an alternative, a generic installer that uses python directly with esptool.py, and avoids distributing any executables seems like a much better long-term solution.

I've got this ready, tested and working on a Pi4+Bullseye, Pi3a+Buster and my Fedora35 workstation. I am just fixing the readme and will drop a PR against DevT asap.

bdring:
Distributing a python script to Windows is a pain.

MitchBradley:
We are amenable to using the python + *.py technique on Linux and MacOS.  The precompiled executable method works pretty well on Windows, but is trouble-prone on other OSs.

easytarget:
#296 closes this nicely.

PR #272, opened 2022-02-04
Generic Linux install scripts based on esptool.py

easytarget:
As mentioned in #240

Additional to the current amd64 linux install scripts; based on them but using `esptool.py` directly
- Tests for and installs `esptool.py` via pypi if it is not present; will also set local binary path if needed
- Only  `tool.sh` and the readme have been modified, all other files are unchanged from the original linux tree

Readme has been updated; and a note added about group permissions and getting tkinter installed on headless machines so you can use fluidterm.
- Tested on Fedora35 (x64), Raspbian Bullseye on a Pi4, Buster on a Pi3a, and FreeBSD13 on x64

I'm happy to assist/support this going forward on linux and on a 'best effort' basis for other *nix related platforms; although I think it is simple enough to be pretty robust. Common issues can be documented as they arise, Issues with python or esptool.py itself (eg not caused by errors in the script) will be handled by referring users to the main esp and/or python forums.

I have not attempted to integrate this into the build system (`build-release.py`), someone more familiar with that side of the scripting will need to do that, it looked a bit non-trivial to me.


easytarget:
I'm just putting together changes to `build-release.py` for this, they should be ready later today.
I've been buried in JS for a month, when I looked at it yesterday it all just blurred.. bur re-looking today I realised how nicely written it is, and how it's quite easy to add another installer bu extending the current script (eg; not a 'hack').

easytarget:
Ok the release script will now build an additional package, `...linux-python3.zip`, and this is tested+works on my Fedora machine, but still needs your review and test etc.
I changes are trivial, added the relevant data for the new package and added logic to specify whether to package esptool or not for each package type.

install_scripts/linux-python3/tools.sh
```diff
@@ -0,0 +1,42 @@
...
+echo "$PATH" | grep '$HOME\/\.local\/bin' 2>&1 >/dev/null
+if test "$?" != "0"; then
+    export PATH="$HOME/.local/bin:$PATH"
+fi
+
+# Install esptool.py if needed
+which esptool.py 2>&1 >/dev/null
+if test "$?" != "0"; then
+    echo esptool.py not found, attempting to install
+    python3 -m pip install --user esptool
```
MitchBradley: Instead of installing esptool with pip, how about we just include esptool.py in the bundle, to ensure that we get the version that we want.

MitchBradley:
Pursuant to your venv idea, I made this script which mostly works:
```
#!/bin/sh
python3 -m pip install -q --user --upgrade pip
# sudo apt-get install python3-venv
python3 -m pip install -q --user virtualenv
python3 -m venv env
. env/bin/activate
python3 -m pip install -q pyserial xmodem
python3 fluidterm/fluidterm.py $*
deactivate
```
The "mostly" part had to do with an error I got on an old Debian Buster installation.  The error message was:
```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt-get install python3-venv
```
I fixed it with `sudo apt-get install python2-venv` per the commented-out line in the script.

install_scripts/linux-python3/tools.sh
easytarget: That is a *very* good idea, especialy given ESP's penchant for putting breaking changes into minor revision updates..

I'll look at this properly tomorrow, the install-python3 script can be given the source zip in the same way as the other individual platform zips, and then `pip install <downloaded src>.zip` (I just tested this, and pip installing a zipfile directly works fine on the commandline)

easytarget: I just pushed #[b12c76c](https://github.com/bdring/FluidNC/pull/272/commits/b12c76cc6ddf13287bff7da6674c7a271d540449) to the request; this copies the correct source bundle to the package and tools.sh then installs from that, rather than pipy.

edit: I also dropped the `--user` flag to the pip install, it causes errors if combined with virtualenvs, it is better to let pip automagically select where to install.

easytarget:
Ok; I just modified the HOW-TO-INSTALL; I'm going to post here rather then immediately push to the PR. Feedback appreciated, It is designed to quickly give the install instructions, then guide users through troubleshooting, fluidterm etc..

My feeling on Virtualenv is to not try and script it in; but rely on documentation instead. Not just here, there is a lot of good documentation about it and it's issues on the web which we can leverage. We should expect anyone with a 'custom' python install to be willing/competent to follow those guides.

I can push this to the PR for review there quite trivially if you want.

```
## Installing FluidNC on your ESP32

### Raspberry PI desktop users (or anyone else using lxde or other 'light' desktop managers) 

Unpack the zip file with this guide into a local folder, eg Downloads.
Browse to the folder and pick the installer you want to run (see below)
Right-click on that and make sure to select 'Run in a Terminal'
- Many other linux desktops's (eg Gnome) disable this, use the commandline.

If the terminal starts correctly skip the next section and 
follow the rest of the guide.

### Quick Install using the commandline.

You must have Python3, version 3.6 or higher installed, to test run:

    python3 --version

This should be the default for any well updated and recent linux system, 
please see the notes below if you are having trouble with getting the 
correct python3 version.

Plug your ESP32 into a USB port.

In the folder where you unpacked this script run the installer script
that matches your needs:

To install the WiFi version:            sh install-wifi.sh
To install the Bluetooth version:       sh install-bt.sh
To replace the ESP32 local filesystem:  sh install-fs.sh

Replacing the local filesystem is only mandatory on your first install, 
see below for more.

### The Installer

The script should find and connect to the ESP32, if you have any trouble
with connecting or finding the correct device please start at the
troubleshooting sections below.

After the install an attempt will be made to run fluidterm, and connect
to the newly installed firmware, this is described in more detail below. 
Use Ctrl-] to exit, send $I to the firmware for version etc, see the 
documentation for further setup and configuration.

### Group membership Considerations

Your user must be a member of the `dialup` group in order to access
the USB serial device, you can use `id -a` to see your current
groups. To add yourself to the 'dialup' group you need to use:

    sudo usermod -a -G dialup <username>

This is standard for all major distributions, but may vary
eg FreeBSD uses 'dialer'

### Caveat

Please note: These instructions apply to the case where you have
downloaded a release bundle from
https://github.com/bdring/FluidNC/releases - a zip file named
fluidnc-vN.N.N-linux-python3.zip .  They do not apply to installing from
a source tree.  This HOWTO file is present in the source tree so that
it can be packed into a release bundle by automated release scripting,
but it does not tell you how to install from source.  For that, visit
https://github.com/bdring/FluidNC/wiki/FluidNC-Compiling#use-vs-code--platformio-to-compile

### Python3 required

These install scripts use Python3 and the Python version of esptool
from https://docs.espressif.com/projects/esptool/en/latest/esp32/

They are suitable for any system supporting esptool.py
eg. non-intel (arm) linux systems, such as Raspberry PI's or
legacy machines running 32bit distributions.

Most Linux distributions will already have Python3
preinstalled.  If you have problems, search the web for:
"install python3 on <OS>"
replacing <OS> with the name and version of your operating system.

The script will attempt to install esptool.py if it is not already
present, if this fails please see the esptool documentation and
forums for support.
https://docs.espressif.com/projects/esptool/en/latest/esp32/installation.html
https://www.esp32.com/viewforum.php?f=23

### Local Filesystem Considerations

Replacing the local filesystem is only useful for the wifi version.
The Bluetooth version can start with an empty local filesystem.

The disadvantage of replacing the local filesystem is that it will
overwrite any files that you might already have there, such as FluidNC
config files, WebUI preferences and macros.  The advantage is that you
will get the latest version of the index.html.gz file that contains
the WebUI code.  Another way to get a new index.html.gz is to upload
it via WebUI from wifi/index.html.gz herein.

A good approach is to use install-fs only on your first FluidNC
installation, to start from a clean slate.

### Connection problems

If esptool is having trouble connecting to the board, or programming
please start be reading the wiki section on using precompiled files:
https://github.com/bdring/FluidNC/wiki/FluidNC-Compiling#using-pre-compiled-files

Some board and programmer combinations require you to put the board
into programming mode manually, see the entry in the FAQ for this:
https://github.com/bdring/FluidNC/wiki/FluidNC-FAQ

The Wiki also has information on how to get further support if all
else fails. 

### Running Fluidterm

The FluidNC install scripts run Fluidterm automatically at the end,
but if you want to run it separately, you can type

  sh fluidterm.sh

or just

  ./fluidterm.sh

### If Fluidterm Won't Start ...

Fluidterm is intended to be helpful but it is not absolutely necessary
for using FluidNC.  Fluidterm lets you interact directly with FluidNC
over a USB serial port.  There are other ways to do that, so if you
have trouble running Fluidterm, you can just ignore it.

On headless (no GUI installed) machines fluidterm may fail with:

    ModuleNotFoundError: No module named 'tkinter'

In this case you will need to install the correct system package for tkinter;
- note that trying to install via pip/pypi is usually not sufficient.

    Debian/Raspbian: $ sudo apt install python3-tk
    Fedora/RHEL: $ sudo dnf install python3-tkinter
etc..

### Alternatives to Fluidterm

Most GCode sender programs have some way to send commands directly to
FluidNC, but it can sometimes be helpful to bypass the complexity of
a sender and use a more direct path.  Fluidterm is one such direct
path, but there are others, typically called "serial terminals".

For Linux, there are many such programs, such as "screen", "minicom", "cu",
"picocom", and "PuTTY".  The one that is most likely to be preinstalled
is named "screen".  If screen is not preinstalled, you might be able to
install it with (on Ubuntu or Debian).

   sudo apt update
   sudo apt install screen

To use screen, go to a shell window and type:

  ls /dev/tty*

That will give you a list of serial ports.  You need to find the one
that is connected to FluidNC.  It will probably have a name that starts
with "/dev/ttyUSB".  Once you have found that name, type

  screen /dev/ttyUSBWHATEVER 115200

To exit from screen, type  Control-A k

Search the web for more documentation about screen, or for instructions
for installing it on other versions of Linux.

### What Can Go Wrong?

esptool.py is only available on Python3.6 and higher, if your system
does not support that you may be able to use the python2.7 version of
esptool.py by manually installing it first:
    pip install esptool.py
The install_* scripts should then be able to complete the firmware install,
but will fail after that when they cannot run fluidterm. See above
for alternatives to fluidterm.

If Python3.6+ is available on your system, but (for whatever reason) you
use a lower Python version as the system default you can use a
python virtual environment (venv) to do the install without needing to
modify your system-wide python environment:
https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/

### FreeBSD and other *Nix *BSD platforms

This tool has been used successfully on FreeBSD 13 and should work
on any unix-like platform where Python3 and esptool.py are available.
However this is unsupported; and for advanced users only.
For reference:
FreeBSD(13):
  User had to be added to 'dialer' group, and py-tkinter installed from
  ports to enable fluidterm.
```

Longus:
I'm not qualified to comment on the technical aspects, but I did spot a couple of probable typos:

> installation, to start from a clean slate.

installation, **or if you want** to start from a clean slate.

> please start be reading the wiki section on using precompiled files:

please start **by** reading the wiki section on using precompiled files:

easytarget:
@Longus ; proofreading is also very welcome :-), I've updated the comment

easytarget:
I just updated the HOW-TO-INSTALL even more, it's not perfect, I find it hard to strike the 'TL;DR' vs 'too short;no info' balance. But I think this is what is needed.
I also pushed another quick addition to the build-release script that puts the esptool version into the readme filename, so that it is easy to check without needing to look in the source bundle.

MitchBradley:
This is starting to look really good.  I will test it soon.

easytarget:
Humm, I decided to add my HOW-TO changes, and I added change that makes the build script download esp packages to a separate folder instead of polluting the repo root, which was annoying me. Probably a bit over-optimistic. I'll rebase my branch.

MitchBradley:
Over the weekend I have been riffing on your work with the intent of reducing the number of release packages to two - one for Windows and one for everything else.  I think that the python version can eliminate the linux-amd64 version and the macos version can be combined into the python one, with the macos differences handled in the scripting.

MitchBradley:
#296 extends your work so there is no longer a need for macos and linux-amd64 installers.

easytarget:
Hi @MitchBradley ; that all looks absolutely brilliant. And I totally agree with the approach taken.

I'm sorry I did not get back on discord asap; there is quite a lot going on here and my discord details are on my laptop, which is elsewhere. Once I 'recover' that I'll pop by on Discord to say HI anyway.

I'll also take the new installer for a test drive because I still have one last thing to do: Upgrade my grbl-esp32 powered cnc..
Up till now I've just been installing to a loose esp32 devboard and testing that I can communicate. And this has been happening since I wanted to ensure Laserweb 4.1 will support FluidNC it out-of-the-box; but I also want to upgrade my machine so I'll be following the documentation, using the config upgrade system, etc, but only once LW4.1 is done.

Issue #285, opened 2022-02-09
WEB UI


sergerzet:
The results of engine operation are not yet displayed on the "CONTROLS" and "GRBL" tabs: coordinates, operating modes. V3.3.2

sergerzet:
When compiled in VS Code, the "Controls" and "GRBL" tabs work until the "Config.yaml" on the other.

sergerzet:
When downloading from Fluid C, the file "config.yaml. Other configuration files don't work. Files with stepper motor control by I2S are considered.

MitchBradley:
I do not understand any of this.  Perhaps you can write it in your native language and a human can translate it to understandable English.

sergerzet:
На вкладке "Controls" не отображаются показания перемещения двигателей, на вкладке "GRBL" режим "IDLE" не меняется на режим "RUN". Это происходит при замене файла "Config.yaml" на файл "6_Pack_stepstick_XYZABC.yaml". Остальные элементы интерфейса работают нормально. Файл "6_Pack_stepstick_XYZABC.yaml":
board: 6 Pack
name: 6 Pack StepStick XYZABC
stepping:
  engine: I2S_STREAM
  idle_ms: 250
  pulse_us: 4
  dir_delay_us: 1
  disable_delay_us: 0

axes:
  shared_stepper_disable_pin: NO_PIN
  x:
    steps_per_mm: 800.000
    max_rate_mm_per_min: 1000.000
    acceleration_mm_per_sec2: 100.000
    max_travel_mm: 300.000
    soft_limits: false
    homing:
      cycle: 2
      positive_direction: false
      mpos_mm: 150.000
      feed_mm_per_min: 100.000
      seek_mm_per_min: 200.000
      settle_ms: 500
      seek_scaler: 1.100
      feed_scaler: 1.100

    motor0:
      limit_neg_pin: gpio.33
      limit_pos_pin: NO_PIN
      limit_all_pin: NO_PIN
      hard_limits: false
      pulloff_mm:1.000
      stepstick:
        ms3_pin: i2so.3
        step_pin: I2SO.2
        direction_pin: I2SO.1
        disable_pin: I2SO.0

  y:
    steps_per_mm: 800.000
    max_rate_mm_per_min: 1000.000
    acceleration_mm_per_sec2: 100.000
    max_travel_mm: 300.000
    soft_limits: false
    homing:
      cycle: 2
      positive_direction: true
      mpos_mm: 150.000
[...]

PR #292, opened 2022-02-11
Fixed bug in previous version of index.html.gz with >4 axes

MitchBradley:


MitchBradley:
Ref #285

MitchBradley:
Please connect with a serial terminal and show us the startup messages after a reset.

PR #294, opened 2022-02-13
ESP32, XYYZ, Relay-Spindle, Ext-drivers LED, Fan config

Domush:
## Example config for those using a plain ESP32 and external drivers with a relay-operated spindle.

- XYYZ axis layout
- External driver support
- Realistic feed rates and acceleration
- PWM Fan support
- Relay driven LED support
- Relay driven spindle support
- Automatic starting of fan on boot
- Individual limit switches


bdring:
We only accept PRs. targeting the devt branch. I will fix.

see 

bdring:
Some of the rates and steps_per_mm are not realistic and a good example.

Domush:
They're my current feed rates for my CNC. Not sure what's unrealistic about them. Been running those rates for months with a Marlin firmware.

I also found it weird how some of them asked for mm/min.. even the largest CNC machines travel to max well within a minute. That number is always going to be huge.

Domush:
Sorry for the incorrect pull target. I've rebased my branch to the Devt branch, although I'm not sure how to move the pull request itself.

MitchBradley:
Many or even most new users of FluidNC are running smaller machines with steppers directly connected to ball screws.  For such machines, 30000 mm/min with a 3mm pitch ball screw would result in a stepper running at 10000 RPM which is way too fast.

It is certainly true that commercial machines with big motors and different drive systems can run at such rates.  We wish to accommodate such machines, but we do not want to make the many novices that we support think that they will be able to run at rates that are out of the question for the machines they are likely to have.

There is no indication in your configuration to let a new user know what kind of drive mechanism is necessary for such rates.  It is not true that low-end machines will not be using external drivers; we have many counterexamples.

Perhaps what is needed is more context about the machine that is targeted.  Just saying "external drivers" does not imply a big machine with particular drive characteristics.

The reason why the max rate is specified in mm/min is because that is the units that are conventionally used for GCode in Grbl.

MitchBradley:
We have moved the example config files to a [different repo](https://github.com/bdring/fluidnc-config-files) so this PR is no longer apropos to this repo.

PR #299, opened 2022-02-16
Fixed limits, jogging, and mpos setting

MitchBradley:
This is a breaking change for machine configurations.

Previously, the homing pulloff value was considered to be inside the machine travel envelope, so with positive_direction:false, mpos_mm:0, max_travel_mm:200, after homing the position is 3 and the travel ranges from 3..200

Now. the pulloff is outside the envelope, so with those same settings, after homing the position is 0 and the travel ranges from 0..200

Doing it this way simplifies various calculations that were error-prone.  We think that this new way is easier to understand.

The down side is that you will might need to update your configuration.


bdring:
Let me catch up on the wiki.

Issue #301, opened 2022-02-16
Dual functions for reset and cycle_start pins


cotepat:
A lot of simple pendants or button boxes are have "Stop" and "Start" buttons. I always hesitate on how to use those... I usually end up using the stop button as a reset button and the start button as a cycle start button... so I have to rely a lot on the UI for some basic functions (namely unlock and feedhold).

The problem is : when you hit reset, you have to go to the UI to unlock (there is no unlock_pin in fluidnc, and unlock "$X" is not configurable as a macro because macros cannot run when machine is locked). And, of course, you have to restart the job completely...

Buttons and pins are often limited on controllers / 

The idea here is to have start / stop buttons that combine reset / feedhold and unlock / cycle_start functions.

I have it working perfectly with very simple changes in the code. See associated PR...

Here is how it works:

- If "control/dual_mode_reset_pin" is true in config file, then reset_pin will act as feedhold when active state is "Cycle" or "Jog". Otherwise the original reset behaviour will be used). To completely kill a running cycle or jog, just press reset twice and state will change from Cycle/Jog->Hold->Idle. I think this is a good option because there will be a "soft stop" (feedhold) whenever possible, and a "hard stop" (reset) when required (for instance when homing or probing).

- If "control/dual_mode_cycle_start_pin" is true in config file, then cycle_start_pin will act as an unlock pin when state is Alarm. Otherwise the original behaviour of cycle_start_pin will be used). This is very useful: no need to go to the UI to unlock and Alarm, and it does so without using a precious input pin / button.

cotepat:
I found a bug in a very specific situation, I will come back with a new pull request when it's fixed.

Fcsteve777:
I like this idea, saves on IO pins

PR #302, opened 2022-02-16
Dual function reset and cycle_start pins

cotepat:
Allows dual functions for reset and cycle_start pins:

If "control/dual_mode_reset_pin" is true in config file, then reset_pin will act as feedhold when active state is "Cycle" or "Jog". Otherwise the original reset behaviour will be used). To completely kill a running cycle or jog, just press reset twice and state will change from Cycle/Jog->Hold->Idle. I think this is a good option because there will be a "soft stop" (feedhold) whenever possible, and a "hard stop" (reset) when required (for instance when homing or probing).

If "control/dual_mode_cycle_start_pin" is true in config file, then cycle_start_pin will act as an unlock pin when state is Alarm. Otherwise the original behaviour of cycle_start_pin will be used). This is very useful: no need to go to the UI to unlock and Alarm, and it does so without using a precious input pin / button.

https://github.com/bdring/FluidNC/issues/301#issue-1140223180

### DIFF
--- a/FluidNC/src/Control.cpp
+++ b/FluidNC/src/Control.cpp
@@ -22,6 +22,8 @@ void Control::init() {
 }
 
 void Control::group(Configuration::HandlerBase& handler) {
+    handler.item("enable_dual_mode_reset_pin", _dual_mode_reset); // reset button becomes a "feedhold button" if active state is "Cycle" or "Jog"
+    handler.item("enable_dual_mode_cycle_start_pin", _dual_mode_cycle_start); // cycle_start button becomes an "unlock button" if active state is "Alarm"
     handler.item("safety_door_pin", _safetyDoor._pin);
     handler.item("reset_pin", _reset._pin);
     handler.item("feed_hold_pin", _feedHold._pin);
--- a/FluidNC/src/Control.h
+++ b/FluidNC/src/Control.h
@@ -20,6 +20,10 @@ public:
     ControlPin _macro3;
 
 public:
+    // Configurable
+    bool _dual_mode_reset = false;
+    bool _dual_mode_cycle_start = false;
+    
     Control();
 
     // Initializes control pins.
--- a/FluidNC/src/Protocol.cpp
+++ b/FluidNC/src/Protocol.cpp
@@ -761,8 +761,8 @@ void protocol_exec_rt_system() {
 #endif
     if (rtReset) {
         if (config->_control->_dual_mode_reset && (sys.state == State::Cycle || sys.state == State::Jog)) {
-            rtFeedHold = true;
             rtReset = false;
+            protocol_do_feedhold();
         } else {
             if (sys.state == State::Homing) {
                 rtAlarm = ExecAlarm::HomingFailReset;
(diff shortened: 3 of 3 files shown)


FluidNC/src/Control.cpp
```diff
@@ -22,6 +22,8 @@ void Control::init() {
 }
 
 void Control::group(Configuration::HandlerBase& handler) {
+    handler.item("enable_dual_mode_reset_pin", _dual_mode_reset); // reset button becomes a "feedhold button" if active state is "Cycle" or "Jog"
+    handler.item("enable_dual_mode_cycle_start_pin", _dual_mode_cycle_start); // cycle_start button becomes an "unlock button" if active state is "Alarm"
```
MitchBradley, line 26: We need better names.  "enable_dual_mode_reset_pin" is both too wordy and insufficiently descriptive - "dual_mode" doesn't really tell you much unless you already know what it means.  Maybe "reset_does_feedhold" and "cycle_start_does_unlock".  I'm not thrilled by those names but they are better than "enable_dual_mode..."

bdring: Several people have asked for toggle style button like feed and cycle start.

The "enable" is not needed in the name. Maybe the names are the 2 functions like "feedhold_reset_pin".

We should also have a way to report these in the status report.

bdring: I suppose the reporting for a reset/feedhold pin, could just report RH when active. The reporting is just done for human eyes, typically during troubleshooting.

bdring: Sorry, I made the incorrect assumption this was defining a new pin. We have [spec for config item names](https://github.com/bdring/FluidNC/wiki/FluidNC-Config-File-Overview#names-and-units). "_pin" at the end would suggest the value should be of type Pin.

I wonder if that would be a better way to do it.

cotepat: Thank you both for your comments! By the way, I am now a sponsor of FluidNC. Very thrilled with the work you guys are doing!

@MitchBradley : Yeah... the names are pretty bad 😬... I think I like "reset_pin_does_feedhold". or "reset_pin_also_holds"? English is a second language, so this is not easy. And us french (Canadian in my case) are often wordy :-)

@bdring Very good point that it should not end with "_pin"! My first iteration was actually done this way, using 2 extra pin variables, but I felt it was a bit redundant to have both a reset_pin and a "reset_and_feedhold_pin". Also, I didn't want to add to the complexity of the realtime behaviour of "reset" (which is done via MotionControl.cpp I think while homing for instance). By using a "reset_also_holds" type of parameter, I was able to keep the changes to a minimum and just rely on existing functions and variables (rtReset, rtFeedhold...).

bdring: Thank you. 

Here is a [discord invite link](https://discord.gg/XpytXt76). If you join Discord, please tell me your users name and I will add you to the "sponsor" group.

I looked at the code and agree with the mess new pin types create in the main code.

I think a standardized way would be to use the pin name first, then the added feature.

 reset_does_feedhold
 cycle_start_does_feedhold
 cycle_start_does_reset

PR #334, opened 2022-03-13
WIP

bdring:
Add an optional "P" parameter that specifies the thickness (or offset from 0) of the probe device.

Upon successful probe the offset of the current system is zeroed with the P value applied. This will make it much easier for displays and senders to deal with probing. Most people trying to roll their own probe UI are doing a very bad job of it. This should not adversely affect any existing senders because the parameter is new and optional.

G38.2 G91 F80 Z-20 P8.00

### DIFF
--- a/FluidNC/src/NutsBolts.cpp
+++ b/FluidNC/src/NutsBolts.cpp
@@ -197,6 +197,17 @@ char* trim(char* str) {
     return str;
 }
 
+// Count the number of bits that are on.
+uint8_t popcount(uint32_t val) {  // TODO Is there a built in function for this. Tried all the usual ones.
+    uint8_t count = 0;
+    for (int i = 0; i < 32; i++) {
+        if (bitnum_is_true(val, i)) {
+            count++;
+        }
+    }
+    return count;
+}
+
 String formatBytes(uint64_t bytes) {
     if (bytes < 1024) {
         return String((uint16_t)bytes) + " B";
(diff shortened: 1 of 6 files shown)


FluidNC/src/NutsBolts.cpp
```diff
@@ -197,6 +197,17 @@ char* trim(char* str) {
     return str;
 }
 
+// Count the number of bits that are on.
```
MitchBradley: Since you really want to know if more than one bit is set, you could use this:
```
bool multiple_bits_set(uint32_t val) {
    // Takes advantage of a quirk of twos-complement math.  If a number has
    // only one bit set, for example 0b1000, then subtracting 1 will clear that
    // bit and set only other bits giving e.g. 0b0111, and anding the two gives 0.
    // If multiple bits are set, subtracting 1 will not clear the high bit.
    return val & (val-1);
}

File: FluidNC/src/NutsBolts.cpp [hunk 2]
```diff
@@ -197,6 +197,17 @@ char* trim(char* str) {
...
+uint8_t popcount(uint32_t val) {  // TODO Is there a built in function for this. Tried all the usual ones.
```
MitchBradley: Even though the bit count will certainly fit in 8 bits, it is better to use int.  Using uint8_t is inefficient on any non-8-bit processor, because to generate numerically accurate code, the compiler must emit extra instructions to mask off high bits.

It is best to reserve uint8_t for character-like data that will be packed into an array.  POG used uint8_t all over the place as an efficiency trick.  It made it hard to determine the actual usage.

For internal computation, and for passing data to and from functions, int is more appropriate because it matches the register size.

bdring: Thanks, I'll use multiple_bits_set

PR #337, opened 2022-03-15
Working version of pin extenders

atlaste:
This PR is a complete implementation for PCA pin extenders. It is tested extensively, on both unit tests and real hardware. The latency that I found on real hardware is typically between 3 and 7 us.

A few other bugs have been fixed in the process. It should also be quite feasible to make other types of [smart] extenders using this code.

At this point, I would say that making a board with a pin extender is quite hard to get right. It's not really the hardware itself that makes it hard, but how pins should be used. For the moment, I would only recommend using the pin extender for endstops, and only if the interrupt pin of the pin extender is wired. Definitely not stepping. Definitely not probing. Other I/O should work, but should not be mixed with endstops. That's hardly a real issue tho, because it frees a lot of GPIO. 

It should be relatively easy to add other I2C pin extenders using this code. From all the data sheets that I have seen, they all work the same way, just with other numbers of ports and other addresses. Nothing the code cannot handle.


bdring:
Do you have a reference schematic?

atlaste:
The datasheet has a reference schematic with all the details on how to use these IC's. I based my board off that: https://www.nxp.com/docs/en/data-sheet/PCA9539_PCA9539R.pdf pg 17. As long as you just use them for a single purpose like I do, these extenders work just fine. 

Once I finish testing these boards and fixing the bugs I found I can order a bigger batch of boards and send you one as well. I'd be happy to sell them in the EU... I can also design a little 6pack module off these for endstops in due time, it's hardly any effort.

robZeilinga:
From a very high level, would this PR be able to handle other I2C extenders ( specifically MCP23017 ) ?

krupnfs:
Hi, can you update this PR from master?

emab123:
So I was looking at the wiki and it still says pin extenders are experimental and points to this PR. Maybe update that and close this?

bdring:
[That approach](http://wiki.fluidnc.com/en/config/extenders) was not completed and was replaced with the [UART channel protocol](http://wiki.fluidnc.com/en/config/uart_sections#channel-io). 

We have [STM example code](http://wiki.fluidnc.com/en/hardware/official/STM32_IO_Extender) and the [Airedale I/O Expander](http://wiki.fluidnc.com/en/hardware/official/airedale) is available.

I think it might be best to leave this PR open to document what was done. I have updated the wiki page to point to the new method.

image

PR #350, opened 2022-03-26
Limit reiteration

MitchBradley:


bdring:
I tested the PR on ...

Dual Axis, one switch
Dual Axis 2 switches

I tests as $HY and $H (with multiaxis homing)

All work.

...like before the change.

MitchBradley:
The fix eliminated unnecessary memory usage.  I suspect that, previously, you could eventually crash the system by doing $cd enough times, until memory runs out.

atlaste:
Limit pin changes look good to me. Cant properly judge the pwm change from here on my phone...

Issue #336, opened 2022-03-15
Filename too long - Web UI wont launch - Loading...


Gerard1983:
Please only submit bugs for the **latest** version of the code. You can check the version number in the startup messages.

**Do not add the bug label**. The developers of this repo will do that after confirming the issue.

Please provide the [FluidNC boot messages per this page](https://github.com/bdring/FluidNC/wiki/Requesting-Help#fluidnc-boot-messages).

Select a COM port
0: COM1 (\Device\Serial0)
1: COM3 (\Device\Silabser0)
Choice: 1
Using COM3
Ctrl-] to exit, Ctrl-U to upload, Ctrl-R to reset
[MSG:INFO: Filename too long]
[MSG:INFO: Filename too long]
[MSG:INFO: Filename too long]
[MSG:INFO: Filename too long]
[MSG:INFO: Filename too long]
[MSG:INFO: Filename too long]
Resetting MCU
[MSG:INFO: Filename too long]
[MSG:INFO: Filename too long]
ets Jun  8 2016 00:22:57

rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0018,len:4
load:0x3fff001c,len:1044
load:0x40078000,len:10124
load:0x40080400,len:5856
entry 0x400806a8

[MSG:INFO: FluidNC v3.4.1]
[MSG:INFO: Compiled with ESP32 SDK:v3.3.5-1-g85c43024c]
[MSG:INFO: Cannot open configuration file:config.yaml]
[MSG:INFO: Using default configuration]
[MSG:INFO: Axes: using defaults]
[MSG:INFO: Machine Default (Test Drive)]
[MSG:INFO: Board None]
[MSG:INFO: SPI not defined]
[MSG:INFO: Stepping:RMT Pulse:4us Dsbl Delay:0us Dir Delay:0us Idle Delay:255ms]
[MSG:INFO: Axis count 3]
[MSG:INFO: Axis X (-1000.000,0.000)]
[MSG:INFO: Axis Y (-1000.000,0.000)]
[MSG:INFO: Axis Z (-1000.000,0.000)]
[MSG:INFO: Kinematic system: Cartesian]
[...]

PR #358, opened 2022-04-01
Fixed multiple replacement of / with /spiffs/

MitchBradley:


MitchBradley:
Root cause fix for #336

Issue #364, opened 2022-04-06
Support for a different PWM Spindle (Simple H-Bridge)


spsspssps:
I have successfully tested fluidnc on my custom CNC board https://github.com/spsspssps/spscnc My card uses an H bridge to drive the spindle, and requires 2 non simultaneous PWM signals.

Current prototype schematics (with straps):
https://github.com/spsspssps/spscnc/blob/releases/1.0/1.0.2/doc/design.description.md

The current PWMSpindle driver supports DIRECTION and OUTPUT, My driver would require OUTPUT_A, OUTPUT_B. 
When the spindle is set CW, OUTPUT_A shall output PWM, and OUTPUT_B shall output GND.
When the spindle is set CCW, OUTPUT_A shall output GND, and OUTPUT_B shall output PWM.

I started a feature branch locally, and I could deliver a pull request if this feature is approved, but I have two possible ways to implement this:

a) Extend the PWM class and add a reverse pin, depending on the direction, PWM is set on OUTPUT_A or OUTPUT_B, name the pins _output_pin and _output_b_pin (But I see the BESC controller is already inheriting from this class, and I don't want to make a mess.
b) Add a derivative class (from PWM or ONOffSpindle and call it SimpleHBridge, override several methods from the PWMSpindle class, so that these changes do not affect any other existing driver.

I can write the feature myself, but I would like to agree on how to introduce this in harmony with the rest of the project.

MitchBradley:
subclassing is probably best

MitchBradley:
The other thing I would ask is that, if your PR is accepted, you lurk on Discord and field any questions related to it, on a long-term basis.  I get tired of supporting other people's code.

spsspssps:
Thanks ,
I can agree to lurking every now and then, and to add documentation for the driver both on the source code and on the main documents. I will start with subclassing, and I will try not do deviate too much from the original classes.

PR #366, opened 2022-04-08
Added HBridge Spindle type

spsspssps:
Implementation of HBridge Spindle as discussed in  https://github.com/bdring/FluidNC/issues/364

Please bear with me, as I am not completely familiar with github PR workflow.

The implementation is very similar to the PWMSpindle implementation. The reason I did not derive from there is that PWMSpindle already has a _direction_pin which the HBridge does not use and would be confusing.

I am willing to suggest a wiki entry, but I am afraid I would need permissions for that.

### DIFF
--- a/FluidNC/src/Spindles/HBridgeSpindle.cpp
+++ b/FluidNC/src/Spindles/HBridgeSpindle.cpp
@@ -1,22 +1,12 @@
 // Copyright (c) 2022 -	Santiago Palomino
 // Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
 
-/*
-    This is a full featured TTL PWM spindle This does not include speed/power
-    compensation. Use the Laser class for that.
-*/
 #include "HBridgeSpindle.h"
 
 #include "../GCode.h"  // gc_state.modal
 #include "../Logging.h"
 #include "../Pins/LedcPin.h"
 #include <esp32-hal-ledc.h>  // ledcDetachPin
-
-// ======================= PWM ==============================
-/*
-    This gets called at startup or whenever a spindle setting changes
-    If the spindle is running it will stop and need to be restarted with M3Snnnn
-*/
 
 namespace Spindles {
     void HBridgeSpindle::init() {
@@ -95,9 +85,8 @@ namespace Spindles {
         if (state == SpindleState::Disable) {  // Halt or set spindle direction and speed.
             dev_speed = 0;
         } else {
-            // PWM: this could wreak havoc if the direction is changed without first
-            // spinning down. But it looks like the framework is stopping the motor
-            // before changing direction M4 is not accepted during M3 if M5 is not run first.
+            // The core is responsible for stopping the motor before changing
+            // direction
             if (state == SpindleState::Cw || state == SpindleState::Ccw) {}
         }
         _state = state;
--- a/FluidNC/src/Spindles/HBridgeSpindle.cpp
+++ b/FluidNC/src/Spindles/HBridgeSpindle.cpp
@@ -93,9 +93,7 @@ namespace Spindles {
         // sys.spindle_speed correctly.
         uint32_t dev_speed = mapSpeed(speed);
         if (state == SpindleState::Disable) {  // Halt or set spindle direction and speed.
-            if (_zero_speed_with_disable) {
-                dev_speed = offSpeed();
-            }
+            dev_speed = 0;
         } else {
             // PWM: this could wreak havoc if the direction is changed without first
             // spinning down. But it looks like the framework is stopping the motor
@@ -157,8 +155,8 @@ namespace Spindles {
             ledcSetDuty(_pwm_ccw_chan_num, 0);
             ledcSetDuty(_pwm_cw_chan_num, duty);
         } else {  // M5
-            ledcSetDuty(_pwm_cw_chan_num, duty);
-            ledcSetDuty(_pwm_ccw_chan_num, duty);
+            ledcSetDuty(_pwm_cw_chan_num, 0);
+            ledcSetDuty(_pwm_ccw_chan_num, 0);
         }
     }
 
--- a/FluidNC/src/Spindles/HBridgeSpindle.cpp
+++ b/FluidNC/src/Spindles/HBridgeSpindle.cpp
@@ -1,0 +1,196 @@
+// Copyright (c) 2022 -	Santiago Palomino
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+/*
+    This is a full featured TTL PWM spindle This does not include speed/power
+    compensation. Use the Laser class for that.
+*/
+#include "HBridgeSpindle.h"
+
+#include "../GCode.h"  // gc_state.modal
+#include "../Logging.h"
+#include "../Pins/LedcPin.h"
+#include <esp32-hal-ledc.h>  // ledcDetachPin
+
+// ======================= PWM ==============================
+/*
+    This gets called at startup or whenever a spindle setting changes
+    If the spindle is running it will stop and need to be restarted with M3Snnnn
+*/
+
+namespace Spindles {
+    void HBridgeSpindle::init() {
+        get_pins_and_settings();
+        setupSpeeds(_pwm_freq);
+
+        if (_output_a_pin.defined()) {
+            if (_output_a_pin.capabilities().has(Pin::Capabilities::PWM)) {
+                auto outputNative = _output_a_pin.getNative(Pin::Capabilities::PWM);
+                _pwm_a_chan_num     = ledcInit(_output_a_pin, -1, (double)_pwm_freq, _pwm_precision);
+            } else {
+                log_error(name() << " output A pin " << _output_a_pin.name().c_str() << " cannot do PWM");
+            }
+        } else {
+            log_error(name() << " output A pin not defined");
+        }
+
+        if (_output_b_pin.defined()) {
+            if (_output_b_pin.capabilities().has(Pin::Capabilities::PWM)) {
+                auto outputBNative = _output_b_pin.getNative(Pin::Capabilities::PWM);
+                _pwm_b_chan_num     = ledcInit(_output_b_pin, -1, (double)_pwm_freq, _pwm_precision);
+            } else {
+                log_error(name() << " output B pin " << _output_b_pin.name().c_str() << " cannot do PWM");
+            }
+        } else {
+            log_info(name() << " no output B pin defined (Only Clockwise support)");
+        }
+
+        _current_state    = SpindleState::Disable;
+        _current_pwm_duty = 0;
+        _enable_pin.setAttr(Pin::Attr::Output);
+
+        if (_speeds.size() == 0) {
+            // The default speed map for a PWM spindle is linear from 0=0% to 10000=100%
+            linearSpeeds(10000, 100.0f);
+        }
+        setupSpeeds(_pwm_period);
+        config_message();
+    }
+
+    // Get the GPIO from the machine definition
+    void HBridgeSpindle::get_pins_and_settings() {
+        // setup all the pins
+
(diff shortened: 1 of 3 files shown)


bdring:
Pull requests should target the devt branch.

It is hard to use Git with the wiki. Maybe you can make changes in a fork and I can copy them.

spsspssps:
Verified PWMSpindle was identical on Devt and main, recompiled and retested on HW with M3,M4,M5 commands and speeds from 0 to 100%

MitchBradley:
Just wondering: what is your use case for CCW rotation?

spsspssps:
To be honest, only because commercial spindles support it, and because I added this feature to my open source card. Not that I expect to personally need it, or that my card has grabbed much attention :) I could have saved some cost and complexity by not providing this feature, but I wanted to make some nice hardware instead of the cheapest. I think most hobby CNC owners only use one direction anyway ...

MitchBradley:
Yeah, hardly anybody has reverse-cutting bits - or needs them.

MitchBradley:
I just did some searching for use cases for left-hand-cutting mill bits (as opposed to right-cutting, left-chip-clearing which is useful for thin stock).  The only thing that made sense to me was when you have a two-spindle machine that can do both sides of the part at the same time.  To keep the forces balanced you have to counter-rotate on one side.

bdring:
I think output_a_pin and b should be cw and ccw

I don't think ccw should be optional. You could use the standard PWM in that case.

I don't think **s0_with_disable:** should be an option. If the off PWM is not 0 per the speed map, that could leave both PWMs on.

A direct switch from M3 to M4 does not work. This is not a recommended thing, but is supported in gcode.

I have some stashed fixes for this PR if everyone agrees with my notes above.

spsspssps:
Hi,

> I think output_a_pin and b should be cw and ccw

That's ok by me.
> 
> I don't think ccw should be optional. You could use the standard PWM in that case.
> 
> I don't think **s0_with_disable:** should be an option. If the off PWM is not 0 per the speed map, that could leave both PWMs on.
> 
> A direct switch from M3 to M4 does not work. This is not a recommended thing, but is supported in gcode.

I know, and I was happy about that to prevent a violent change in the driver. I think this could potentially be emulated if it is desired.

> 
> I have some stashed fixes for this PR if everyone agrees with my notes above.

bdring:
I merged my changes into this PR

bdring:
Note: I did not fully implement the suggests above. I was requesting your opinion first.

I can finish if you like, or you can do it.

FluidNC/src/Spindles/HBridgeSpindle.cpp
```diff
@@ -119,42 +122,44 @@ namespace Spindles {
...
-	  ledcSetDuty(_pwm_a_chan_num, duty);
-	}
+
+        if (_state == SpindleState::Cw) {
+            ledcSetDuty(_pwm_cw_chan_num, 0);
+            ledcSetDuty(_pwm_ccw_chan_num, duty);
+        } else if (_state == SpindleState::Ccw) {
+            ledcSetDuty(_pwm_ccw_chan_num, 0);
+            ledcSetDuty(_pwm_cw_chan_num, duty);
+        } else {  // M5
```
sps-EmLogic, line 142: I think for M5 the correct behaviour is to set PWM to 0 for both instead.

spsspssps:
> Note: I did not fully implement the suggests above. I was requesting your opinion first.
> 
> I can finish if you like, or you can do it.

You may finish, since you have already implemented most of the changes already. I can test this branch on my hardware during the weekend.

FluidNC/src/Spindles/HBridgeSpindle.h
```diff
@@ -0,0 +1,104 @@
...
+        uint8_t      _pwm_precision;  // auto calculated
+        bool         _duty_update_needed = false;
+
+        // _disable_with_zero_speed forces a disable when speed is 0
+        bool _disable_with_zero_speed = false;
+
+        // The parent class does not support a reverse pin
+        // Reverse pin is intended for HW in which two PWM outputs are needed
+        // Clock wise is achieved setting PWM on the output_pin and LOW reverse_pin.
+        // CClock wise is achieved setting PWM on the reverse pin and LOW on output_pin
```
MitchBradley: Typo in comment  "Clockwise" not "CClock wise"

FluidNC/src/Spindles/HBridgeSpindle.cpp
```diff
@@ -0,0 +1,195 @@
...
+        ledcDetachPin(_output_ccw_pin.getNative(Pin::Capabilities::PWM));
+        _output_cw_pin.setAttr(Pin::Attr::Input);
+        _output_ccw_pin.setAttr(Pin::Attr::Input);
+        _enable_pin.setAttr(Pin::Attr::Input);
+    }
+
+    // Configuration registration
+    namespace {
+        SpindleFactory::InstanceBuilder<HBridgeSpindle> registration("HBridgeSpindle");
+    }
```
MitchBradley, line 183: No other spindle name apart from "NoSpindle" includes "Spindle" as part of the name.  I think it would be better to stick to the convention.  Also affects the "name()" method and the config file example.

FluidNC/src/Spindles/HBridgeSpindle.cpp
```diff
@@ -0,0 +1,195 @@
+// Copyright (c) 2022 -	Santiago Palomino
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+/*
+    This is a full featured TTL PWM spindle This does not include speed/power
+    compensation. Use the Laser class for that.
+*/
```
MitchBradley: The comment is leftover from the PWM file and should be changed to be like the one in HBridgeSpindle.h, or deleted.

FluidNC/src/Spindles/HBridgeSpindle.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include "HBridgeSpindle.h"
+
```
MitchBradley, line 5: The "Spindle" suffix in the name is inconsistent with the other files in the Spindles directory.

FluidNC/src/Spindles/HBridgeSpindle.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include "../GCode.h"  // gc_state.modal
+#include "../Logging.h"
+#include "../Pins/LedcPin.h"
+#include <esp32-hal-ledc.h>  // ledcDetachPin
+
+// ======================= PWM ==============================
```
MitchBradley: This leftover comment line is not useful.

FluidNC/src/Spindles/HBridgeSpindle.h
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2022 -	Santiago Palomino
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+#pragma once
+
+/*
+	This is a full PWM spindle driver without speed compensation for an H Bridge controller
```
MitchBradley: What does "full" compare to?

FluidNC/src/Spindles/HBridgeSpindle.cpp
```diff
@@ -0,0 +1,195 @@
...
+            log_warn(name() << " spindle pins not defined");
+        }
+
+        // We always use mapSpeed() with the unmodified input speed so it sets
+        // sys.spindle_speed correctly.
+        uint32_t dev_speed = mapSpeed(speed);
+        if (state == SpindleState::Disable) {  // Halt or set spindle direction and speed.
+            dev_speed = 0;
+        } else {
+            // PWM: this could wreak havoc if the direction is changed without first
```
MitchBradley: What information does the "PWM:" part of this line convey?  Actually, the whole comment could probably be eliminated because it is the core code's responsibility to take care of direction changes and this module is not different from any other spindle in this respect.

FluidNC/src/Spindles/HBridgeSpindle.h
```diff
@@ -0,0 +1,104 @@
...
+   with support for two directions. The external HW has the following PINS:
+     enable_pin : optional.
+     output_a_pin : Clockwise PWM signal
+     output_b_pin : Counter Clockwise PWM signal
+     When the output A is toggling, B is set LOW, and viceversa.
+
+     Features which could be added afterwards:
+     # Soft start to prevent unnecessary current spikes on the spindle supply. The spindleDelay function seems to be a way to delay the application from using a spindle which has not accelerated yet. A similar feature could be added completing or replacing that functionality for this spindle type.
+
```
MitchBradley, line 22: Line too long.  Why the # at the beginning?  In fact, just eliminate the whole "Features which could" section because if we add soft start, it should be done more comprehensively instead of inside one driver.

FluidNC/src/Spindles/HBridgeSpindle.h
```diff
@@ -0,0 +1,104 @@
...
+        int          _pwm_cw_chan_num  = -1;
+        int          _pwm_ccw_chan_num = -1;
+        uint32_t     _pwm_period;     // how many counts in 1 period
+        uint8_t      _pwm_precision;  // auto calculated
+        bool         _duty_update_needed = false;
+
+        // _disable_with_zero_speed forces a disable when speed is 0
+        bool _disable_with_zero_speed = false;
+
+        // The parent class does not support a reverse pin
```
MitchBradley: Eliminate this line and the next because they are confusing.  "Reverse pin" is irrelevant for this hardware.  It has reversing capability, but it has no reverse pin so talking about one is confusing.

MitchBradley (commented):
Lots of picky stuff with the intention of making the comments helpful instead of misleading.

spsspssps:
I look first into the small things and rename the files afterwards as a separate commit.

FluidNC/src/Spindles/HBridgeSpindle.cpp
sps-EmLogic: I see. The file names in that directory contain the Spindle subfix, but not the classes. 

I do not have strong arguments on this, only that an HBridge is a very generic name. HBridges are also used internally in the stepper controllers. The company I work for enforces naming class files with the complete class name, to simplify finding things. But I can rename the class as part of the next commit.

MitchBradley: Just be consistent with the existing files.  I understand the value of making the file name match the class name, but if we change that, we should change it everywhere, instead of having some one way and some the other.

FluidNC/src/Spindles/HBridgeSpindle.h
sps-EmLogic: Simplifying the comment. Anyway, I think it is good to have a disclaimer as this is a limitation which is only apparent for this kind of driver. Most of other spindles types have their own power supply. So, their drivers do not need to consider soft starting. In this case, not doing soft start when using a lab power supply with current limit, makes the input voltage drop, which in my particular case triggers a safe shutdown.

FluidNC/src/Spindles/HBridgeSpindle.cpp
sps-EmLogic: No problem, implementing as suggested.

spsspssps:
I will test this momentarily and report back

spsspssps:
Hardware test concluded on SN1 and SN2 (HW Rev 1.0.2) 
All tests passed. Switching from M3 to M4 directly works as well although it generates some power issues when S10000 was typed in advanced (such issues are only a problem with the lab power supply. A soft acceleration software driver would prevent that)

Ready to integrate if there are no further comments.

bdring:
I'll check to see if sending an S value before M3 is working properly.

MitchBradley (approved):

bdring:
S before M3/M4 seems to work fine for me.

PR #381, opened 2022-04-14
Add Polar Arm Kinematics

fluxin:
Adds kinematics of or axis which is an arm of specified length which rotates for the Y Axis, and a linear X axis

image

straight_y_position: is the distance from Y0 to the center point of the rotating arm in the y axis in mm
arm_length: is the length of the arm
x: 0 should be set where the arm is parallel with the edge of the work area, (arm is extended to it's furthest)


bdring:
https://discord.gg/MpwFYgf7

FluidNC/src/Kinematics/PolarArm.h
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2020 -	Bart Dring
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+#pragma once
+
+/*
+	WallPlotter.h
```
MitchBradley: Should be PolarArm.h

Editorial comment: This is an example of why I dislike including filenames in comments.  They are frequently wrong due to failure to update them.

MitchBradley (commented):
We also need some text and images for the wiki describing the type of machine that this applies to, and how it is parameterized

PR #402, opened 2022-05-03
Fix YAML Formatting

ajquick:
I noticed a few config file examples that were not valid YAML. I have corrected the spacing on the variable _pulloff_mm_. It's not a major change and I know the config files do not need to be overly strict, but it is useful for the examples to be valid.


bdring:
I have noticed that as well. FluidNC does not care, but some editors do not colorize it correctly.

I also changed the target of your PR to the devt branch per our [PR guidelines](https://github.com/bdring/FluidNC/wiki/Pull-Request-Guidelines).

Issue #429, opened 2022-05-16
Can not compile in platformio


be-engineer:
Please only submit bugs for the **latest** version of the code. You can check the version number in the startup messages.

**Do not add the bug label**. The developers of this repo will do that after confirming the issue.

Please provide the [FluidNC boot messages per this page](https://github.com/bdring/FluidNC/wiki/Requesting-Help#fluidnc-boot-messages).

Please answer the following questions.

What version of the firmware are you using?
V3.4.4
Is the problem repeatable?
When I compile in vscode with platformio,I got some error messages:
FluidNC/test/Pins/PinOptionParsing.cpp: In static member function 'static void Pins::PinOptionParsing_NoArgs_Test::runDetail()':
```
FluidNC/test/Pins/PinOptionParsing.cpp:19:19: warning: variable 'it' set but not used [-Wunused-but-set-variable]
         for (auto it : parser) {
                   ^
FluidNC/test/Pins/GPIO.cpp: In constructor 'Pins::GPIOISR::GPIOISR(int, int, int)':      
FluidNC/test/Pins/GPIO.cpp:304:43: error: expected primary-expression before ',' token   
             gpio16.attachInterrupt<GPIOISR, &GPIOISR::HandleISR>(this, mode);
                                           ^
FluidNC/test/Pins/GPIO.cpp:304:72: warning: left operand of comma operator has no effect 
[-Wunused-value]
             gpio16.attachInterrupt<GPIOISR, &GPIOISR::HandleISR>(this, mode);
                                                                        ^
FluidNC/test/Pins/GPIO.cpp:304:76: error: invalid operands of types 'void (Pins::GPIOISR::*)()' and 'int' to binary 'operator>'
             gpio16.attachInterrupt<GPIOISR, &GPIOISR::HandleISR>(this, mode);
                                                                            ^
Compiling .pio\build\wifi\lib882\TMCStepper\source\CHOPCONF.cpp.o
*** [.pio\build\wifi\src\test\Pins\GPIO.cpp.o] Error 1
============================== [FAILED] Took 24.99 seconds ==============================
Environment    Status    Duration
-------------  --------  ------------
wifi           FAILED    00:00:24.986
========================= 1 failed, 0 succeeded in 00:00:24.986 =========================
```
I just clone project from github, without any modification. What can I do?
I tried compile Grbl_Esp32, it has no problem.

Under what conditions does the bug occur?

[...]

PR #431, opened 2022-05-16
Make it work with platformio version 6

MitchBradley:


MitchBradley:
Possible fix for #429

PR #456, opened 2022-05-28
Add possibility to perform only combination of axis homing (ex : XY or XZ or XYZ or ...)

Avataar120:
Hello,

I have added a user command to execute only XY home.
This is very useful when I don't want to change my Z or to risk any small deviation on XY when Z is moving (I'm using a laser machine with motorized bed)

Command is : $HXY

Related to this discussion : https://discord.com/channels/780079161460916227/979771214815977594

Thanks

### DIFF
--- a/FluidNC/src/ProcessSettings.cpp
+++ b/FluidNC/src/ProcessSettings.cpp
@@ -334,8 +334,43 @@ static Error home(int cycle) {
     }
     return Error::Ok;
 }
-static Error home_all(const char* value, WebUI::AuthenticationLevel auth_level, Channel& out) {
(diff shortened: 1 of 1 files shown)


MitchBradley:
Another way to do it would be to have $H1, $H2, $H3, etc to run the specified homing cycle.  The difficulty with $HXY is that it only solves one special case.  If we add $HXY, then somebody is bound to ask for $HXZ and $HYZ and, eventually, every possible combination of axes.

Yet another general solution would be to have $H parse arguments, so you could say $H=XY or $H=1 or whatever.

Avataar120:
I close this pull request as it has been integrated by Mitch in #467

Issue #482, opened 2022-06-08
Some codes suspend Bluetooth communication


tuxcnc:
### Controller Board

MKS DLC32

### Help From Board Vendor

- [ ] Yes
- [ ] No
- [ ] Not Applicable

### Machine Description

XYZ

### Configuration file

```yaml
board: MKS-DLC32
name: CNC6040
meta: 
stepping:
  engine: I2S_static
  idle_ms: 255
  pulse_us: 4
  dir_delay_us: 1
  disable_delay_us: 0

axes:
  shared_stepper_disable_pin: I2SO.0
  x:
    steps_per_mm: 200.000
    max_rate_mm_per_min: 3000.000
    acceleration_mm_per_sec2: 200.000
    max_travel_mm: 345.000
    soft_limits: true
    homing:
      cycle: 2
      mpos_mm: 0
      feed_mm_per_min: 500.000
      seek_mm_per_min: 2000.000
      positive_direction: false
    motor0:
      limit_neg_pin: NO_PIN
      limit_pos_pin: NO_PIN
      limit_all_pin: gpio.36:low
      hard_limits: false
      pulloff_mm: 1.000
      stepstick:
        step_pin: I2SO.1
        direction_pin: I2SO.2:high
        disable_pin: NO_PIN
        ms1_pin: NO_PIN
        ms2_pin: NO_PIN
        ms3_pin: NO_PIN
        reset_pin: NO_PIN

  y:
    steps_per_mm: 200.000
    max_rate_mm_per_min: 3000.000
    acceleration_mm_per_sec2: 200.000
    max_travel_mm: 550.000
    soft_limits: true
    homing:
      cycle: 2
      mpos_mm: 550
      feed_mm_per_min: 500.000
      seek_mm_per_min: 2000.000
      positive_direction: true
    motor0:
      limit_neg_pin: NO_PIN
      limit_pos_pin: NO_PIN
      limit_all_pin: gpio.35:low
      hard_limits: false
      pulloff_mm: 1.000
      stepstick:
        step_pin: I2SO.5
        direction_pin: I2SO.6:low
        disable_pin: NO_PIN
        ms1_pin: NO_PIN
        ms2_pin: NO_PIN
        ms3_pin: NO_PIN
        reset_pin: NO_PIN

  z:
    steps_per_mm: 200.000
    max_rate_mm_per_min: 3000.000
    acceleration_mm_per_sec2: 200.000
    max_travel_mm: 100.000
    soft_limits: true
[...]

PR #483, opened 2022-06-10
Queue channels

MitchBradley:
Might help with #482


tuxcnc:
I have compilled sources from QueueChannels (VS/PlatformIO).
Sorry, not works.
I mean not works at all, not only where you made changes.
For example i got this error :

ok
$X0
error:3

There is no "0" character in transmitted file, I have check it even in hex editor.
Homing sometimes works, in most cases hangs when reaches Z switch.
The USB connection not works too, although worked before changes.

MitchBradley:
How did you send the file and what are its complete contents.

I tried typing $x from FluidTerm and nothing strange happened.

MitchBradley:
The latest set of commits might work better.  I fixed several problems, including one that affects homing.

tuxcnc:
> I tried typing $x from FluidTerm and nothing strange happened.

Terminal and run program these are completely different procedures.
$X in terminal is NOT the same as $X in g-code program.
For example program is :

$X
$H
$SS
M3
G0X10
G0X0
M5
M30

Copy, paste to text editor, save as test.nc and run from some sender, I use bCNC, but it is not important which, because problem is in FluidNC.

VERY IMPORTANT INFORMATION : PROGRAM test.nc IS GOOD.
I run it without any problems, but not via Bluetooth embedded in FluidNC.

Sorry, I'm not interested in repeat many times to someone who is not listening.
How many times I told, it is a program and not run via Bluetooth, but run any other way ?

Now, I wired HC-05 (or HC-06, I'm not sure) module to CH340 chip on my controller board.
Works well.
HC-05 is transparent for data, I send it via Bluetooth, but controller receives by UART (like via USB)'
NO PROBLEMS !

The problem is in your code only.
You may not believe me that your code is bad, but it is bad, and you have to fix it some day, you won't avoid it.

For me case is over.
I will use wi-fi version with HC-05 module, I will get more and easiest.

Regards.

tuxcnc:
I have tested new configuration, with wi-fi firmware and HC-06 connected direct to 1,2,3,16 pins of CH340 chip (JDY-31 will not works because electrical parameters).
ALL works as expected.
This is serious bug in FluidNC-bt firmware, I suppose.
Maybe bad library, I don't know.
My advice is don't touch anything but Bluetooth. 
Maybe will be better and faster to write this code again, not repair existing.

Regards.

MitchBradley:
Using bt firmware compiled from this PR as of commit https://github.com/bdring/FluidNC/pull/483/commits/f046bac6a28af54143ab4428f14df8ff9df09fea , I paired Windows to the FluidNC BT, resulting in COM9 being the BT serial port.  Using TeraTerm Pro as a terminal emulator,  I put the sequence of commands you cited above in a text file and sent it without any flow control via TeraTerm's "File>Send file..." command.  TeraTerm was set for local echo to show the text as it was sent.  FluidNC had $Message/Level=Debug so the phases of homing are shown. The results were:
```
[MSG:INFO: BT Connected with FC:01:7C:58:22:56]
$X
$H
$SS
M3
G0X10
G0X0
M5
M30
$X
ok
$H
[MSG:DBG: Homing XY]
[MSG:DBG: PrePulloff]
[MSG:DBG: Fast approach]
[MSG:DBG: Pulloff0]
[MSG:DBG: Slow approach]
[MSG:DBG: Pulloff1]
ok
$SS
[MSG:INFO: FluidNC v3.4.6 (QueueChannels-f046bac-dirty)]
[MSG:INFO: Compiled with ESP32 SDK:v4.4.1-1-gb8050b365e]
[MSG:INFO: Local filesystem type is SPIFFS]
[MSG:INFO: Configuration file:t.flnc]
[MSG:DBG: Running after-parse tasks]
[MSG:DBG: Checking configuration]
[MSG:INFO: Machine 6 Pack External XYYZ (XY Swapped)]
[MSG:INFO: Board 6 Pack]
[MSG:INFO: I2SO BCK:gpio.22 WS:gpio.17 DATA:gpio.21]
[MSG:INFO: SPI SCK:gpio.18 MOSI:gpio.23 MISO:gpio.19]
[MSG:INFO: SD Card cs_pin:gpio.5 detect:NO_PIN]
[MSG:INFO: Stepping:I2S_stream Pulse:4us Dsbl Delay:0us Dir Delay:1us Idle Delay:255ms]
[MSG:INFO: Axis count 3]
[MSG:INFO: Axis X (0.000,1200.000)]
[MSG:INFO:   Motor0]
[MSG:INFO:     standard_stepper Step:I2SO.2 Dir:I2SO.1 Disable:I2SO.0]
[MSG:INFO:     Neg Limit gpio.34:low]
[MSG:INFO: Axis Y (-400.000,0.000)]
[MSG:INFO:   Motor0]
[MSG:INFO:     standard_stepper Step:I2SO.5 Dir:I2SO.4 Disable:I2SO.7]
[MSG:INFO:     Pos Limit gpio.12:low:pu]
[MSG:INFO:   Motor1]
[MSG:INFO:     standard_stepper Step:I2SO.10 Dir:I2SO.9 Disable:I2SO.8]
[MSG:INFO:     Pos Limit gpio.15:low]
[MSG:INFO: Axis Z (-128.000,0.000)]
[MSG:INFO:   Motor0]
[MSG:INFO:     standard_stepper Step:I2SO.13 Dir:I2SO.12 Disable:I2SO.15]
[MSG:INFO:     Pos Limit gpio.35:low]
[MSG:INFO: Macro 0 gpio.33:low:pu]
[MSG:INFO: Kinematic system: Cartesian]
[MSG:INFO: Using spindle NoSpindle]
[MSG:DBG: Begin Bluetooth setup]
[MSG:INFO: BT Started with FluidNC2]
ok
M3
ok
G0X10
ok
G0X0
ok
M5
ok
M30
[MSG:INFO: Program End]
ok
```
I performed the same test using the USB serial port (COM16), with identical results.

I then used Linux (WSL running under Windows) to cat the file, using
```
cat /tmp/p >/dev/ttyS9    # WSL's ttyS9 goes to Windows COM9 which is the BT serial port
cat /tmp/p >/dev/ttyS16  # WSL's ttyS16 goes to Windows COM16 which is the USB serial port
```
It worked as expected.

I tried it using FluidTerm, by pasting the sequence of commands into the window.  That worked on both ports, with the caveat that it was necessary to manually type the final Enter to execute the M30 - because Windows paste suppresses the final newline when pasting multiple lines.

I tried using FluidTerm's Ctrl-U upload command to upload files to SPIFFS.  That worked on both ports.

I tried attaching FluidTerm to the USB serial port and the sending the file via the BT port.  That worked - FluidTerm showed the debugging messages that resulted from the commands being sent on the other port.

I tried the opposite - FluidTerm on BT and sending the file via USB.  It worked.

MitchBradley:
Using the wifi build, I also tried sending the command sequence above with no flow control via telnet and websockets.  For websockets, I used the [websocat program](https://github.com/vi/websocat) as follows:
```
websocat.i686-pc-windows-gnu.exe ws://192.168.68.125:81 -n <p
```
The file "p" contains:
```
$X
$H
$SS
M3
G0X10
G0X0
M5
M30
```
In all cases, it worked correctly, performing the homing operation, proceeding to execute the following commands, and ending with a `[MSG:INFO: Program End]` report.

At this point I have no recipe to reproduce a failure.

tuxcnc:
> It worked.

Have you got a REAL MACHINE with REAL HARDWARE ?
I have got a real machine and I tested on it.

The bt firmware is working for me too - on STANDALONE,  NOT CONFIGURED AND NOT CONNECTING TO REAL HARDWARE controller.
Works because "[MSG:ERR: No homing cycles defined] ok" IS NOT a real homing.

Your code does not work on a real machine, and I explain it to you for the last time.
I soldered the cables, connected the HC-06 module, uploaded Wi-Fi firmware, everything works for me and I have more than I wanted.
This is not my problem anymore that your bt code is useless.
You don't want to - don't fix it.

MitchBradley:
Please stop shouting (CAPS) and disrespecting the developers who spend uncompensated time on this project.  Your "don't want to fix it" claim is provably false.  The existence of this PR, as well as the detailed description of the tests I have performed, disprove your claim.

The root cause of the problem was a deficiency in the code that handles lookahead for realtime characters and subsequent queuing of regular command characters.  That code was different between the UART and BT channels.  The initial version of this PR did not get the fix right but the latest version, as of the three commits ending with https://github.com/bdring/FluidNC/pull/483/commits/f046bac6a28af54143ab4428f14df8ff9df09fea, works, as tested above.

I am using a test rig with motors that turn when told to do so and pushbutton switches that I can press to trigger limits.  The test rig uses a 6-pack controller, A4988 stepsticks, and NEMA-23 motors.  It has dual motors, each with a separate limit switch, on the Y axis, so I can test both the single motor case (X) and the dual motor/dual limit case (Y).  It has an LED on the PWM spindle pin so I can see the results of M3 and M5 (although, since your code does not include an S word, the spindle is always off unless I pre-issue an S word with an appropriate nonzero value).

Homing cycle 1 includes X and Y, and the cycle works as expected. homing both axes concurrently.  The motors run until I activate the switches, then pull off, then approach again until I press the switches, then pull off again.  After the homing cycle has thus completed, the queued-up commands are then processed correctly.

I also tested with X on homing cycle 1 and Y on cycle 2.  That too worked - the X motor performed the approach/pulloff/approach/pulloff sequence according to the X limit switch, then the Y motors did the same, then the other commands executed.  It worked on all interfaces.

Your results with HC-06 are irrelevant to the question of whether or not this patch helps with the BT version.  They only prove that the serial channel works, which was true before and after the patch.  It is possible that your only point is that you have given up on trying to use the FluidNC BT features, perhaps before the latest round of fixes to this PR.  If that is the case, then your recent comments are not useful, except insofar as they suggest that you might not have actually tested the latest fixes.

tuxcnc:
> Please stop shouting (CAPS) and disrespecting the developers who spend uncompensated time on this project. 

It is very sad you work hard for free and your work not works.
Bluetooth is for data like envelope for letter.
Do you write different letter when you will use different envelope?
Did you know why HC-06 works?
Because it is transparent for data.

bdring:
Seems to work fine for me with those commands and gcodes in a file from UGS and fluidterm over BT.

Nice Work!

PR #504, opened 2022-06-21
Fix486

MitchBradley:
There are a lot of cases to test:
M3 and M4 modes.
Switching from G1 to G0 and back, with and without axis words.
I used an LED as a proxy for a laser.

### DIFF
--- a/FluidNC/src/GCode.cpp
+++ b/FluidNC/src/GCode.cpp
@@ -161,17 +161,26 @@ Error gc_execute_line(char* line, Channel& channel) {
     size_t axis_words = 0;  // XYZ tracking
     size_t ijk_words  = 0;  // IJK tracking
     // Initialize command and value words and parser flags variables.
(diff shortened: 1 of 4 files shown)


Avataar120:
Hi Mitch,

Thanks to have worked on my problem :)
I've patched it with much less lines of code 4 in one file and it's working fine for me.
Perhaps, it was not the best way to do it, but I didn't have to take several kind of movement into consideration

To Bart's question related to disable_with_s0 and , I'm not able to use it as it seems there is a real time issue on it ...
I don't know exactly how it works behind, but when I activate this option, when lightburn issues a G0 movement that therefore put PWM signal at 0, PWM goes to 0 immediately but even before end of the movement ... (clearly visible on the oscilloscope)

Resulting in not finished cut for instance.

MitchBradley:
Testing is always welcome.  We will hold off merging this change pending your tests.

Avataar120:
Working for me ! Thx again

PR #507, opened 2022-06-23
Fix FluidTerm upload

MitchBradley:
There was a bug that showed up only if there is more than one COM
port,  In that case, the next time that you issue ^U to upload a file,
when it prompts for the remote file name, it does not wait for you to
enter a name, it just acts as if you hit Enter immediately, taking the
default name.  The cause was a conflict between the use of formatted
input (std:cin >> value) in the select menu and unformatted
input (std::getline()) in the upload case.  formatted input does not
consume the newline at the end, so it is sitting there waiting when
getline() tries to read a line.


MitchBradley:
Superseded by #508

PR #508, opened 2022-06-23
Ft1.2.0

MitchBradley:


fluidterm/fluidterm.py
```diff
@@ -703,11 +712,13 @@ def writer(self):
                 for c in data:
                     if not self.alive:
                         break
+                    if c == '\x17':     # CTRL+W -> clear screen
```
brianredbeard, line 715: Why `CTRL` + `W` as opposed to `CTRL` + `L` (the default for clear screen in `man 3 readline`)?  I know that by default `miniterm` assigns `CTRL`+`L` to an EOL transformation mode, but that option is only manifested while in "menu" mode (which is set to `CTRL` + `T`) which doesn't seem work in the current code base.

brianredbeard (commented):
I can confirm that the functionality works, personally I would be slightly reticent about assigning clear screen to `CTRL`+`W` because that diverges from normative behavior in many POSIX environments.

The `readline` library used by many POSIX tools assigns the function `unix-word-rubout` (Kill the word behind point, using white space as a word boundary.) to the key combination `CTRL` + `W`.

`CTRL` + `L` on the other hand is assigned to `clear-screen` (Clear the screen, then redraw the current line, leaving the current line at the top of the screen.)

brianredbeard:
Ignore this.  I going through open Kanban cards on the Development project (https://github.com/bdring/FluidNC/projects/1).  Looks like there's quite a few cards completed still hanging out.

fluidterm/fluidterm.py
MitchBradley: I don't remember why CTRL-W was chosen but I do remember considering a lot of options.  Some time ago I stopped worrying much about readline() compatibility, simply because the number of Posix users is a small percent of all FluidNC users, and among those Posix users, only a few of those are likely to have readline keys hardwired into their fingers.  Heavy command line usage is just not as common as it way back in the day.  And I say that as a person who spent something like 30 years working for companies whose lifeblood was Unix derivatives.

PR #520, opened 2022-07-08
FileSystem class

MitchBradley:
Centralize all filesystem operations into a FileSystem class so that the code does not have to handle SD, SPIFFS, and LittleFS filesystems in different ways.


MitchBradley:
Superseded by the new scheme based on C++ std::filesystem

PR #528, opened 2022-07-15
Disable CR-adding when transferring files via XModem

breiler:
The UART channel have a feature for adding extra CR on line feeds enabled, which will break the XModem protocol. This fix will disable the extra CR adding and restore the setting after the file is sent.

It would be more sensible to toggle this in xmodem.cpp if this would be used in other places. But that function has multiple returns which makes it really hard and would require a bit of refactoring:
https://github.com/bdring/FluidNC/blob/main/FluidNC/src/xmodem.cpp#L243

### DIFF
--- a/FluidNC/src/ProcessSettings.cpp
+++ b/FluidNC/src/ProcessSettings.cpp
@@ -644,10 +644,10 @@ static Error xmodem_send(const char* value, WebUI::AuthenticationLevel auth_leve
         log_info("Cannot open " << value);
         return Error::DownloadFailed;
     }
-    bool prevAddCr = out.setCr(false);
+    bool oldCr = out.setCr(false);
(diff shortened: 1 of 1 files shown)


FluidNC/src/ProcessSettings.cpp
```diff
@@ -644,8 +644,10 @@ static Error xmodem_send(const char* value, WebUI::AuthenticationLevel auth_leve
         log_info("Cannot open " << value);
         return Error::DownloadFailed;
     }
+    bool prevAddCr = out.setCr(false);
```
MitchBradley: How about changing the name to oldCr so it is the same as the corresponding code in xmodem_receive() ?  With that change I will merge the change.  Thanks for debugging it.

PR #529, opened 2022-07-16
Allow to query the build info even if the controller is in an alarm state

breiler:
When connecting to a controller we need to know that it is a FluidNC machine and which version, and the best way is (as far as I know) to query the build info. 

If the controller is in an alarm state this information can't be queried, making it impossible to know what controller we have connected to. I propose to make this command available in any state.

### DIFF
--- a/FluidNC/src/ProcessSettings.cpp
+++ b/FluidNC/src/ProcessSettings.cpp
@@ -736,7 +736,7 @@ void make_user_commands() {
     new UserCommand("HC", "Home/C", home_c, notIdleOrAlarm);
 
     new UserCommand("SLP", "System/Sleep", go_to_sleep, notIdleOrAlarm);
-    new UserCommand("I", "Build/Info", get_report_build_info, notIdleOrAlarm);
+    new UserCommand("I", "Build/Info", get_report_build_info, anyState);
     new UserCommand("N", "GCode/StartupLines", report_startup_lines, notIdleOrAlarm);
     new UserCommand("RST", "Settings/Restore", restore_settings, notIdleOrAlarm, WA);
 


MitchBradley:
I'm not sure this patch does what the title suggests.  It is already possible to run $I in alarm state.  "notIdleOrAlarm" means the command is rejected if the state is neither Idle nor Alarm (nor ConfigAlarm).  As proof:
```
[MSG:INFO: Telnet started on port 23]

Grbl 3.5 [FluidNC v3.5.0 (wifi) '$' for help]
[MSG:INFO: '$H'|'$X' to unlock]
<Alarm|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>
$i
[VER:3.5 FluidNC v3.5.0:]
[OPT:PHS]
[MSG: Machine: Stringer]
[MSG: Mode=AP:SSID=FluidNC:IP=192.168.0.1:MAC=7C-87-CE-12-98-51]
ok
```
$I will not be accepted in states like Cycle, Jog, Homing, SafetyDoor, Hold, Sleep.  I am not sure that restriction is still valuable, but neither am I sure that opening it up to those states actually solves the problem that you cite, since it already works in Alarm state.

Is there a specific circumstance where it failed?  There are some "states" that are not really states in the protocol engine sense, but instead hard loops that wait for a specific event, usually exited only by a realtime character instead of a line-oriented command.  If the system is stuck in such a loop, opening up the permissions on $I will not help, because lines are not processed at all in such loops.  One such loop is the special kind of alarm that is caused by a hard limit, which can only be exited by a CTRL-X reset.

breiler:
Hmm, no you are right this change is no good. It is odd that when I tried it with UGS it worked, once, but not now.

Yes, there is a edge case where if the limit pin is triggered no more input is allowed. If I connect with a sender at this state it will not know what to do. This state is a bit inconsistent as some alarms allow $I but other doesn't. Would it be ok to change the behaviour of the "wait-for-reset-event-loop" so that it also allow input commands, or are we limited to real time commands?

```
Grbl 3.5 [FluidNC v3.5.0 (wifi) '$' for help]
[MSG:INFO: '$H'|'$X' to unlock]
?
<Alarm|MPos:0.000,0.000,0.000|FS:0,0|WCO:106.280,292.987,-3.000>
$I
[VER:3.5 FluidNC v3.5.0:]
[OPT:MPHS]
[MSG: Machine: ESP32 Dev Controller V4]
[MSG: Mode=AP:SSID=FluidNC:IP=192.168.0.1:MAC=FC-F5-C4-54-CA-31Mode=AP:SSID=FluidNC:IP=192.168.0.1:MAC=FC-F5-C4-54-CA-31Mode=AP:SSID=FluidNC:IP=192.168.0.1:MAC=FC-F5-C4-54-CA-31Mode=AP:SSID=FluidNC:IP=192.168.0.1:MAC=FC-F5-C4-54-CA-31]
ok
$X
[MSG:INFO: Caution: Unlocked]
ok
ALARM:1        <----- I triggered a limit switch pin
[MSG:INFO: Reset to continue]
?                     <----- Some real time commands are allowed but not $BI
<Alarm|MPos:0.000,0.000,0.000|FS:0,0|Ov:100,100,100>
```

daxliniere:
@breiler, I can confirm that UGS 2202-07-07 will _usually_ connect, whereas 2022-07-15 never will.

MitchBradley:
> Would it be ok to change the behaviour of the "wait-for-reset-event-loop" so that it also allow input commands, or are we limited to real time commands?

I wanted to do that but Bart discovered some edge cases and we deferred it until we can sort them all out.

What I really want to do is get rid of every such loop and make the system behave like a proper single-entry event driven state machine.  I have made a lot of progress in that direction but I keep getting sidetracked.

daxliniere:
Yeah, I can't even reset the controller using the FNC protocol:

```
*** Connecting to jserialcomm://COM5:115200
*** Fetching device status
[verbose] > ?
[verbose] <Alarm|MPos:0.000,0.000,0.000|FS:0,0>
[verbose] <Alarm|MPos:0.000,0.000,0.000|FS:0,0>
*** Fetching device firmware version
[verbose] > $Build/Info
*** Could not establish connection with the controller
*** Connection closed
```


If I connect with GRBL protocol:
```
**** Connected to COM5 @ 115200 baud ****
[verbose] <Alarm|MPos:0.000,0.000,0.000|FS:0,0>
[Error] Error while processing response <ok>: An unexpected command was completed by the controller.
Grbl 3.5 [FluidNC v3.5.0 (wifi) '$' for help]
>>> $$
[MSG:INFO: '$H'|'$X' to unlock]
$10 = 1    (Status report options, mask)
ok
>>> $G
[verbose] GrblFeedbackMessage{message='[GC:G0 G54 G17 G21 G90 G94 M5 M9 T0 F0 S0]', distanceMode='G90', units='G21'}
[GC:G0 G54 G17 G21 G90 G94 M5 M9 T0 F0 S0]
ok
[verbose] <Alarm|MPos:0.000,0.000,0.000|FS:0,0|WCO:7.051,4.064,-29.890>
[verbose] <Alarm|MPos:0.000,0.000,0.000|FS:0,0|Ov:100,100,100>
[verbose] <Alarm|MPos:0.000,0.000,0.000|FS:0,0>
```

By the way, that 'alarm' is a soft limit trigger that I've been unable to clear (without power cycling the ESP32 or using CTRL+R in FluidTerm).

breiler:
@MitchBradley that sounds like a great idea. I tried the same scenario on GRBL and the hard limit alarm causing the same 
problem with the controller hanging in an endless loop. But in GRBL this is not a problem as we always reset the controller on startup with a CTRL+X.

I think that by breaking out those loops you would greatly improve the protocol and is (at least what I think) needed in order to implement a reliable sender implementation.

In the meantime I'm thinking of two workarounds for this implementation in UGS:
1. Always send CTRL+X after connecting in the same way as the old GRBL-protocol
2. Relax the initialisation protocol and make it semi-connect on Alarms that will enable buttons to be able to clear alarms. 

The problem with option number two is that on alarms we normally enable homing, soft reset and unlock. In this special alarm state neither unlock or homing will do anything. A soft reset is the only option but will might lead to a new alarm state after the reset (the one with the message `[MSG:INFO: '$H'|'$X' to unlock]`). In this alarm state the home and unlock can be used.

I'm leaning towards option 1 as this will always turn the controller to a known state. 

I'll experiment a bit to see what feels more usable, thanks for your time.

MitchBradley:
One concern with always issuing ^X is that it will kill a job in progress.  What about starting with '?', then look at the status report to ensure the system is not running?  That should work for all Grbl variants and Smoothieware too.  G2core with text mode enabled at compile time (the default) will accept it as a status report request too - but G2core requires a newline before it processes anything.

So, the safe approach, I think, is to send `?\r\n` .  Grbl variants will response with a `<..>` status report followed by an `ok` line triggered by the newline.  G2Core will respond with a JSON status report like
 `{"r":{"sr":{"line":0,"posx":0,"posy":0,"posz":0,"feed":0,"vel":0,"unit":1,"coor":1,"dist":0,"stat":1}},"f":[1,0,2]}`

It might be even more robust to first send `\r\n` to clear out any line junk, and to ensure that the ? is on the beginning of the line for G2core.  If there is any junk, you might get an `error` response.

All of this seems a bit complicated, but I cannot think of an approach that is better.  ? is always safe and generally reliable.  If you do get a response back, it tells you some important information like is the protocol basically Grbl, and whether it is safe to proceed without interrupting something.  ^x is sort of like hitting it with a hammer.

breiler:
Good idea, it already sends a `?` to begin with to check if the controller is booting, it already need  special handling for HOLD or DOOR state as they don't respond to `$Build/Info` either. To make this work with an ALARM state as well should not be a problem.

Sure, this is a bit more complicated than I would like it to be and already requires six steps. There is a lot more that could go wrong than a missing CR in welcome response if you know what I mean. 😅 
But we can try it and see how it turns out.

1. Get current state of the controller (using ?), on hold/door reset the controller. Retry 10 times to make sure controller has properly booted.
2. **If state is ALARM reset the controller and get current state of the controller (using ?). Retry 10 times to make sure controller has properly booted.**
3. Get the build info ($Build/Info)
4. Get the firmware settings ($Config/Dump)
5. Get error codes ($Errors/List)
6. Get alarm codes ($Alarms/List)
7. Get the parser state of the controller ($GCode/Modes)

MitchBradley:
Ideally you could auto-detect the firmware type.  ? then $I can get you a long way.

MitchBradley:
I am an old master at things going wrong, so yes, I totally know what you mean.

PR #530, opened 2022-07-16
Fix for OpenBuilds-CONTROL.  LF -> CR-LF was inconsistent.

MitchBradley:


MitchBradley:
The LF -> CR LF translation failed in the case where a bare LF was sent all by itself in a single write().  The OpenBuilds-CONTROL version detection code failed to wake up on lines ending only with LF, requiring a full "\r\n" sequence to call it a line.  The code that reports the Grbl version happened to trip over the bare LF problem so OpenBuilds-CONTROL did not see the "Grbl version ..." line, regardless of what you put in "version ...".

PR #531, opened 2022-07-19
VESC Spindle via UART

LukasGossmann:
This PR adds support for connecting a VESC to FluidNC via UART and using it as a spindle.
It also checks for any faults occuring in the motor controller and stops gcode execution + sets an alarm.
VESC can be controlled with 3 modes:

- Speed: Set the ERPM and VESC will keep it constant with its PID controller
- Duty: Sets to motor duty cycle. Rough speed control. Speed drops under load.
- Current: Sets the current through the motor. No speed control! Needs a constant load on the motor otherwise the motor will "run away". More or less constant torque.

https://vesc-project.com/node/1796
https://electric-skateboard.builders/t/vesc-faq-duty-cycle-vs-current-control-ppm-settings/1218

Can be configured as follows (speed mode):
```yaml
VESC:
  uart:
    txd_pin: gpio.17
    rxd_pin: gpio.4
    rts_pin: gpio.16
    baud: 115200
    mode: 8N1
  tool_num: 0
  spinup_ms: 1000
  spindown_ms: 1000
  speed_map: 0=0% 0=20% 10000=20% 50000=100%
  control_type: Speed
```

Modes are called like this:  
```yaml
control_type: Speed
control_type: Duty
control_type: Current
```


FluidNC/src/xmodem.cpp
```diff
@@ -50,35 +51,6 @@ static void _outbytes(uint8_t* buf, size_t len) {
     serialPort->write(buf, len);
 }
 
-/* CRC16 implementation acording to CCITT standards */
```
LukasGossmann, line 53: In case anyone wonders why i decided to extract this to its own file:
I didn't want to add a redundant implementation for CRC16 if there is already a perfectly working copy available.

bdring:
- You will need to provide an update to the [wiki page](https://github.com/bdring/FluidNC/wiki/Fluid-Spindle-Setup). You can attach a file(s) in a reply.
- You need to provide author information per the [pull request guidelines](https://github.com/bdring/FluidNC/wiki/Pull-Request-Guidelines).

FluidNC/src/xmodem.cpp
MitchBradley: Instead of just calling it "crc.cpp", how about identifying the variant, i.e. "crc16_ccitt.cpp"

FluidNC/src/xmodem.cpp
```diff
@@ -129,13 +101,13 @@ static void flushinput(void) {
 static bool    held = false;
 static uint8_t held_packet[1024];
 static void    flush_packet(size_t packet_len, size_t& total_len) {
-    if (held) {
-        // Remove trailing ctrl-z's on the final packet
+       if (held) {
```
MitchBradley: Please undo the formatting breakage.  Run clang-format on everything that you touch.

FluidNC/src/xmodem.cpp
LukasGossmann: Done

FluidNC/src/xmodem.cpp
LukasGossmann: Should be fixed now. For some reason the autoformat did that... Had to disable it to be able to save the file without it changing that :)

MitchBradley: Maybe your autoformat is not using clang-format?  If it is using clang-format, then the .clang-format rules in the repo should, in principle, govern the styling.

LukasGossmann: I think that's what I'm using as visual studio code should pick up on that automatically as far as i can tell. All i did was open the project in visual studio code with platformio and some other add-ons installed. As far as i know i haven't done any other formatting configurations. Maybe one of the add-ons is interfering with that somehow... I need to check that next time.

MitchBradley: Maybe this will help <https://code.visualstudio.com/docs/cpp/cpp-ide> .  In principle, if you have the C++ extensions enabled, you get clang-format, but maybe you have some other extensions that use a different formatter, or there is some other setting override.  Vscode is pretty much infinitely customizable, which means that no two people have a similar experience.

bdring:
Would it be better to add a poles parameter to the config file, so the firmware and speed_map can use real RPM?

LukasGossmann:
That would certainly be a possibility however the vesc_tool only shows ERPM too so it would add a point of confusion there instead. Would renaming the RPM mode to ERPM be an acceptable solution ?

bdring:
Would M3S10000 be referring to ERPM in your case? That seems like a bad idea, especially for use with CAM software that uses real RPM.

LukasGossmann:
Yes that would be ERPM. So i think you are right. Adding that parameter would be safer.
Ill add that tomorrow :)

bdring:
Poles is similar to flutes on a bit. You tell the CAM software about the flutes and it then can work in RPM.

bdring:
OK, I guess the rough analogy is lost on you, but we agree on the required changes to the code.

LukasGossmann:
Yes sorry about that :)
The reason i don't agree with that analogy is that if I'd have a 2 flute endmill in the spindle and the motor has 7 pole pairs and then i do an M3 S1000 the spindle would spin at 1000RPM.
If i were to swap out the endmill for a 4 flute one the spindle would still be spinning at 1000RPM for an M3 S1000.
Now imagine i swapped out the motor for a different one with 14 pole pairs. That M3 S1000 would then result in only 500RPM instead of the expected 1000.

Anyway long story short. I'll add that option and a description of how to determine the number of poles for a given motor to properly configure said option.

LukasGossmann:
Updated documentation
[vesc_spindle.md](https://github.com/bdring/FluidNC/files/9147587/vesc_spindle.md)
n

LukasGossmann:
Noticed some issues with the queue filling up due to repeated calls to setSpeedfromISR

LukasGossmann:
Added a workaround similar to what is done in VFDSpindle.
Please have a look at this thread on discord as this workaround seems like its the wrong way to do it.
https://discord.com/channels/780079161460916227/1001091533333856357

PR #533, opened 2022-07-23
Motor position

atlaste:
Small PR, that stores and is able to retrieve the motor position. I assume PinExtender is taken for the PR here.

It's useful to keep track of the motor position at all times. After homing, the difference in absolute motor position should be pretty much zero - or a step has been lost. This is useful for f.ex. acceleration and max speed tuning.

This PR tracks the motor position and makes it accessible through $MP.


MitchBradley:
How about cherry picking fb56fff onto Devt and PR'ing that, so it really is a "small PR".  It will take us awhile to swallow the pin extender work.  It has to wait until I resolve the problems that keep cropping up with the switchover to the new framework.

PR #534, opened 2022-07-23
Proportional acceleration

atlaste:
This PR is branched on top of the pinextender branch.

Added acceleration that is proportional to the requested feed rate. This should ensure that you can set high accelerations for rapids without compromising high torque jobs. Without it, you have to nerf the acceleration for your G0 moves because otherwise the G1 moves would stall.

This is a setting per axis, and it's off by default.

usage:

```yaml
proportional_acceleration: true
min_acceleration_mm_per_sec2: 25
```


MitchBradley:
Same comment as for 533.  I would prefer to keep topics separate, instead of piling stuff on top of pin extenders.

PR #538, opened 2022-07-24
Fixed mist cooling status

LukasGossmann:
As the title suggests this is a small fix for the status of mist cooling.

Config:
Mist pin is defined in config

Behavior before this fix:
M7 would report correctly
M9 would report correctly
M3 SXXX and M4 SXXX would report mist cooling being enabled even when it isnt

Behavior after this fix:
M7 would report correctly
M9 would report correctly
M3 SXXX and M4 SXXX no longer report mist cooling being enabled

What was happening is that the existence of a mist cooling pin was reported instead of the actual status.

### DIFF
--- a/FluidNC/src/Report.cpp
+++ b/FluidNC/src/Report.cpp
@@ -433,7 +433,7 @@ void report_build_info(const char* line, Print& channel) {
     channel << "[VER:" << grbl_version << " FluidNC " << git_info << ":" << line << "]\n";
     channel << "[OPT:";
     if (config->_coolant->hasMist()) {
-        channel << "M";  // TODO Need to deal with M8...it could be disabled
+        channel << "M";
     }
     channel << "PH";
     if (ALLOW_FEED_OVERRIDE_DURING_PROBE_CYCLES) {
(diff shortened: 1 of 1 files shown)


FluidNC/src/Report.cpp
```diff
@@ -433,7 +433,7 @@ void report_build_info(const char* line, Print& channel) {
     channel << "[VER:" << grbl_version << " FluidNC " << git_info << ":" << line << "]\n";
     channel << "[OPT:";
     if (config->_coolant->hasMist()) {
-        channel << "M";  // TODO Need to deal with M8...it could be disabled
```
LukasGossmann, line 436: The status of the cooling doesn't matter since OPT: reports compile time (or in this case runtime config) options and not the actual status.

bdring:
@LukasGossmann Do you have a special need for an OPT response? The concept does not fit FluidNC very well and we are thinking about dropping the command entirely.

FluidNC does not really have many compile time options. All of the info in OPT (and more) is available elsewhere. Why show mist and not flood?

MitchBradley:
To be more precise, we are thinking about dropping the OPT response to the build info command.  The command would still exist, and would return only the VER line.

LukasGossmann:
@bdring No i don't have any need for that as long as UGS Platform works without that :)
I just went ahead and removed that comment since it was a TODO for something that is not part of the grbl interface spec and i was modifying that file anyway so might as well.

PR #539, opened 2022-07-24
Fix i2 s stream homing

Avataar120:
Fix I2S_Stream issue
Add XYA config file
Revert to I2S_STREAM in all YAML


Avataar120:
Sry mistake

PR #556, opened 2022-08-05
Fixed problem with held packet length causing corrupt files when using Xmodem

breiler:
While implementing XModem support I found that some files got corrupt missing a couple of bytes.

After some debugging I found a scenario when transmitting a larger file that data from packet 4 was missing 896 bytes. This is sample log of a transmission session from the sender transmitting a file with 4862 bytes:

```
T: Using Xmodem CRC/1K
T: Building Packet No. 1 (payload is 1024 bytes)
T: Building Packet No. 2 (payload is 1024 bytes)
T: Building Packet No. 3 (payload is 1024 bytes)
T: Building Packet No. 4 (payload is 1024 bytes) <-- this packet will be missing 896 bytes of 1024 bytes.
T: Building Packet No. 5 (payload is 128 bytes)
T: Building Packet No. 6 (payload is 128 bytes)
T: Building Packet No. 7 (payload is 128 bytes)
T: Building Packet No. 8 (payload is 128 bytes)
T: Building Packet No. 9 (payload is 128 bytes)
```

According to the xmodem-spec a packet may be 128 or 1024 bytes and can be switched at any time. In order to flush the packet we need to store the received packet length that is held until the next flush.


breiler:
As a side note there seems to be yet another problem with to extra bytes at the end of the file which I will look further into.

The last `1a 00` (CTRL-Z and NULL) in this hexdum is added when transmitting a file using XModem, From what I can tell we are not sending those from the sender so I assume it is added by FluidNC somehow.

```
0001290 73 30 5f 77 69 74 68 5f 64 69 73 61 62 6c 65 3a
00012a0 20 74 72 75 65 0a 20 20 73 70 69 6e 75 70 5f 6d
00012b0 73 3a 20 31 30 30 30 0a 20 20 73 70 69 6e 64 6f
00012c0 77 6e 5f 6d 73 3a 20 31 30 30 30 0a 20 20 74 6f
00012d0 6f 6c 5f 6e 75 6d 3a 20 30 0a 20 20 73 70 65 65
00012e0 64 5f 6d 61 70 3a 20 30 3d 30 2e 30 30 30 25 20
00012f0 31 30 30 30 3d 31 30 30 2e 30 30 30 25 0a 1a 00
0001300
```

MitchBradley:
Xmodem cannot send arbitrary length files; they must be aligned to a packet size boundary.  The typical way of dealing with that is to pad the last packet with CTRL-Z characters, CTRL-Z being EOF on old CPM and DOS systems.  I'll investigate this further tomorrow.

MitchBradley:
I did not realize that it is possible to switch packet sizes in the middle of a session.  I have always seen it where the sender chose a size and stuck with it.

breiler:
You probably found this already, but as you wrote - it should be able to handle both STX and SOH.

http://pauillac.inria.fr/~doligez/zmodem/ymodem.txt

    4.3  XMODEM-1k 1024 Byte Block

    An STX (02) replaces the SOH (01) at the beginning of the transmitted
    block to notify the receiver of the longer block length.  The transmitted
    block contains 1024 bytes of data.  The receiver should be able to accept
    any mixture of 128 and 1024 byte blocks.  The block number (in the second
    and third bytes of the block) is incremented by one for each block
    regardless of the block length.

MitchBradley:
I like your changes.  I retargeted the PR from main to Devt.  We do everything on Devt then promote that to main in one go when we do a release.

MitchBradley:
I think it is best to avoid 1024 byte packets in the sender->FluidNC direction.  The ESP32 UART has a 128-byte hardware FIFO.  We use that directly, instead of invoking an addition interrupt-driven layer to move from the FIFO to a software queue, thus avoiding possible synchronization and locking issues. There was a problem once where a framework update botched the synchronization, which lead to hard-to-diagnose random character loss.  With 1024-byte packets, the xmodem protocol loop must keep the FIFO drained, which is difficult to guarantee when there is network activity.  With 128-byte packets, there is less chance of overrun.  We may want to just reject STX.

MitchBradley:
In the FluidNC->sender direction, 1024 is probably okay.  Host machines have larger buffers.

PR #575, opened 2022-08-20
add reportCommsCheck

grittershub:
Replaces TMC2208Driver::test(), TMC2209Driver::test() and TMC5160Driver::test() with new code to verify the communications path is functional. This is done as suggested by @MitchBradley by reading IFCNT, resetting GSTAT and then rereading IFCNT and detecting the change.

Adds a reportCommsCheckfunction to TrinamicBase to report the outcome of the new test function.


FluidNC/src/Motors/TMC2208Driver.cpp
```diff
@@ -64,7 +64,19 @@ namespace MotorDrivers {
         }
     }
 
-    bool TMC2208Driver::test() { return TrinamicBase::reportTest(tmc2208->test_connection()); }
+    bool TMC2208Driver::test() { 
+        uint16_t ifcnt_before = tmc2208->IFCNT();
+        tmc2208->GSTAT(0); // clear GSTAT to increase ifcnt
+        uint16_t ifcnt_after = tmc2208->IFCNT();
+        if(ifcnt_after != ifcnt_before) // Check if ifcnt has changed, if so we're communicating ok. ifcnt wraps so can't use a >
```
MitchBradley: This would be better expressed as 
```
return TrinamicBase::reportCommsCheck(ifcnt_after != ifcnt_before);
```

FluidNC/src/Motors/TMC2209Driver.cpp
```diff
@@ -102,7 +102,19 @@ namespace MotorDrivers {
         }
     }
 
-    bool TMC2209Driver::test() { return TrinamicBase::reportTest(tmc2209->test_connection()); }
+    bool TMC2209Driver::test() { 
+        uint16_t ifcnt_before = tmc2209->IFCNT();
+        tmc2209->GSTAT(0); // clear GSTAT to increase ifcnt
+        uint16_t ifcnt_after = tmc2209->IFCNT();
+        if(ifcnt_after != ifcnt_before) // Check if ifcnt has changed, if so we're communicating ok. ifcnt wraps so can't use a >
```
MitchBradley: Ditto

FluidNC/src/Motors/TMC5160Driver.cpp
```diff
@@ -29,7 +29,19 @@ namespace MotorDrivers {
         TrinamicBase::config_motor();
     }
 
-    bool TMC5160Driver::test() { return TrinamicBase::reportTest(tmc5160->test_connection()); }
+    bool TMC5160Driver::test() { 
+        uint16_t ifcnt_before = tmc5160->IFCNT();
+        tmc5160->GSTAT(0); // clear GSTAT to increase ifcnt
+        uint16_t ifcnt_after = tmc5160->IFCNT();
+        if(ifcnt_after != ifcnt_before) // Check if ifcnt has changed, if so we're communicating ok. ifcnt wraps so can't use a >
```
MitchBradley: Ditto

FluidNC/src/Motors/TrinamicBase.cpp
```diff
@@ -120,6 +120,19 @@ namespace MotorDrivers {
         }
     }
 
+    bool TrinamicBase::reportCommsCheck(bool result) {
```
MitchBradley: In general, please conform to the codebase's existing formatting conventions.  The approved way to do that is to run clang-format on the files that have changed, with the .clang-format ruleset that is at the top of the tree.

In this case, a better expression would be:
```
log_info(axisName() << " communications check " << (result ? "passed" : "failed"));
return result.
```
But it would be better still to have report CommsCheck() return void instead of propagating the bool through the function.  I realize that the propagation leads to a tighter expression in ::test(), but at the expense of the reader having to wonder how reportCommsCheck might possibly compute a new value, when in fact it does not change the value.  With the void return, ::test() would look like:
```
bool okay = ifcnt_after != ifcnt_before;
reportCommsCheck(okay);
return okay;
```
but wait, I'm still not done nit-picking.  The check would be more precise as follows:
```
bool okay = ((ifcnt_before +1) & 0xff) == ifcnt_after;
```
and one final thing:  We generally avoid displaying "success" messages in startup logs because too much verbiage makes it hard to find the interesting stuff.  The existing messages typically contain extra information like pin numbers, setting values, etc.  So it would be better to issue a message only on failure.  reportCommsCheck(bool) would thus become reportCommsFailure(void).

nigel-bmlt:
Updates complete, ready for review.

grittershub:
Updates complete, ready for review.

FluidNC/src/Motors/TMC5160Driver.cpp
```diff
@@ -29,7 +29,17 @@ namespace MotorDrivers {
         TrinamicBase::config_motor();
     }
 
-    bool TMC5160Driver::test() { return TrinamicBase::reportTest(tmc5160->test_connection()); }
+    bool TMC5160Driver::test() {
```
MitchBradley, line 32: From https://www.trinamic.com/fileadmin/assets/Products/ICs_Documents/TMC5160A_Datasheet_Rev1.14.pdf  page 33:
image

We only support SPI mode for TMC5160s so I don't think this particular instance of the change will work.

MitchBradley:
Changes applied via #595.  In the future, issuing PRs against Devt, not main, will avoid some problems with the workflow.

PR #582, opened 2022-08-22
fix small bug : logical conjunction is always false in Setting

Priyanka328:


MitchBradley:
This issue was already fixed by #559.

PR #624, opened 2022-09-21
Support LittleFS and use C++ std::filesystem as the basis for all file operations

MitchBradley:
Supersedes #605


MitchBradley:
This branch is a total mistake

PR #627, opened 2022-09-23
Dynamixel msg fix

bdring:
- The bulk message was using the wrong message length which caused it not work right. Sometimes it got 2 out of 3 servos to move and sometimes none of them moved.
- Changed stepping idle ms to 32 bit so a longer delay could be used. This helps closed loop servos settle at the right location.
- Allow longer idle delays

### DIFF
--- a/FluidNC/src/Motors/Dynamixel2.cpp
+++ b/FluidNC/src/Motors/Dynamixel2.cpp
@@ -77,7 +77,7 @@ namespace MotorDrivers {
     }
 
     void Dynamixel2::config_message() {
-        log_info("    " << name() << " id::" << _id << " Count(" << _countMin << "," << _countMax << ")");
+        log_info("    " << name() << " id:" << _id << " Count(" << _countMin << "," << _countMax << ")");
     }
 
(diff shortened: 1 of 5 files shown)


FluidNC/src/Motors/Dynamixel2.cpp
```diff
@@ -70,11 +74,11 @@ namespace MotorDrivers {
         LED_on(true);
         vTaskDelay(100);
         LED_on(false);
-
-        startUpdateTask(_timer_ms);
     }
 
-    void Dynamixel2::config_message() { log_info("    " << name() << " id::" << _id << " Count(" << _countMin << "," << _countMax << ")"); }
+    void Dynamixel2::config_message() {
+        log_info("    " << name() << " id::" << _id << " Count(" << _countMin << "," << _countMax << ")");
```
MitchBradley: Is id:: a typo?  Should it be just id:  ?

bdring: yes

MitchBradley (approved):

PR #707, opened 2022-11-19
Add 'P' to $Limits printout when probe is active

nickmayer:
The Probe pin is very similar to the axis limit pins and it would be nice to be able to easily see the status in $Limits (really useful when reconnecting the machine to validate everything is correct).


MitchBradley:
Please submit PRs against Devt not main.

PR #731, opened 2022-11-29
Add 'P' to $Limits printout when probe is active

daxliniere:
Seems solid, right? Any reason not to merge?


bdring:
Adding "Probe" to the header. Closing because it is easier for me to do it via my own PR

Issue #736, opened 2022-12-05
Problem: SD Card not work.


zamby:
### Controller Board

WEMOS ESP32 LOLIN

### Machine Description

I am only  testwith  board ESP32.

### Configuration file

```yaml
name: "ESP32 Dev Controller V4"
board: "ESP32 Dev Controller V4"

stepping:
  engine: RMT
  idle_ms: 250
  dir_delay_us: 1
  pulse_us: 2
  disable_delay_us: 0

axes:
  shared_stepper_disable_pin: gpio.13:low
  
  x:
    steps_per_mm: 40
    max_rate_mm_per_min: 500
    acceleration_mm_per_sec2: 5
    max_travel_mm: 1000
    homing:
      cycle: 2
      mpos_mm: 10
      positive_direction: false
    
    motor0:
      limit_all_pin: gpio.17:low:pu
      stepstick:
        direction_pin: gpio.14
        step_pin: gpio.12
    motor1:
      null_motor:

  y:
    steps_per_mm: 800
    max_rate_mm_per_min: 2000
    acceleration_mm_per_sec2: 25
    max_travel_mm: 1000
    homing:
      cycle: 2
      mpos_mm: 10
      positive_direction: false

    motor0:
      limit_all_pin: gpio.4:low:pu
      stepstick:
        direction_pin: gpio.15
        step_pin: gpio.26
    motor1:
      null_motor:

  z:
    steps_per_mm: 800
    max_rate_mm_per_min: 2000
    acceleration_mm_per_sec2: 25
    max_travel_mm: 1000
    homing:
      cycle: 1
      mpos_mm: 10
      positive_direction: true

    motor0:
      limit_all_pin: gpio.16:low:pu
      stepstick:
        direction_pin: gpio.33
        step_pin: gpio.27
    motor1:
      null_motor:

spi:
  miso_pin: gpio.19
  mosi_pin: gpio.23
  sck_pin: gpio.18

sdcard:
  cs_pin: gpio.5
  card_detect_pin: NO_PIN

coolant:
  flood_pin: gpio.25
  mist_pin:  gpio.21

        
probe:
  pin: gpio.32:low:pu

PWM:
  pwm_hz: 5000
  output_pin: gpio.2
  enable_pin: gpio.22
  direction_pin: NO_PIN
  disable_with_s0: false
  s0_with_disable: true
  spinup_ms: 0
  spindown_ms: 0
  tool_num: 0
  speed_map: 0=0% 10000=100%
```

[...]

Issue #671, opened 2022-10-15
Problem: SD not work.


DEST1981:
### Controller Board

DIY board 4-axis. But setup to 3 axes
![esp32 board](https://user-images.githubusercontent.com/94748001/196007585-edafe9ae-2432-4d31-bf71-7e6ffb4e7994.JPG)

### Help From Board Vendor

- [ ] Yes
- [ ] No
- [X] Not Applicable

### Machine Description

External driver TB6600, PWM spindel. Limit switches pull up to +3.3v directly connect.

### Configuration file

```yaml
name: "ESP32 Dev Controller V4"
board: "ESP32 Dev Controller V4"

stepping:
  engine: RMT
  idle_ms: 0
  dir_delay_us: 1
  pulse_us: 10
  disable_delay_us: 0

axes:
  shared_stepper_disable_pin: gpio.13:high
  
  x:
    steps_per_mm: 100
    max_rate_mm_per_min: 2500
    acceleration_mm_per_sec2: 200
    max_travel_mm: 400
    homing:
      cycle: 2
      positive_direction: false
    
    motor0:
      limit_neg_pin: gpio.35:low
      standard_stepper:
        direction_pin: gpio.22
        step_pin: gpio.21
    motor1:
      null_motor:

  y:
    steps_per_mm: 100
    max_rate_mm_per_min: 2500
    acceleration_mm_per_sec2: 200
    max_travel_mm: 400
    homing:
      cycle: 2
      positive_direction: false

    motor0:
      limit_all_pin: gpio.34:low
      standard_stepper:
        direction_pin: gpio.17:low
        step_pin: gpio.16
    motor1:
      null_motor:

  z:
    steps_per_mm: 100
    max_rate_mm_per_min: 2500
    acceleration_mm_per_sec2: 200
    max_travel_mm: 220
    homing:
      cycle: 1
      allow_single_axis: true
      positive_direction: true
      mpos_mm: 1.000
      seek_mm_per_min: 600
      feed_mm_per_min: 50
      seek_scaler: 1.0
      feed_scaler: 1.0

    motor0:
      limit_neg_pin: NO_PIN
      limit_pos_pin: gpio.39:low
      limit_all_pin: NO_PIN
      hard_limits: false
      pulloff_mm: 1.000
      standard_stepper:
        direction_pin: gpio.12
        step_pin: gpio.14
    motor1:
      null_motor:

    
spi:
  miso_pin: gpio.19
  mosi_pin: gpio.23
[...]

PR #742, opened 2022-12-09
Configurable SD SPI speed via sdcard/frequency_hz:

MitchBradley:
The SPI speed for the SD card can now be configured via

```yaml
sdcard:
    frequency_hz: 10000000
```
That sets the SPI clock frequency to be used with the SD Card to 10 MHz, as opposed to the 20 Mhz that the ESP-IDF sdspi driver uses by default.

If frequency_hz is omitted or set to 0, the system default will be used.

Note that, in the SD Card probing process, the driver cycles through several frequencies, going slowly when probing the card, then switching to the highest frequency that it thinks the system and particular card can handle.  In the default case, the system often settles on 20 MHz, which typically works well for onboard SD card sockets with short wires, but is sometimes too fast for external adapters with level translators.  In that case, dialing back the frequency could make those adapters work.

Possible fix for #671 and #736

### DIFF
--- a/FluidNC/esp32/sdspi.cpp
+++ b/FluidNC/esp32/sdspi.cpp
@@ -67,7 +67,7 @@ static void call_host_deinit(const sdmmc_host_t* host_config) {
     }
 }
 
-bool sd_init_slot(int cs_pin, int cd_pin, int wp_pin) {
+bool sd_init_slot(uint32_t freq_hz, int cs_pin, int cd_pin, int wp_pin) {
     esp_err_t err;
 
     // Note: esp_vfs_fat_sdmmc/sdspi_mount is all-in-one convenience functions.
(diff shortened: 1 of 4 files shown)


jelsdon:
Positive feedback from my testing, working down in 1MHz increments from 10MHz, sdcard functioned at 9MHz.
I tested five iterations of rolling between 9MHz and undefined with it succeeding each time to both
correct and restore behavior. 

```
[MSG:INFO: FluidNC v3.6.5 (ConfigSDSPIspeed-8f917d91)]
[MSG:INFO: Compiled with ESP32 SDK:v4.4.1-1-gb8050b365e]

```


```
sdcard:
  card_detect_pin: NO_PIN
  cs_pin: gpio.5
  frequency_hz: 9000000
```

SD Card reading had been failing from  >3.4

MitchBradley:
@jelsdon Would you be willing to update the FluidNC wiki to add documentation for this new SD Card speed feature?  We could give you edit permissions, and you could search for SD Card references and determine where relevant information needs to be added.

jelsdon:
Yep, sounds good

alvarotorijano:
Hi guys thanks for adding this feature. Please could you also add it to the documentation?

bdring:
Done

FredGenius:
> The SPI speed for the SD card can now be configured via
> 
> ```yaml
> sdcard:
>     frequency_hz: 10000000
> ```
This worked for me (Lolin32 on breadboard, µSD card module connected with Dupont jump wires). 10MHz worked, 20MHz dint. Split the difference a few times and eventually arrived at 18MHz. When I move to prototype board maybe I can go higher, but I don't think this will be a problem.

Thanks @MitchBradley

PR #746, opened 2022-12-13
Added Openbuilds BlackBox X32 Example Config

synman:
This configuration was fully tested on a QueenBee Pro using a BlackBox X32 controller in an XYYZ configuration, Door & Limits NC, Probe NO


bdring:
I am OK with the example, but you need to remove the WebuUI/Hibernate stuff. We also target the devt branch (for now)

synman:
that is a 100% mistake on part.  I did not mean to do that commit on this branch.  I'll get it straightened out and familiarize myself with PR guidelines / etc.

synman:
I'll close this PR and do it properly as you describe

PR #755, opened 2022-12-22
added blackbox x32 example config

synman:
Assumes the following:

- Ganged Y motors (XYYZ)
- Door switch is connected and NC
- Limit switches are connected and NC


petervanderwalt:
> * Door switch is connected and NC
> * Limit switches are connected and NC

Hi @synman - can we please make the "default" line up more with the typical OpenBuilds BlackBox machine config for a better initial setup:
- Our Xtension Limits are Normally Open 
- Door switches are rarely installed - could leave it in the config but commented out?

synman:
Sure thing

PR #757, opened 2022-12-26
Uint32 t item

bdring:
This fixes the uint32_t config item. It was using int32_t with a min value of zero and a max value of INT32_MAX. A full 32 bit range is needed for the 32 bit Trinamic registers,

### DIFF
--- a/FluidNC/src/Configuration/RuntimeSetting.cpp
+++ b/FluidNC/src/Configuration/RuntimeSetting.cpp
@@ -72,6 +72,18 @@ namespace Configuration {
         }
     }
 
+    void RuntimeSetting::item(const char* name, uint32_t& value, uint32_t minValue, uint32_t maxValue) {
+        if (is(name)) {
+            isHandled_ = true;
+            if (newValue_ == nullptr) {
+                out_ << "$/" << setting_ << "=" << value << '\n';
+            } else {
+                value = atoi(newValue_);
+            }
+        }
+    }
+
+
     void RuntimeSetting::item(const char* name, float& value, float minValue, float maxValue) {
         if (is(name)) {
             isHandled_ = true;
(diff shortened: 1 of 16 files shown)


FluidNC/src/Configuration/RuntimeSetting.cpp
```diff
@@ -72,6 +72,22 @@ namespace Configuration {
         }
     }
 
+    void RuntimeSetting::item(const char* name, uint32_t& value, uint32_t minValue, uint32_t maxValue) {
+        if (is(name)) {
+            isHandled_ = true;
+            if (newValue_ == nullptr) {
+                out_ << "$/" << setting_ << "=" << value << '\n';
+            } else {
+                if (newValue_[0] == '-') {  // constrian negative values to 0
```
MitchBradley, line 81: Constrain not contstrian

FluidNC/src/Settings.cpp
```diff
@@ -120,6 +120,7 @@ void IntSetting::setDefault() {
 }
 
 Error IntSetting::setStringValue(char* s) {
+    log_info("IntSetting::setStringValue");
```
MitchBradley, line 123: Did you mean to leave in this message?

bdring:
No. I missed that one.

bdring:
I'll fix it with a direct change to devt.

FluidNC/src/Configuration/RuntimeSetting.cpp
```diff
@@ -72,6 +72,22 @@ namespace Configuration {
...
+                    value = 0;
+                    log_warn("Negative value not allowed");
+                } else
+                    value = atoll(newValue_);
+            }
+            constrain_with_message(value, minValue, maxValue);
```
MitchBradley, line 87: The canonical way to do error checking is to call strtoull with a non-null "endptr" argument.  Then check to see if the value returned in endptr is the same as startptr + strlen(startptr).  That will handle all manner of invalid characters.

PR #766, opened 2023-01-07
Move the machine position init to kinematics

bdring:
- Remove position initialization from main.cpp
- Most kins will init to 0,0...
- Some need to init to special values
- removed weakly define machine()...not used.


MitchBradley (commented):
If init _position is called only from init() do we need the method, as opposed to just putting the code directly in the kinematic init method?  In general I don't like unnecessary methods because it means I have to search to see if anybody uses them

bdring:
I thought the main program might want to call it some day.

MitchBradley:
Okay, merge it

PR #767, opened 2023-01-10
add option to configure Wifi regulatory domain 

### LINKED ISSUE #331: Wifi Country Code für using channel 11,12 + 13
I can connect to my wifi router only, if the channel on the Router ist 10 or less.

So i assume, the most commen range is as default setting, maybe US with channel 1-11 ? Even channel 11 here not work.

My request, a config setting (WebU| config.yaml ) für enable channel 11,12 and 13. This is useful if the router itself selects the channel depending on the load on the 2.4 Ghz band

kdomanski:
This introduces a new setting `$AP/Country`.
If set to a non-empty string, it will set the Wifi country on AP startup.
This impacts the available channels and maximum transmission power.
The ESP32 boots with `CN` set as the regulatory domain by default.

Fixes #331

### DIFF
--- a/FluidNC/src/WebUI/WifiConfig.cpp
+++ b/FluidNC/src/WebUI/WifiConfig.cpp
@@ -42,6 +42,103 @@ namespace WebUI {
         { "AP", WiFiAP },
         { "STA>AP", WiFiFallback },
     };
+
+    enum WiFiContry {
+        WiFiCountry01 = 0, // country "01" is the safest set of settings which complies with all regulatory domains
+        WiFiCountryAT,
+        WiFiCountryAU,
+        WiFiCountryBE,
+        WiFiCountryBG,
+        WiFiCountryBR,
+        WiFiCountryCA,
+        WiFiCountryCH,
+        WiFiCountryCN,
+        WiFiCountryCY,
+        WiFiCountryCZ,
+        WiFiCountryDE,
+        WiFiCountryDK,
+        WiFiCountryEE,
+        WiFiCountryES,
+        WiFiCountryFI,
+        WiFiCountryFR,
+        WiFiCountryGB,
+        WiFiCountryGR,
+        WiFiCountryHK,
+        WiFiCountryHR,
+        WiFiCountryHU,
+        WiFiCountryIE,
+        WiFiCountryIN,
+        WiFiCountryIS,
+        WiFiCountryIT,
+        WiFiCountryJP,
+        WiFiCountryKR,
+        WiFiCountryLI,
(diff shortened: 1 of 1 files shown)


MitchBradley (commented):
1. Please provide documentation in the form of additions/modifications to the wiki.
2. How does a naive user know which values can be used?  Ideally this would be an enumerated type.

kdomanski:
1. I reached out to you on Discord regarding activation of my wiki account.

2. [The list is well known](https://web.archive.org/web/20220425041126/https://content.cisco.com/chapter.sjs?uri=/searchable/chapter/www.cisco.com/content/en/us/td/docs/wireless/wcs/3-2/configuration/guide/wcscfg32/wcscod.html.xml), we can link to it in the docs. An invalid value will result in an error on boot and the startup will continue with the default setting (`CN`).
Screenshot 2023-01-12 at 10 54 58

kdomanski:
> I think I found you on the wiki. I activated you and gave you edit rights.

Yep, I got it. Thanks. Shall I update the wiki before this PR is merged?

> The ESP docs mention other values. Are these important?
> 
> https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/wifi.html
> 
> ```
> wifi_country_t config = {
>     .cc = "CN",
>     .schan = 1,
>     .nchan = 13,
> };
> ```

There are two ways you can set/read it: complete `wifi_country_t` struct (with `esp_wifi_set_country`) or just the code (with `esp_wifi_set_country`).

The [API reference for `esp_wifi_set_country`](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv420esp_wifi_set_countryPK14wifi_country_t) says:
```
It is discouraged to call this API since this doesn’t validate the per-country rules,
it’s up to the user to fill in all fields according to local regulations.
Please use `esp_wifi_set_country_code` instead.
```

kdomanski:
> It would still be better to make it an enumeration type. Many of our users do not read the docs, or do not manage to find the relevant section. Search for messageLevels in the source to see an example.

Understood, will do.

MitchBradley:
phy_init.c in the esp-idf source has a table of supported Country codes. It is static so you cannot use it directly but you could copy it out and make an enum setting.

kdomanski:
Done.

MitchBradley:
Everything looks good to me.  It seems to work.

kdomanski:
Wiki has been updated.

kdomanski:
Anything missing?

Issue #763, opened 2022-12-31
Problem: Flood Coolant toggle doesn't work


balthisar:
### Controller Board

MKS TinyBee

### Machine Description

n/a - controls related

### Input Circuits

```yaml
n/a
```

### Configuration file

```yaml
# HE0 - i2so.17 - mist enable (dust) M7
# HE1 - i2so.18 - flood enable (coolant) M8
coolant:
  mist_pin: i2so.17
  flood_pin: i2so.18
  delay_ms: 0
```

### Startup Messages

```text
<Idle|MPos:0.000,0.000,0.000|FS:0,0>
<Idle|MPos:0.000,0.000,0.000|FS:0,0>
$SS
[MSG:INFO: FluidNC v3.6.5]
[MSG:INFO: Compiled with ESP32 SDK:v4.4.1-1-gb8050b365e]
[MSG:INFO: Local filesystem type is spiffs]
[MSG:INFO: Configuration file:config.yaml]
[MSG:DBG: Running after-parse tasks]
[MSG:DBG: Checking configuration]
[MSG:INFO: Machine Queenbee Pro]
[MSG:INFO: Board MKS TinyBee V1.0]
[MSG:INFO: I2SO BCK:gpio.25 WS:gpio.26 DATA:gpio.27]
[MSG:INFO: SPI SCK:gpio.18 MOSI:gpio.23 MISO:gpio.19]
[MSG:INFO: SD Card cs_pin:gpio.5 detect:NO_PIN]
[MSG:INFO: Stepping:I2S_static Pulse:10us Dsbl Delay:2us Dir Delay:1us Idle Delay:25ms]
[MSG:INFO: User Digital Output:0 on Pin:I2SO.16]
[MSG:INFO: Axis count 3]
[MSG:INFO: Axis X (0.000,750.000)]
[MSG:INFO:   Motor0]
[MSG:INFO:     standard_stepper Step:I2SO.1 Dir:I2SO.2 Disable:I2SO.0]
[MSG:INFO:  X Neg Limit gpio.33:low:pu]
[MSG:DBG:  X Neg Limit 0]
[MSG:INFO: Axis Y (0.000,1260.000)]
[MSG:INFO:   Motor0]
[MSG:INFO:     standard_stepper Step:I2SO.4 Dir:I2SO.5 Disable:I2SO.3]
[MSG:INFO:  Y Neg Limit gpio.32:low:pu]
[MSG:DBG:  Y Neg Limit 0]
[...]

PR #771, opened 2023-01-11
WebUI update

MitchBradley:
Supports auto reporting, coolant toggles work, upload to SD subdirectories works, ...


MitchBradley:
Fixes #763

PR #774, opened 2023-01-13
M7.1 and M8.1 to turn off individual coolants

MitchBradley:
Several people seem to want individual turn-off of Flood and Mist, instead of individual turn-on and combined turn-off.  The current workaround is to use the override toggles via macros, but that is clumsy, in that they are toggles that depend on the current state, and the hack to inject override characters only works with the internal macros, not file-based macros.

This patch adds M7.1 to turn off M7 and M8.1 to turn off M8.

One objection to this formulation is that no other M code has a .N suffix, so this could be seen as a violation of basic GCode syntax rules, to the extent that is a thing.

One alternative would be to use M17 and M18.  They do not conflict with any LinuxCNC Mcodes, but they do conflict with Marlin (M17/M18 are enable/disable steppers) and with Haas (turret rotation forward/reverse).  Or some other MCode numbers.

Yet another alternative would be to add a P parameter, but M7 and M8 can be used in a block with other GCodes, so that is problematic.

Another alternative is to invent some $ commands like $coolant/flood=off

Or maybe just not solve this problem.

### DIFF
--- a/FluidNC/src/GCode.cpp
+++ b/FluidNC/src/GCode.cpp
@@ -480,7 +480,7 @@ Error gc_execute_line(char* line) {
                 break;
             case 'M':
                 // Determine 'M' command and its modal group
-                if (mantissa > 0) {
+                if (mantissa > 0 && !(int_value == 7 || int_value == 8)) {
                     FAIL(Error::GcodeCommandValueNotInteger);  // [No Mxx.x commands]
                 }
                 switch (int_value) {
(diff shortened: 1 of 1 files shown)


bdring:
This is a commonly requested feature. There is no perfect parameter. G code seem to be OK with the dot number. With M it seems like P is the most common ad hoc parameter. It seems like P0 feels more like an off than P1.

I have no strong objections to either way.

MitchBradley:
P in the context of M7 is sort of a non-starter, given the validity of "G0 X10 M7"

MitchBradley:
Marlin uses P with some Mcodes, but not ones that can appear on the same line with G.

bdring:
Good points. M7.1 seems like the way.

PR #777, opened 2023-01-16
Channel-based I2C_OLED

### LINKED ISSUE #775: Problem: Compiling with oled basic errors out on infile not declared in scope. line 161
### Controller Board

ESP32 DIY board. 

### Machine Description

DIY laser board based on ESP32.

### Input Circuits

_No response_

### Configuration file

```yaml
Not relevant. Compiling is not working.
```

### Startup Messages

```text
Not relevant. Compiling is not working.
```

### User Interface Software

_No response_

### What happened?

Tried compiling the software using visual studio code & platform IO. Uncommented the relevant lines to  include OLED support for the basic oled 0.96".  Got an error 
FluidNC/src/Custom/oled_basic.cpp:161:13: error: 'infile' was not declared in this scope

### Other Information

_No response_

MitchBradley:
This is a start.  It needs some work to display limit switches and whatnot.

It is configurable via:

```
i2c0:
  sda_pin: gpio.14
  scl_pin: gpio.13
  frequency: 400000

oled:
  i2c_num: 0
  i2c_address: 60
  width: 128
  height: 64
  radio_delay_ms: 2000
```
You can also have an i2c1 bus.

The i2c_address 60 is 0x3c in hex.  That is the default value so it can be omitted.
The display layout is hardcoded, so changing the width and height is probably not helpful.

This might resolve #775


bdring:
Works for me.

We should probably think about the config name to make sense if other are added. Maybe ***i2c_micro_oled:*** The size could be an enum for the size like 128x64. Maybe ***type:** would be better and allow options.

I can fill in some of the missing display features.



image

MitchBradley:
I dislike everything about the way this is configured.  My main focus was to figure out the Channel plumbing and write the report parser.  Here is how I think it should be configured:
```
i2cbus0:
   sda_pin: gpio.14
   scl_pin: gpio.13

oled_64x48:
   interface: i2cbus0@3c,400000
```
If the geometry is configurable, as opposed to being wired into the name, then we would have to solve the problem of adjusting the layout depending on the geometry.  That is pretty difficult for such small displays.

bdring:
I added the limit switch indicator boxes primarily to get a feel for the new methods. I think only parsing the POG status response will be very limiting. 

For example: I might want a three state checkbox for switches. Open, checked and invisible if the switch is not defined. Why show a probe box if there is no probe. The code already looks at the config for n_axis. Is it ok to look at the config for other things?

MitchBradley:
Ideally you would look at the config only for stuff that does not change from report to report.  If you delve into the guts for stuff that changes, there is problem of when to look.  It would be better if no separate polling task were required.  We could invent some new reports or report fields that only appear via auto reporting.

bdring:
I agree about no polling. I will try to do everything with the ***init*** and the status ping. Is it better to store a local flag during ***init*** whether the probe exists or read it from the config each time, like is being done with ***n_axis***

MitchBradley:
> Is it better to store a local flag during init whether the probe exists or read it from the config each time, like is being done with n_axis

It makes little difference.  Accessing information via config->a->b takes about 6 machine instructions - a small fraction of a microseconds - insignificant compared to the I2C transactions necessary to write to the display.  The display code, as with all Channel operations, is now running on the other core from the GCode/stepping code, so it is competing for cycles with other things that are not realtime-sensitive.

MitchBradley:
I separated out the i2c bus config.  There is an example config in example-configs/6_Pack_OLED.cpp

MitchBradley:
> can I use the 12 scl and 15 sda pin for oled？

You can if your board wiring has not already used those pins for something else.

MitchBradley:
I am still working on making it compile correctly.  Give me time.

MitchBradley:
New version:

1. The name is now oled, not i2c_oled.  If we want to support SPI later, we can add parameters "spi_number" and "spi_cs_pin" to distinguish.
2. Small displays like 64x48 (as in the old tiny_oled) are now supported.  The code automatically adapts to the smaller display by changing fonts and omitting certain information like the SD filename (but it does display the SD percent).  On a small display, the limit checkboxes are not used, but the letter "L" appears after the axis name if that axis is limited.
3. The code no longer reaches into the system state variables, but instead gets information like radio stuff by parsing startup messages.

PatrikRindlisbacher:
I hope I can find someone to help me with the UART test.   
Today I will test my **Light Tower** on the additional UART channel.  
https://github.com/PatrikRindlisbacher/grbl-Status-light-tower-M5-Atom-Matrix

bdring:
Should the old OLED_tiny and OLED_basic files be removed?

PatrikRindlisbacher:
Excuse me if I make my comment

From my point of view, an external solution for a display would be the better solution.

So all wishes can be fulfilled.

For example, Duet3D does it like this.
image

bdring:
This one is just a simple dumb display. A display with an external processor could parse the status itself.

bdring:
Anyone who is spending more than a few dollars on a display should consider a tablet. You can't beat the features of a tablet with cloud access, high resolution, etc.

PatrikRindlisbacher:
Duet has its own processor with firmware.     
As a basic version, an M5 Core would be a good version. Doesn't cost very much.  
Even a separate solution with ESP Pico would not be very expensive.  
I am currently working on solutions that will go in this direction.  
  
I would be very happy if **Fluid NC stays performant**. :-)

MitchBradley:
> Should the old OLED_tiny and OLED_basic files be removed?

Yes

MitchBradley:
> From my point of view, an external solution for a display would be the better solution.

Yes, especially for anything that has interactive capabilities.  I find that the Tiny OLED display is mainly useful for basic status, especially information about whether it is connected to WiFi and what is the IP address.

The current OLED display code uses the "parse information from channel output" model, so it can be seen as a template for an external display.  As such, it run on the communications core, not on the realtime core, so it is unlikely to cause performance problems.

sotchrys:
great work...!

PR #785, opened 2023-01-23
Support for external uart configuration

marsjo:
This is an attempt to use multiple UARTs for TMC2009. 
I am using this branch for a while now to run my MPCNC with five TMC2009 drivers.

Note: You need to specify the UART before using within a TMC2009 config:

```yaml
uart_xy:
  rxd_pin:  gpio.2
  txd_pin:  gpio.15
  baud: 115200
  mode: 8N1

uart_z:
  rxd_pin: gpio.16
  txd_pin: gpio.4
  baud: 115200
  mode: 8N1

[...]

axes:  
  x:
   motor0:
      limit_neg_pin: gpio.36:low
      pulloff_mm: 2.000
      tmc_2209:
        uart: uart_xy
        addr: 0
        r_sense_ohms: 0.110
        [...]
  y:
    motor0:
      limit_neg_pin: gpio.32:low
      pulloff_mm: 1.000
      tmc_2209:
        uart: uart_z
        addr: 0
        [...]
```

You can still use a direct specification of the UART - but with the known limitations of using a single UART.

### DIFF
--- a/FluidNC/src/Configuration/Parser.cpp
+++ b/FluidNC/src/Configuration/Parser.cpp
@@ -7,9 +7,11 @@
 #include "../EnumItem.h"
 
 #include "../Config.h"
+#include "../Uart.h"
 
 #include <climits>
 #include <math.h>  // round
+#include <regex.h>
 
 namespace Configuration {
     Parser::Parser(const char* start, const char* end) : Tokenizer(start, end) {}
@@ -38,8 +40,48 @@ namespace Configuration {
         return result;
     }
 
+    char const* Parser::match(const char* pattern) {
+        if (token_.state != TokenState::Matching || token_.keyStart_ == nullptr) {
+            return nullptr;
+        }
+
+        regex_t reg;
+        if (regcomp(&reg, pattern, REG_EXTENDED | REG_NOSUB) != 0) {
+            log_error("Can not compile regex: " << pattern);
+            return nullptr;
+        }
+
+        size_t length = 0;
+        for (const char* pos = token_.keyStart_; pos != token_.keyEnd_; pos++) {
+            length++;
+        }
+
+        char name[40];
+        strncpy(name, token_.keyStart_, length);
+        name[length] = '\0';
+
+        int result = regexec(&reg, name, 0, 0, 0);
+        regfree(&reg);
+
+        if (result == 0) {
+            char* result = new char[length+1];
+            strncpy(result, name, length + 1);
+            return result;
+        } else {
+            return nullptr;
+        }
+    }
+
     StringRange Parser::stringValue() const {
         return StringRange(token_.sValueStart_, token_.sValueEnd_);
+    }
+
+    void Parser::uartValue(Uart*& uart) const {
+        size_t length = token_.sValueEnd_ - token_.sValueStart_;
+        char name[40];
+        strncpy(name, token_.sValueStart_, length);
+        name[length] = '\0';
+        Uart::externals.get(name, uart);
     }
 
     bool Parser::boolValue() const {
--- a/FluidNC/src/Configuration/Parser.cpp
+++ b/FluidNC/src/Configuration/Parser.cpp
@@ -11,7 +11,7 @@
 
 #include <climits>
 #include <math.h>  // round
-#include <regex.h>
+#include <regex>
 
 namespace Configuration {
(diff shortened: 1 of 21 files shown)


FluidNC/src/Configuration/Parser.cpp
```diff
@@ -38,10 +40,50 @@ namespace Configuration {
...
+        regex_t reg;
+        if (regcomp(&reg, pattern, REG_EXTENDED | REG_NOSUB) != 0) {
+            log_error("Can not compile regex: " << pattern);
+            return nullptr;
+        }
+
+        size_t length = 0;
+        for (const char* pos = token_.keyStart_; pos != token_.keyEnd_; pos++) {
+            length++;
+        }
```
MitchBradley: Why not "size_t length = token_.keyEnd_ - token_.keyStart;" Am I missing something subtle?

FluidNC/src/Machine/MachineConfig.cpp
```diff
@@ -35,6 +35,8 @@ namespace Machine {
         handler.item("name", _name);
         handler.item("meta", _meta);
 
+        handler.extSection("uart_", &Uart::externals);
+
```
MitchBradley, line 39: It would appear that "uart_" is a regex pattern that is intended to match "uart1" etc.  I don't see how that works, since _ is not a regex special character.  I would have expected the pattern to be "uart.".

FluidNC/src/Configuration/Parser.cpp
```diff
@@ -7,9 +7,11 @@
 #include "../EnumItem.h"
 
 #include "../Config.h"
+#include "../Uart.h"
 
 #include <climits>
 #include <math.h>  // round
+#include <regex.h>
```
MitchBradley: If we use the standard C++ regex library this becomes "#include \<regex\>"

FluidNC/src/Configuration/Parser.cpp
```diff
@@ -38,10 +40,50 @@ namespace Configuration {
...
+        if (token_.state != TokenState::Matching || token_.keyStart_ == nullptr) {
+            return nullptr;
+        }
+
+        regex_t reg;
+        if (regcomp(&reg, pattern, REG_EXTENDED | REG_NOSUB) != 0) {
+            log_error("Can not compile regex: " << pattern);
+            return nullptr;
+        }
+
```
MitchBradley: std::regex rp(pattern, std::regex_constants::basic);

FluidNC/src/Configuration/Parser.cpp
```diff
@@ -38,10 +40,50 @@ namespace Configuration {
...
+            length++;
+        }
+
+        char name[40];
+        strncpy(name, token_.keyStart_, length);
+        name[length] = '\0';
+
+        int result = regexec(&reg, name, 0, 0, 0);
+        regfree(&reg);
+
```
MitchBradley: if (std::regex_search(name, rp) {

FluidNC/src/Configuration/Parser.cpp
marsjo: I believe it was just me not knowing what I am doing ;-/

FluidNC/src/Machine/MachineConfig.cpp
marsjo: Sorry, you are right. I switched to `uart_first` instead of `uart_1`. The idea was to be able to use any name you like as long as it starts with `uart_`, e.g. `uart_xy` and `uart_z`.

MitchBradley: So the use of regular expressions is just a way of implementing "startsWith" - although the regex formulation will also match "foouart".

MitchBradley: I am trying to figure out a way to use the GenericFactory mechanism for this, like for Spindles and Motors and Kinematic systems.

marsjo: I knew that this is your longterm plan. I am just not sure how to do it nicely together with what is already there. Of course you could also define a `extSection` with `^motor\d` and storing it to `&Motor::external`.
I just do not like those static external lists in the objects. But my knowledge about c/c++ is very limited.

PR #787, opened 2023-01-25
Uart sections

MitchBradley:
You can now define UART sections at the top level of the config file and refer to them where they are used:

```yaml
uart1:
  txd_pin: gpio.0
  rxd_pin: gpio.4
  rts_pin: gpio.13
  baud: 115200
  mode: 8N1

uart2:
  txd_pin: gpio.22
  rxd_pin: gpio.21
  baud: 115200
  mode: 8N1
```

uart0 is the USB serial console one; you cannot define or change it because it is so important for startup messaging.

Within a Spindle class, you can still use the old syntax where there is a subordinate "uart:" subsection, or, alternatively, you can refer to an externally-defined uart section with:
```yaml
NowForever:
  uart_num: 1
  modbus_id: 1
```

Trinamic UART devices like TMC2208 and TMC2209 no longer support the subordinate "uart:" subsection (making that work was just too hard); you must use the external UART section with "uart_num: N":

```yaml
axes:
  x:
    motor0:
      tmc_2209:
        uart_num: 2
        addr: 2
        ... 
  z:
    motor0:
      tmc_2209:
        uart_num: 2
        addr: 1
        ...
```
With this scheme it should be possible to use, say, 5 TMC UART devices by using two UARTs, getting around the limitation of 4 device addresses.

Dynamixel2 is like Trinamic; you must use "uart_num: N":
```yaml
axes:
  x:
    motor0:
      dynamixel2:
        uart_num: 1
        id: 3
```
The Dynamixel driver, as currently written, can use only one UART for all the Dynamixels.

Now for the new feature - You can attach a serial channel to a UART:
```yaml
uart_channel1:
   uart_num: 2
```
That new channel will accept input just like the USB serial channel, the telnet channels, the websocket channels, etc.
You can also do uart_channel2, but no more.  uart_channel0 is the implied USB serial channel.

There are only three physical UART engines on ESP32, so if you try to do too much you will run out of UARTs.  If you are using a UART for a VFD spindle and another for TMC drivers, you can't make another uart_channel .  UARTs can only be shared among like devices, such as multiple TMC drivers or multiple modbus VFD spindles.

This needs a lot of testing.  I tried a number of config snippets and verified that they do what I think, but I have not tested with actual VFD, TMC, Dynamixel, and secondary serial pendant hardware.


PatrikRindlisbacher:
I will need this function to connect an additional serial control.
 Is there anything I can test or help with?

MitchBradley:
You could build from the branch and see if it works for you.

PatrikRindlisbacher:
I will start and test secondary serial pendant hardware :-) Thanks

PatrikRindlisbacher:
Which pins are suitable on the 6er Bart CNC controller?

MitchBradley:
It is possible that Uart2 might run at a higher baud rate if you use GPIO16 for RxD and GPIO17 for TxD, because of the way the ESP32 IOMUX and GPIO Matrix works. You would have to study the ESP32 Technical Reference manual to determine how much faster, if any.  Uart1 cannot be sped up because the high speed pins for UART1 are used for the SPI FLASH.

If you do not use the "special" pins, most functions can be mapped to any pin via the GPIO Matrix, so any pins that are available would be okay for UART functions.

Since you have done some testing of this, you might be the expert, any you should advise us.

PatrikRindlisbacher:
Hi Mitch I'm Patrick from Switzerland :-)

I am happy to make my knowledge and time available to this project.
I have a big laboratory at home with lots and .... lots of material.

The question regarding the pin for the external UART was more from the point of view of the Bart Dring 6 CNC controller. So that a pluggable module could be built.

Today I will delve a little deeper into the topic and report back.

"Einen schönen Gruss aus der Schweiz" --> greeting Patrik

bdring:
I made some UART test modules for the 6 pack last year. I think they worked fine on sockets 2, 3, and 4 of the 6 pack. 

image000000

PatrikRindlisbacher:
Hello bdring

You can send such a TTL module with the module from the tindie order Number: 371480
:-)

PatrikRindlisbacher:
:-) Enjoy your holidays first :-)

PatrikRindlisbacher:
image
That's exactly what I'm up to with the universal HUB. Should be easy and very flexible.

MitchBradley:
In order to work as an input expander for low-latency things like limit switches, it will need to use a different protocol compared to just sending GCode and $ commands via the GRBL protocol.  The line protocol is too complicated for low-latency stuff.  It might be possible to extend the set of realtime characters, but that would still not be nearly as fast as a dedicated packet protocol.

PatrikRindlisbacher:
I was just annoyed that GRBL didn't do the status information very efficiently.
My HUB makes sense for user inputs and tasks.
The real-time communication could be packed with it.
So you wouldn't have to double anything that GRBL can already do.

MitchBradley:
> The real-time communication could be packed with it in addition.
That would not be easy because of the way that channels are polled.  We are dealing with a lot of history here.  I don't have time to explain all the problems.

PatrikRindlisbacher:
For now I will concentrate on testing the UART Sectiona. :-) So that this can soon be migrated to the master.

PatrikRindlisbacher:
hello mitch
I'm making good progress with my UART test and project.

I need info on the $RI command

https://github.com/users/PatrikRindlisbacher/projects/1/views/1?pane=issue&itemId=20880354

MitchBradley:
Somebody needs to fix the wiki.
$Report/Interval=N works as follows:
If the system is not in a motion state, a channel that has issued $RI=(nonzero) will receive automatic reports (on that channel only) - as with ?, $G, and # - when any field in one of those reports changes.
If the system is in a motion state, that channel will receive ? reports <...> at intervals of N milliseconds.

PatrikRindlisbacher:
Thank you very much...that helps me a lot

PatrikRindlisbacher:
I hope I can find someone to help me with the UART test. 
Today I will test my Light Tower on the additional UART channel.
https://github.com/PatrikRindlisbacher/grbl-Status-light-tower-M5-Atom-Matrix

FluidNC/src/Motors/TMC2208Driver.cpp
```diff
@@ -12,19 +12,17 @@
...
+        TrinamicUartDriver::init();
+        if (!_uart) {
+            return;
         }
-
         if (_r_sense == 0) {
             _r_sense = TMC2208_RSENSE_DEFAULT;
         }
 
         tmc2208 = new TMC2209Stepper(_uart, _r_sense, _addr);
```
marsjo, line 23: `new TMC2209Stepper(...` -> `new TMC2208Stepper(...`

Old Code but looks like a c&p error to me.
If I get it right it does not matter at all - but should be changed probably.

marsjo (commented):
While trying to understand the code I found a minor C&P issue in older, but related code.

bdring:
Seems to work fine.

```yaml
uart2:
  txd_pin: gpio.26
  rxd_pin: gpio.4
  baud: 115200
  mode: 8N1
  
uart_channel1:
   uart_num: 2
 ```

image

This is a second FluidTerm session

```
Select a COM port
0: COM4 (\Device\Silabser1)
1: COM35 (\Device\Silabser0)
Choice: 0
FluidNC v1.2.0 using COM4
Exit: Ctrl-C, Ctrl-Q or Ctrl-], Clear screen: CTRL-W
Upload: Ctrl-U, Reset ESP32: Ctrl-R, Send Override: Ctrl-O
<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>
<Idle|MPos:0.000,0.000,0.000|FS:0,0|Ov:100,100,100>
<Idle|MPos:0.000,0.000,0.000|FS:0,0>
<Idle|MPos:0.000,0.000,0.000|FS:0,0>
<Idle|MPos:0.000,0.000,0.000|FS:0,0>
```

PatrikRindlisbacher:
Hi Mitch
I will download and test the current merged version in the next few days. Thanks very much

Rich235:
Hi, Thanks for all the above work on UARTs. I'm trying to use UART1 for to add a serial pendant to FluidNC. BUT I also want to connect FluidNC to my computer via USB - UART0. I've a setup on my desktop using TTL<>USB dongle into UART1 and the normal ESP32 USB to the PC and UGS or CNCjs as sender. When I connect to the RX of the dongle and use a sender via USB I can see the Mpos info being sent out over UART1 and the sender works normally. However, when I also connect the TX pin to UART1 it takes over so the sender over USB is no longer listened to, it then just repeats info over its consol. (Same with two FluidTerm consols, UART1 becomes the only input channel.)

I then noticed this comment in Patrik's excellent status light tower example:
"Selection of the UART  ... you can use RX & TX directly on the controller. (if you don't use the USB port simultaneously)"

[https://github.com/PatrikRindlisbacher/grbl-Status-light-tower-M5-Atom-Matrix](url)

Can they not operate together? I'd hoped I could use both together and allow jogging over a serial pendant when the pendant checked the machine was in idle, it can send jog commands and the sender over USB would continue to function and update its DROs. When I stop jogging I can then start a job via the sender. Is this possible using the two UARTs? 

I'm doing this to avoid wireless comms and have lowest latency and, in theory, it looked the simplest approach. Any pointers much appreciated. Thanks.

MitchBradley:
I cannot tell what you might be doing wrong without full details of your configuration.  File a github issue and provide all the data.

MitchBradley:
There is no way to do it from a config file, because we depend so heavily on UART0 interaction for debugging and other basic procedures.  It seems that you too need to use UART0.  That being the case, why do you want to reuse the same pins for two different UARTn hardware modules?  If you want to use one pair of pins, why not use the same UART0 module?

It would be possible to modify the code to switch the pin 1/3 pair, but it seems unlikely that we would interested in making that a standard feature.

Rich235:
Yes, I'm going to have to change my hardware so I can use UART2 on pins 16 and 17. I realise now the USB port is via a separate chip wired into the UART0 on pins 1 and 3. I've just come from Teensy which was native and I wasn't thinking straight. Thanks for the pointers.

PR #798, opened 2023-01-30
Added Beeper control and full pin map info

villamany:
A MKS DLC32 v2.1 configuration file used for laser. -Added the control of Bepper pin of this board missing in others configuration files. -Added full pin map info at the end of the file as comments


bdring:
Have you seen the wiki page?

http://wiki.fluidnc.com/en/hardware/3rd-party/MKS_DLC32

It recommends against I2SO_STREAM for lasers.

I suggest adding ***@villamany on Github*** to the meta section if people have questions.

villamany:
Thanks is correct. Replaced by I2S_STATIC as recomended for this board. 
Tested in STREAM mode and ghost engravings appears

MitchBradley (changes requested):
This looks good, but it appears that there are two separate copies of the file, one at the top level and another in example_configs.  I think the top level file should be deleted.

villamany:
Done, thanks

PR #802, opened 2023-02-04
add a setting to block HTTP server while in motion

kdomanski:
- Blocking webserver while in motion can now be configured through a setting, instead of necessitating a recompile
- It blocks fetching any file from flash, not just index.html.gz
- javascript-based redirect pages have been replaced with HTTP redirects

### DIFF
--- a/FluidNC/src/WebUI/WebServer.h
+++ b/FluidNC/src/WebUI/WebServer.h
@@ -18,6 +18,7 @@ class WebServer;
 
 namespace WebUI {
     static const int DEFAULT_HTTP_STATE = 1;
+    static const int DEFAULT_HTTP_BLOCKED_DURING_MOTION = 1;
     static const int DEFAULT_HTTP_PORT  = 80;
 
     static const int MIN_HTTP_PORT = 1;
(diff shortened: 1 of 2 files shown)


kdomanski:
Hey @bdring , this has been hanging for a while without any feedback. Is there something fundamentally hard to swallow here (e.g. the redirect changes) or is just lack of time? Cheers.

MitchBradley:
Bart just got back from vacation and I have been overwhelmed with repair jobs around the house.  The code in the PR looks good but we just want to test it.

FluidNC/src/WebUI/WebServer.h
```diff
@@ -18,6 +18,7 @@ class WebServer;
 
 namespace WebUI {
     static const int DEFAULT_HTTP_STATE = 1;
+    static const int DEFAULT_HTTP_BLOCKED_DURING_MOTION = 0;
```
MitchBradley: Please make the default value true (1).

MitchBradley:
It works well.  The default value for the setting needs to be true, and the wiki needs to be searched for places that should be updated to reflect the new setting.

kdomanski:
Gotcha, I didn't mean rush you.
I just changed the default to true.

kdomanski:
- Added a note at http://wiki.fluidnc.com/e/en/features/local_file_system
- Added the setting to the list at http://wiki.fluidnc.com/en/features/commands_and_settings
- Added the setting to the list at http://wiki.fluidnc.com/en/features/wifi_bt

PR #817, opened 2023-03-01
Devt

christianbanach:
Hello

I really appreciate the effort that You put to get it better every day.

Wanted to know if it is possible to add TS35 screen support as there are many boards with this screen everywhere and it Will be very usefull.

Thank you very much


MitchBradley:
Someone else would have to do the work and support it.  I personally am not interested because it would be a never-ending source of people asking for it to behave ever so slightly differently.

MitchBradley:
I think this should be a Feature Request issue, not a Pull Request.  Pull Requests are for when you are offering new code for us to consider.

PR #827, opened 2023-03-09
Add  skew correction.

elf128:
Added ability to add and tweak skew correction, for cases when physical


elf128:
Skew class holds references to various SkewAxis, which are holders of initial settings, but based on that Skew compute matrices for direct and reverse coordinate transformation.

When amount of Skew along certain axis is edited using web interface, validate is called. This is the point when matrices are recompute. Since internal state of a configuration class become modified during validation, this method cannot be const. The alternative is to store matrices somewhere else, and most logical place is inside of Kinematics. However in that case we need the way to propagate data or change event from Configuration Class outside and in that case Configuration class has to hold reference to external data, which is less then ideal.

So in my opinion, it's better to lift const restriction on validation method. Also, any other statefull Configuration in the future would require to be able to modify itself on validate anyway.

MitchBradley:
Please rebase this PR to the Devt branch per http://wiki.fluidnc.com/en/development/pull_request_guidelines#branches

elf128:
Done. New #829  have been submitted.

Will close this one.

PR #829, opened 2023-03-09
Devt

elf128:


elf128:
Next iteration after #827 .

elf128:
Never mind. Found bugs in this PR. Will fix and re-submit.

Sorry for confusion.

MitchBradley:
1. Please do the removal of "const" from validate as a separate PR.  That will be easy to accept as a separate item.
2. Please use clang-format, with the .clang-format definitions in the top directory, to ensure that your submissions conform to our style.
3. For the substantive part of the patch, it would be better to make a separate kinematic system, to isolate cartesian from bugs or changes.

elf128:
> 1. Please do the removal of "const" from validate as a separate PR.  That will be easy to accept as a separate item.
Sure. Will do. 

>     2. Please use clang-format, with the .clang-format definitions in the top directory, to ensure that your submissions conform to our style.
I tried not to mess it up, but probably I did. Will check it.

>     3. For the substantive part of the patch, it would be better to make a separate kinematic system, to isolate cartesian from bugs or changes.
I have no problems with moving it into separate kinematic system. Please advise on naming.

MitchBradley:
The name "skewed" comes to mind, but perhaps someone has a better name?  The generalization is a linear transform.

For a long time I have considered breaking out motors as separate from axes, with a transformation matrix between them.  That would handle skew, rotations, and even CoreXY kinematics.

elf128:
> The name "skewed" comes to mind, but perhaps someone has a better name? The generalization is a linear transform.
> 
> For a long time I have considered breaking out motors as separate from axes, with a transformation matrix between them. That would handle skew, rotations, and even CoreXY kinematics.

What do you think about making GenericCartesian, which just hold matrices ( identity by default ) use it for transformation, get initialised with identity by default and also allows to initialise  transform, while taking care of matrix inversion in that case.

And then I'll implement Skewed inherited from it. Rotation fill be easy to add later.

MitchBradley:
How does it affect homing?  Homing has been the fly in the ointment of every alternative kinematic system to date, especially when multiple motors are involved.  I have spent countless hours dealing with homing glitches.

elf128:
This is a good question. 

It doesn't. Or at least it should not in my opinion. I did go far into how homing is implemented, so I might be wrong, but I assume, homing just work in motor space. Ant then you set motor to zero and that's is.
Homing is the way to find the limit of the motor, how much can you go in direction of the motor. Cause mechanically endstop is placed on the physical axis. I can imagine exotic cases when it's not true but then it should be implemented separately, and I imagine  It also set motor's zero and machine's zero, since zero vector times any matrix gives back zero vector.  

I know, you can quickly get into weird case by, let's say 90° rotation. Then when you press button "home X" which axis should be homed motorX or "whatever axis looking into x direction after transformation". I can expect heated debate about this matter.

MitchBradley:
So, in other words, you don't really know.

elf128:
> So, in other words, you are guessing. Homing is not exactly in motor space, because the limit switches are typically at the end of axes, How do you put a limit switch on a motor? If there is skew, when you home to, say, the X limit, the distance you move depends on the Y coordinate. Then when you reach the X limit, when you go toward the Y limit in motor space, you might pull off the switch or grind into the limit.
> 
> In CoreXY, which is just a linear transform of cartesian. homing has to be done in axis space. If you say $H=X, the motors have to move simultaneous to drive in the X direction towards the limit switch on the axis. There is no obvious way to put a limit switch on the motor itself. It gets even more complicated for multi-axis homing like $H=XY. In that case you have to make a diagonal move (one motor) until you hit one or the other limit switch, and then you have to switch over to an axis-aligned move on the non-limited axis. The rate computation is also complicated by the directions, as are the pulloffs.
> 
> People will expect it to work.

Ok, so the source of confusion is my choice of words. When I said, the limit switch is on the motor, I mean, not laterally on the motor itself, but on the physical axis driven by the motor. Motor spin the screw or pull the belt in order to move the carriage, which sits on a rails of some sort. This rail define motor's direction and limits. The switch is at the end of the rail.
 In Kinematic System the space where motor's positions defined called "motor", that's why I'm calling it motor space. So, it's meter of naming.
 
So in that light, like I said, there's no point to try to implement homing in any transformed space, whether it's model or machine. You home motors first ( by that, I mean, you move one motor in a defined direction until switch is hit ), during that all transformation has to be ignored. I know, there're cases when this rule doesn't apply and we need hacks.

elf128:
> Motors rotate so motor space is not linear (in the sense of "straight line", not in the sense of a "linear system" or "linear transfrom"). For any cartesian robot, there is always an implicit transform from that rotation space to the axis space. In the simple case, the transform is multiplication by a diagonal matrix where the diagonals are the steps/mm values, plus a constant vector. Homing is the process of establishing the constant vector, which is optimized out by setting steps[axis] = 0 when the limit is found - but then the constant vector is put back in via the mpos-to-wpos transform. The presence of the constant term makes the total transform non-linear, as linear transforms always map 0 to 0.

There's a trick we do in graphics al the time. You add additional dimension and agree that it has to be equal to 1. Then translation just goes into additional column and whole thing become linear.
Not sure we have to do the same here. Just a thought. 

> When you consider multiple motors on a gantry - which is very common - it get even worse. Each motor has, effectively, a different constant term. Squared homing is the process of finding those constant terms. Doing that without racking the system is quite tricky. In a sense, if you have multiple motors on an axis with a limit switch on each end of the gantry, the limit switch is in "motor space", but it is a different sort of motor space. It is not the result of an NxN linear transform from N axes to N motors, but rather a secondary transform from one step space component to two ganged motors. Basically, homing is a nightmare.

When you have more then one motor controlling the axis, this is not GenericCartesian any more. Like I said, we'll have more issues in that case. Simply from math perspective, your "motor space" is not linearly independent, so matrix is degenerate and not-inversable, so right here you already need a custom solution. 
And true, homing in that case is a mine field.

elf128:
BTW, most of the board I've seen so far in both CNC and 3DPrinting, they just feed dual-motors from the same stepper, so you cannot do anything about it in the FW. I know, there're boards where you can control those motors separately, but I only seen references to those in sources and never personally.

MitchBradley:
There are a lot of FluidNC boards that have enough motor drivers to do individual control, instead of just hanging multiple motors off of one driver.  Bart's 6-pack, MKS tinybee, etc.  Individual control is the norm in the FluidNC space.

MitchBradley:
And stress-free (anti racking) squared homing, with per-Y-motor limit switches. is also very common in dual-Y FluidNC configurations.

PR #848, opened 2023-03-28
Add new Kinematic system with skew correction

elf128:
New system adds ability to tweak skew correction, for cases when physical
Also, there's implementation of generic abstraction for transformation to-from axis space, it may be used for future implementation of rotated or any other exotic coordinate systems.

**Known issues:** Machine will do homming in transformed coordinate system as well, which may or may not by desirable. Idealy, config should have an option where user explicitly define homming behavior.


elf128:
Oh, wait. My bad. PR into wrong branch.

PR #849, opened 2023-03-28
Add new Kinematic system with skew correction

elf128:
New system adds ability to tweak skew correction, for cases when physical
Also, there's implementation of generic abstraction for transformation to-from axis space, it may be used for future implementation of rotated or any other exotic coordinate systems.

**Known issues:** Machine will do homming in transformed coordinate system as well, which may or may not by desirable. Ideally, config should have an option where user explicitly define homming behavior.

### DIFF
--- a/FluidNC/src/Kinematics/GenericCartesian.cpp
+++ b/FluidNC/src/Kinematics/GenericCartesian.cpp
@@ -1,0 +1,187 @@
+// Copyright (c) 2020 -	Bart Dring
+// Copyright (c) 2023 -	Vlad A.
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+#include "GenericCartesian.h"
+
+#include "src/Machine/MachineConfig.h"
+#include "src/Machine/Axes.h"  // ambiguousLimit()
+//#include "Skew.h"              // Skew, SkewAxis
+#include "src/Limits.h"
+
+namespace Kinematics {
+    void GenericCartesian::init() { 
+        log_info("Kinematic system: " << name());
+        init_position();
+    }
+
+    // Initialize the machine position
+    void GenericCartesian::init_position() {
+        auto n_axis = config->_axes->_numberAxis;
+        for (size_t axis = 0; axis < n_axis; axis++) {
+            set_motor_steps(axis, 0);  // Set to zeros
+        }
+    }
+
+    bool GenericCartesian::cartesian_to_motors(float* target, plan_line_data_t* pl_data, float* position) {
+        if ( _mtx ) {
+            _mtx->transform( _buffer, target );
+            return mc_move_motors( _buffer, pl_data);
+        } else
+            // Without skew correction motor space is the same cartesian space, so we do no transform.
+            return mc_move_motors(target, pl_data);
+    }
+
+    void GenericCartesian::motors_to_cartesian(float* cartesian, float* motors, int n_axis) {
+        if ( _rev ) {        
+            _rev->transform( cartesian, motors );
+        }
+        else
+            // Without skew correction motor space is the same cartesian space, so we do no transform.
+            copyAxes(cartesian, motors);
+    }
+
+    void GenericCartesian::transform_cartesian_to_motors(float* motors, float* cartesian) {
+        // Without skew correction motor space is the same cartesian space, so we do no transform.
+        if ( _mtx ) {
+            _mtx->transform( motors, cartesian );
+        }
+        else
+            copyAxes(motors, cartesian);
+    }
+
+    bool GenericCartesian::canHome(AxisMask axisMask) {
+        if (ambiguousLimit()) {
+            log_error("Ambiguous limit switch touching. Manually clear all switches");
+            return false;
+        }
+        return true;
+    }
+
+    bool GenericCartesian::limitReached(AxisMask& axisMask, MotorMask& motorMask, MotorMask limited) {
+        // For Cartesian, the limit switches are associated with individual motors, since
+        // an axis can have dual motors each with its own limit switch.  We clear the motors in
+        // the mask whose limits have been reached.
+        clear_bits(motorMask, limited);
+
+        auto oldAxisMask = axisMask;
+
+        // Set axisMask according to the motors that are still running.
+        axisMask = Machine::Axes::motors_to_axes(motorMask);
+
+        // Return true when an axis drops out of the mask, causing replan
+        // on any remaining axes.
+        return axisMask != oldAxisMask;
+    }
+
+    void GenericCartesian::releaseMotors(AxisMask axisMask, MotorMask motors) {
+        auto axes   = config->_axes;
+        auto n_axis = axes->_numberAxis;
+        for (int axis = 0; axis < n_axis; axis++) {
+            if (bitnum_is_true(axisMask, axis)) {
+                auto paxis = axes->_axis[axis];
+                if (bitnum_is_true(motors, Machine::Axes::motor_bit(axis, 0))) {
+                    paxis->_motors[0]->unlimit();
+                }
+                if (bitnum_is_true(motors, Machine::Axes::motor_bit(axis, 1))) {
+                    paxis->_motors[1]->unlimit();
+                }
+            }
+        }
+    }
+
+    GenericCartesian::~GenericCartesian() {
+        if (_mtx)
+            delete _mtx;
+
+        if (_rev)
+            delete _rev;
+    }
+
+    ////////////////////
+
+    template< typename number >
+    void GenericCartesian::Mtx< number >::dumpRow( const uint idx ) const {
+        // Prints a row of the matrix into log.
+        // Useful for debuging.
+        char line[256];
+        char* C = line;
+
+        for( uint i = 0; i < _pitch; ++i ) {
+            number V = value( idx, i );
+            if ( V >= 0.0 )
+                sprintf( C, " %4.4f ", V );
+            else
+                sprintf( C, "%4.4f ", V );
+
+            C = C + strlen( C );
+        }
+
+        log_info( line );
+    }
+
+    template< typename number >
+    void GenericCartesian::Mtx< number >::dump() const {
+        for( uint i = 0; i < _lines; ++i ) {
+            dumpRow( i );
+        }
+    }
+
+    template< typename number >
+    void GenericCartesian::Mtx< number >::transform(  number* to, const number* from ) const {
+        for( uint j = 0; j < _pitch; ++j ) {
+            number A = 0.0;
+            for( uint i = 0; i < _lines; ++i )
+                A += from[ i ] * value( i, j );
+
+            to[ j ] = A;
+        }
+    }
+
+    bool GenericCartesian::GJ_invertMatrix( const uint size, const Mtx<float>* A, Mtx<float>* const B ) {
+        //log_info( "GJ_invertMatrix" );
+        // Gauss Jordan Matrix inversion. 
+        Mtx<double> T( size, size * 2 );
+        uint i,j,k;
+
+        T.allocate();
+        
+        for( i = 0; i < size; ++i ) {
+            for( j = 0; j < size; ++j ) {
+                *T.ptr( i, j )        = A->value( i, j );
+                *T.ptr( i, j + size ) = ( i == j ) ? 1.0 : 0.0;
+            }
+        }
+
+        //T.dump();
+        for( i = 0; i < size; ++i ) {
+            if ( T.value( i,i ) == 0 ) {
+                T.deallocate();
+                return false;
+            }
+
+            for( j = 0; j < size; ++j ) {
+                if ( i != j ) {
+                    double S = T.value( j, i ) / T.value( i, i );
+                    for( k = 0; k < size*2; ++k )
+                        *T.ptr( j, k ) = T.value( j, k ) - S * T.value( i, k );
+                    
+                }
+            }
+        }
+
+        //log_info( "After elimination" );
+        //T.dump();    
+        for( i = 0; i < size; ++i ) {
+            for( j = 0; j < size; ++j ) 
+                *B->ptr( i, j ) = static_cast< float >( T.value( i, j + size ) / T.value( i,i ) );
+        }
+
+        T.deallocate();
+        return true;
+    }
+
+    template void GenericCartesian::Mtx< float >::transform( float* to, const float* from  ) const;
+    template void GenericCartesian::Mtx< float >::dump() const;
+
+}
--- a/FluidNC/src/Kinematics/GenericCartesian.h
+++ b/FluidNC/src/Kinematics/GenericCartesian.h
@@ -1,0 +1,91 @@
+// Copyright (c) 2020 -	Bart Dring
+// Copyright (c) 2023 -	Vlad A.
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+#pragma once
+
+/*
+	GenericCartesian.h
+
+	This is a kinematic system for where the motors operate in the cartesian space.
+
+    An abstract class which should be a base for diferent transformation cases.
+
+    Unlike Cartisian kinamatic system, this system allow motor space to vary from machine space. 
+	Transformation from system space to motor space is defined as matrix.
+    All axis must by linearly independent. Matrix has to be inversable.
+
+*/
+
+#include "Kinematics.h"
+
+#define LOG_MATRIX_CONTENT
+
+namespace Kinematics {
+
+    class GenericCartesian : public KinematicSystem {
+    public:
+        GenericCartesian() = default;
+
+        GenericCartesian(const GenericCartesian&)            = delete;
+        GenericCartesian(GenericCartesian&&)                 = delete;
+        GenericCartesian& operator=(const GenericCartesian&) = delete;
+        GenericCartesian& operator=(GenericCartesian&&)      = delete;
+
+        // Kinematic Interface
+        virtual void init() override;
+        virtual void init_position() override;
+
+        virtual bool cartesian_to_motors(float* target, plan_line_data_t* pl_data, float* position) override;
+        void         motors_to_cartesian(float* cartesian, float* motors, int n_axis) override;
+        void         transform_cartesian_to_motors(float* cartesian, float* motors) override;
+
+        bool canHome(AxisMask axisMask) override;
+        void releaseMotors(AxisMask axisMask, MotorMask motors) override;
+        bool limitReached(AxisMask& axisMask, MotorMask& motors, MotorMask limited) override;
+
+        // Configuration handlers:
+//        void afterParse() override {}
+//        void group(Configuration::HandlerBase& handler) override;
+//        void validate() override {}
+
+        // Name of the configurable. Must match the name registered in the cpp file.
+//        const char* name() const override { return "Generic Cartesian"; }
+
+    protected:
+        template <typename number>
+        class Mtx {
+            uint    _pitch;
+            uint    _lines;
+            number* _buffer;
+
+        public:
+            Mtx(const uint row, const uint col) : _pitch(col), _lines(row) { _buffer = new number[_pitch * _lines]; };
+
+            void allocate() {  }
+            void deallocate() {
+            }
+
+            number* getBuffer() { return _buffer; }
+            number  value(const uint row, const uint col) const { return _buffer[row * _pitch + col]; }
+            number* ptr(const uint row, const uint col) { return _buffer + row * _pitch + col; }
+            void    transform(number* to, const number* from) const;
+
+            void dumpRow(const uint idx) const;
+            void dump() const;
+
+            ~Mtx() {
+                if ( _buffer )
+                    delete[] _buffer;
+            }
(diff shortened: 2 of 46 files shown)


FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+        for (size_t axis = 0; axis < n_axis; axis++) {
+            set_motor_steps(axis, 0);  // Set to zeros
+        }
+    }
+
+    bool GenericCartesian::cartesian_to_motors(float* target, plan_line_data_t* pl_data, float* position) {
+        if ( _mtx ) {
+            _mtx->transform( _buffer, target );
+            return mc_move_motors( _buffer, pl_data);
+        } else
```
MitchBradley: Style - There is no need for else since the previous clause ends with return;

FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+            // Without skew correction motor space is the same cartesian space, so we do no transform.
+            return mc_move_motors(target, pl_data);
+    }
+
+    void GenericCartesian::motors_to_cartesian(float* cartesian, float* motors, int n_axis) {
+        if ( _rev ) {        
+            _rev->transform( cartesian, motors );
+        }
+        else
```
MitchBradley: Style - `} else {` on one line

All clauses must have braces - no unbraced single line clauses.

FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+            // Without skew correction motor space is the same cartesian space, so we do no transform.
+            copyAxes(cartesian, motors);
+    }
+
+    void GenericCartesian::transform_cartesian_to_motors(float* motors, float* cartesian) {
+        // Without skew correction motor space is the same cartesian space, so we do no transform.
+        if ( _mtx ) {
+            _mtx->transform( motors, cartesian );
+        }
+        else
```
MitchBradley: Style - } else {

FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+                }
+                if (bitnum_is_true(motors, Machine::Axes::motor_bit(axis, 1))) {
+                    paxis->_motors[1]->unlimit();
+                }
+            }
+        }
+    }
+
+    GenericCartesian::~GenericCartesian() {
+        if (_mtx)
```
MitchBradley: Style - brackets

FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+    template< typename number >
+    void GenericCartesian::Mtx< number >::dumpRow( const uint idx ) const {
+        // Prints a row of the matrix into log.
+        // Useful for debuging.
+        char line[256];
+        char* C = line;
+
+        for( uint i = 0; i < _pitch; ++i ) {
+            number V = value( idx, i );
+            if ( V >= 0.0 )
```
MitchBradley: Style - braces

FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    ////////////////////
+
+    template< typename number >
+    void GenericCartesian::Mtx< number >::dumpRow( const uint idx ) const {
+        // Prints a row of the matrix into log.
+        // Useful for debuging.
+        char line[256];
+        char* C = line;
```
MitchBradley: idx and i should be row and col.  i is just shorthand for idx which is just shorthand for index, and neither convey the intention.

How about doing this with iostreams like other places in FluidNC
```c++
    std::ostringstream msg;
    for( uint col = 0; i < _pitch; ++col ) {
        number v = value( row, col );
        msg << std::fixed << std::setprecision(4) << std::setw(9) << v;
    }
    log_info(msg.str());
```

FluidNC/src/Kinematics/GenericCartesian.h
```diff
@@ -0,0 +1,91 @@
...
+            number* getBuffer() { return _buffer; }
+            number  value(const uint row, const uint col) const { return _buffer[row * _pitch + col]; }
+            number* ptr(const uint row, const uint col) { return _buffer + row * _pitch + col; }
+            void    transform(number* to, const number* from) const;
+
+            void dumpRow(const uint idx) const;
+            void dump() const;
+
+            ~Mtx() {
+                if ( _buffer )
```
MitchBradley: Style - braces

FluidNC/src/Kinematics/GenericCartesian.h
```diff
@@ -0,0 +1,91 @@
...
+#pragma once
+
+/*
+	GenericCartesian.h
+
+	This is a kinematic system for where the motors operate in the cartesian space.
+
+    An abstract class which should be a base for diferent transformation cases.
+
+    Unlike Cartisian kinamatic system, this system allow motor space to vary from machine space.
```
MitchBradley: Misspellings etc ...

Should read:
Unlike the Cartesian kinematic system which uses the identity transform between the
axis and motor spaces, this system uses an arbitrary linear transform defined by an
invertible matrix.

FluidNC/src/Kinematics/GenericCartesian.h
```diff
@@ -0,0 +1,91 @@
...
+//        void group(Configuration::HandlerBase& handler) override;
+//        void validate() override {}
+
+        // Name of the configurable. Must match the name registered in the cpp file.
+//        const char* name() const override { return "Generic Cartesian"; }
+
+    protected:
+        template <typename number>
+        class Mtx {
+            uint    _pitch;
```
MitchBradley: uint    _pitch;  // The number of columns in a matrix row
uint    _lines;   // The number of rows

But it would better still to name them _columns and _rows

FluidNC/src/Kinematics/Skewed.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "src/Machine/MachineConfig.h"
+#include "src/Machine/Axes.h"  // ambiguousLimit()
+#include "src/Limits.h"
+
+namespace Kinematics {
+
+    void SkewAxis::group(Configuration::HandlerBase& handler) {
+        handler.item("dist", _dist, 1.0f, 100000.0f );
+        handler.item("x", _x[0], -1000.0f, 1000.0f );
+        handler.item("y", _x[1], -1000.0f, 1000.0f );
```
MitchBradley: What are the units of x, y, z, ... ?

FluidNC/src/Kinematics/Skewed.h
```diff
@@ -0,0 +1,62 @@
...
+#include "GenericCartesian.h"
+
+namespace Kinematics {
+    class SkewAxis  : public Configuration::Configurable {
+        uint _axisIdx;
+    public:
+        SkewAxis( int currentAxis ) : _axisIdx( currentAxis ) { _x[5] = _x[4] = _x[3] = _x[2] = _x[1] = _x[0] = 0.0f; };
+
+        float    _dist     = 10.0f;
+        float    _x[ 6 ];
```
MitchBradley: Name it _offsets, not _x .  x means the X axis, whereas this is an array of offsets.

FluidNC/src/Kinematics/Skewed.h
```diff
@@ -0,0 +1,62 @@
...
+        void group(Configuration::HandlerBase& handler) override;
+        void init();
+
+        void getRow( const uint count, float* buf );
+    };
+
+
+    class Skewed : public GenericCartesian {
+        static constexpr const char* _names = "xyzabc";
+        uint _numberAxis;
```
MitchBradley: Rename this to, for example, _numberSkewAxis, because _numberAxis is too easy to confuse with the global _numberAxis config variable, even though it is inside the namespace.

FluidNC/src/Kinematics/Skewed.h
```diff
@@ -0,0 +1,62 @@
...
+        void afterParse() override {};
+        void group(Configuration::HandlerBase& handler) override;
+        void init();
+
+        void getRow( const uint count, float* buf );
+    };
+
+
+    class Skewed : public GenericCartesian {
+        static constexpr const char* _names = "xyzabc";
```
MitchBradley: Eliminate this and just use config->_axes->axisName(axis)   We want the axis names to be defined in only one place.

FluidNC/src/Kinematics/Skewed.cpp
```diff
@@ -0,0 +1,132 @@
...
+        for (size_t i = 0; i < n_axis; ++i) {
+            tmp[0] = _names[i];
+            tmp[1] = 0;
+
+            handler.section(tmp, _axis[i], i);
+        }
+    }
+
+    void Skewed::validate() {
+        log_info("validation for Skewed");
```
MitchBradley: Either eliminate these log_ messages entirely (preferably), or make them log_verbose().

FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+            delete _mtx;
+
+        if (_rev)
+            delete _rev;
+    }
+
+    ////////////////////
+
+    template< typename number >
+    void GenericCartesian::Mtx< number >::dumpRow( const uint idx ) const {
```
MitchBradley: use size_t instead of uint .  The type uint does not appear anywhere else in the code base so using it here makes this module seem out of place.

FluidNC/src/Kinematics/Skewed.h
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2020 -	Bart Dring
+// Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
+
+#pragma once
+
+/*
+	Cartesian.h
```
MitchBradley: The comment block needs to be rewritten instead of just copied from Cartesian

FluidNC/src/Kinematics/Skewed.cpp
```diff
@@ -0,0 +1,132 @@
+#include "Skewed.h"
+
+#include "src/Machine/MachineConfig.h"
+#include "src/Machine/Axes.h"  // ambiguousLimit()
+#include "src/Limits.h"
+
+namespace Kinematics {
+
+    void SkewAxis::group(Configuration::HandlerBase& handler) {
+        handler.item("dist", _dist, 1.0f, 100000.0f );
```
MitchBradley: Use "distance_mm" because configuration items name their units.

FluidNC/src/Kinematics/Skewed.h
```diff
@@ -0,0 +1,62 @@
...
+        // Configuration handlers:
+        void afterParse() override;
+        void group(Configuration::HandlerBase& handler) override;
+        void validate() override;
+
+        // Name of the configurable. Must match the name registered in the cpp file.
+        const char* name() const override { return "Skew corrected Cartesian"; }
+
+    protected:
+        SkewAxis* _axis[6];
```
MitchBradley: Name it _skewAxis because "axis" usually means an axis number.

FluidNC/src/Kinematics/GenericCartesian.cpp
```diff
@@ -0,0 +1,187 @@
...
+            dumpRow( i );
+        }
+    }
+
+    template< typename number >
+    void GenericCartesian::Mtx< number >::transform(  number* to, const number* from ) const {
+        for( uint j = 0; j < _pitch; ++j ) {
+            number A = 0.0;
+            for( uint i = 0; i < _lines; ++i )
+                A += from[ i ] * value( i, j );
```
MitchBradley: braces

FluidNC/src/Kinematics/GenericCartesian.h
```diff
@@ -0,0 +1,91 @@
...
+
+    Unlike Cartisian kinamatic system, this system allow motor space to vary from machine space. 
+	Transformation from system space to motor space is defined as matrix.
+    All axis must by linearly independent. Matrix has to be inversable.
+
+*/
+
+#include "Kinematics.h"
+
+#define LOG_MATRIX_CONTENT
```
MitchBradley: Eliminate the #define and use log_debug().

FluidNC/src/Kinematics/Skewed.cpp
```diff
@@ -0,0 +1,132 @@
...
+            delete _rev;
+
+        _mtx = new Mtx<float>(_numberAxis, _numberAxis);
+        _rev = new Mtx<float>(_numberAxis, _numberAxis);
+
+        for (size_t axis = 0; axis < _numberAxis; axis++) {
+            auto a = _axis[axis];
+            if (a) {
+                a->init();
+                a->getRow( _numberAxis, _mtx->ptr( axis, 0 ) );
```
MitchBradley: getRow is hard to understand because the SkewAxis method reaches out and injects something into the outside matrix.  Something like this would be clearer and simpler:
```c++
    if (a) {
        a->init();
        auto row = _mtx->ptr(axis, 0);
        for (size_t i = 0; i < _numberAxis; ++i) {
            row[i] = (i == axis) ? 1.0 : 0.0;  // Start with 1 on the diagonals
            row[i] += a->_offsets[i] / a->_axis_length;  // Add adjustment factors
        }
```
With that formulation, _axisIdx and getRow are unnecessary

FluidNC/src/Kinematics/Skewed.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include "GenericCartesian.h"
+
+namespace Kinematics {
+    class SkewAxis  : public Configuration::Configurable {
+        uint _axisIdx;
+    public:
+        SkewAxis( int currentAxis ) : _axisIdx( currentAxis ) { _x[5] = _x[4] = _x[3] = _x[2] = _x[1] = _x[0] = 0.0f; };
+
+        float    _dist     = 10.0f;
```
MitchBradley, line 21: _axis_length is easier to understand

MitchBradley (commented):
The math looks correct but there are a number of stylistic issues that will make the code easier to understand.

Also we need some documentation with examples for how you would describe a machine with specific amounts of skew.

Does it do the right thing if items are left unspecified, i.e. can you write
   skewed:
      x:
        axis_length_mm: 300
        y_offset_mm: 2

elf128:
Wow, that's a detailed review. Thank you. I'll go over all your points tomorrow. 

Meanwhile, I would like to point out couple of things in general.
Yes, axis you don't want to correct can be omitted in config. Everything you didn't specified will be assumed to be zero. In case of empty section, Skewed will act as Cartesian. 
As a test, I entered this as a test
```
kinematics:
  Skewed:
    x:
      dist: 100.0
      z: 0.475
    y:
      dist: 100.0
      z: -1.1
```
      
And that's what i've got:

```
[MSG:INFO: Kinematic system: Skew corrected Cartesian]
[MSG:INFO:       Skew ( 0.000, 0.000, 0.475 ) over 100.000mm]
[MSG:INFO:       Skew ( 0.000, 0.000, -1.100 ) over 100.000mm]
[MSG:INFO:       Skew ( 0.000, 0.000, 0.000 ) over 10.000mm]
[MSG:INFO: Direct transform]
[MSG:INFO:  1.0000  0.0000  0.0047 ]
[MSG:INFO:  0.0000  1.0000 -0.0110 ]
[MSG:INFO:  0.0000  0.0000  1.0000 ]
[MSG:INFO: Reverse transform]
[MSG:INFO:  1.0000  0.0000 -0.0047 ]
[MSG:INFO:  0.0000  1.0000  0.0110 ]
[MSG:INFO:  0.0000  0.0000  1.0000 ]
[MSG:INFO: validation is done] 
```

Your example will work as you expect, aside from syntax. Yes, I've seen your suggestion about naming and units. You, probably right, but I would like you to take couple of things into consideration. 

1. About units. Actually, this way of definition is unit agnostic. As much, I'm not ever missing the opportunity to piss-off imperial folks and enforce metric here, this definition is distance over distance. Therefor as long as you use same units for all measurements, you should be good.
2. About axis_length, well, this might be misleading. For example. How would I measure the skew. Let's say, I have a perfectly rectangular piece for aluminium 150mm long. So, I place it on a table and move the tool such as it's touching one corner. Then I move the tool 150mm in one direction and measure, the rise of a tool over known 150mm of travel. That's what goes into the settings. ( Obviously, I'm assuming, that by that time, user already knows which direction cause tool rise, so he/she will do it in the right direction and not plunge the tool into the metal. )  
With naming axis_length this might look like, you have to define actual length of the axis not the distance that been chosen for measurements.

MitchBradley:
Re axis_length - point taken.

Re units - you could make the same argument for all FluidNC units that are expressed in mm.  The units really should be "GCode units in G21 mode".  Conventionally that is taken to be mm, but GCode fundamentally operates in arbitrary units and you just have to be consistent.  But that is less understandable to most folks than just calling it mm.

Another approach that would eliminate both problem would be to have "y_factor" which is the dimensionless offset/distance .  That would mean that the user has to do a division.  Regardless of how it is parameterized, it will be necessary to document the process quite clearly.

MitchBradley:
I found this attractive matrix class implementation on stackexchange:
```c++
class my_matrix {
  std::vector<std::vector<bool> >m;
public:
  my_matrix(unsigned int x, unsigned int y) {
    m.resize(x, std::vector<bool>(y,false));
  }
  class matrix_row {
    std::vector<bool>& row;
  public:
    matrix_row(std::vector<bool>& r) : row(r) {
    }
    bool& operator[](unsigned int y) {
      return row.at(y);
    }
  };
  matrix_row& operator[](unsigned int x) {
    return matrix_row(m.at(x));
  }
};
// Example usage
my_matrix mm(100,100);
mm[10][10] = true;
```
Obviously, replace <bool> with <float>

FluidNC/src/Kinematics/GenericCartesian.h
elf128: Well, this is the whole point, to decouple matrix geometry, from memory layout, and make sure no one confuse one term and another. Pitch is equal to column number in this particular case, but this may be not always true. 
This is very common source of mistakes, when people use the same naming which starts to imply the way data stored in memory. Pitch always define an offset between lines and line can store whatever. In general matrices can be stored in column-major or row-major order. And it's really hard to predict which one will be more optimal and then it's hard to change it when all variables related to storage has naming related to matrix geometry. 

And yes, I know what you'll say, that this is not relevant to ESP32, where memory layout is almost irrelevant. And I'll answer, that, yes you're right. I just cannot get rid of habit to name things the way it should be. 

Sure, I can rename it, if you insist.

FluidNC/src/Kinematics/Skewed.cpp
elf128: Well, the real question is, are those are the same units. Some dimensions can be angular.

FluidNC/src/Kinematics/Skewed.h
elf128: Ok, I'll rename it. :)
But it's very typical to use name x for point in space in multi dimensional case, when x is a vector. It's usually **X**, but that's goes against your naming convention.

FluidNC/src/Kinematics/Skewed.cpp
elf128: Ahh, that's my debugging stuff. Forgot to remove it.

elf128:
OMG, no. This code is a good example how to trash your cache for good. 

Imagine how far and inconsistent the pitch will be in this case. There's even no guaranties they will be stored in order. Every time you do matrix operation using data stored like this, you can jump all over the memory.

elf128:
as for _mm. Ok, I'm sold.

FluidNC/src/Kinematics/Skewed.cpp
elf128: The function returns basis-vector. And return it into provided buffer is fairly typical. Especially on the systems where you don't have defined float2, float3, ... types. 
I would say that when you have one function which build whole matrix is harder to read, but emmm. Ok. I can throw everything into one place.

elf128:
Hey, 

Did I forgot anything?

FluidNC/src/Kinematics/GenericCartesian.h
```diff
@@ -0,0 +1,81 @@
...
+        template <typename number>
+        class Mtx {
+            size_t  _pitch; // Here, it's equal the number of columns in a matrix row
+            size_t  _lines; // The number of rows
+            number* _buffer;
+
+        public:
+            Mtx(const size_t row, const size_t col) : _pitch(col), _lines(row) { _buffer = new number[_pitch * _lines]; };
+
+            void allocate() {}
```
The-Futur1st, line 55: Empty functions, may be get rid of them, or move real allocation and deallocation to those funcs.
Another approach:
- get rid of those functions and dtor,
- make buffer in the RAII style
std::unique_ptr<number[]> _buffer;
...
ctor() {
  _buffer.reset(new number[_pitch * _lines])
}

PR #855, opened 2023-04-08
Add timer_ms config option

MarkJB:
When using fast servos, the default timer update frequency of 75ms results in rough motion.

Adding the timer_ms config option allows a user to define how often the PWM signal is updated.


bdring:
This issue has already been fixed and merged to the devt branch.

BTW: We do not accept PRs to the main branch.

PR #867, opened 2023-04-19
Use timers instead of tasks for simple periodic operations

MitchBradley:


MitchBradley:
> Hey Mitch, out of curiosity, what's the benefit here?

1. It reduces RAM usage because each separate task requires a stack and other data structures, whereas the system task that handles tasks is already "paid for".  Each additional timer requires only a small amount of RAM.
2. It lets us have different timer values for different servos, because each servo can have its own separate timer.  The previous code had a task that handled all servos, with on time interval that applied to all the servos.  Allowing separate intervals would have required either separate tasks - even more RAM usage - or essentially duplicating the freertos timer logic.
3. The code is simpler because you don't need all the task boilerplate.  The timer dispatch boilerplate is just a couple of lines.

PR #889, opened 2023-05-06
S3 wip glen

MitchBradley:


sergioegutierrez:
Great work! I was about to ask how to make this change as I'm working in a S3 plotter.
What do you think about adding some note on pins 19 and 20 to remain this are for USB communication?

PR #893, opened 2023-05-11
Search File in List of Uploaded files

mullataufiq:
If number of file are there in memory card of controller then we are not able to search the file from list of uploaded files. Web UI must have search option to search files.


MitchBradley:
> If number of file are there in memory card of controller then we are not able to search the file from list of uploaded files. Web UI must have search option to search files.

Please create an issue for this.  Pull requests are for a different purpose.

PR #937, opened 2023-06-23
Use `std::string_view` in favor of `StringRange`

dymk:
Uses C++17 `std::string_view` over a pair of `char *begin, *end` pointers, or `StringRange`. Desugars to almost the same thing - a `char*` and a `size_t` length under the hood, but affords niceties like methods found on `std::string`.

A smallish PR and refactor to get a little more comfortable with the YAML parsing stack, and to consolidate on standard library types rather than hand-rolled custom implementations of the same.

### Testing

Ran on esp32 devboard, ran `$CD` and observed YAML was dumped, indicating it was parsed correctly 

### Performance
Before:
```
RAM:   [==        ]  18.3% (used 59824 bytes from 327680 bytes)
Flash: [=====     ]  51.3% (used 1007813 bytes from 1966080 bytes)
```

After:
```
RAM:   [==        ]  18.2% (used 59780 bytes from 327680 bytes)
Flash: [=====     ]  51.2% (used 1005773 bytes from 1966080 bytes)
```

Saves a few RAM and flash bytes, albeit a very small amount. Nice to see it didn't go up though, despite the removal of `inline` qualifiers (modern compilers tend to ignore those anyways).


bdring:
We don't accept PRs against the main branch. See our pull request guidelines.

dymk:
Sorry about that - [this one](https://github.com/bdring/FluidNC/pull/939) should be against the correct branch

PR #939, opened 2023-06-23
Use `std::string_view` in favor of `StringRange`

dymk:
Uses C++17 `std::string_view` over a pair of `char *begin, *end` pointers, or `StringRange`. Desugars to almost the same thing - a `char*` and a `size_t` length under the hood, but affords niceties like methods found on `std::string`.

A smallish PR and refactor to get a little more comfortable with the YAML parsing stack, and to consolidate on standard library types rather than hand-rolled custom implementations of the same.

### Testing

Ran on esp32 devboard, ran `$CD` and observed YAML was dumped, indicating it was parsed correctly 

### Performance
Before:
```
RAM:   [==        ]  18.3% (used 59824 bytes from 327680 bytes)
Flash: [=====     ]  51.3% (used 1007813 bytes from 1966080 bytes)
```

After:
```
RAM:   [==        ]  18.2% (used 59780 bytes from 327680 bytes)
Flash: [=====     ]  51.2% (used 1005773 bytes from 1966080 bytes)
```

Saves a few RAM and flash bytes, albeit a very small amount. Nice to see it didn't go up though, despite the removal of `inline` qualifiers (modern compilers tend to ignore those anyways).

PR #945, opened 2023-06-25
Added static_cast to suppress compiler warning.

MitchBradley:
PR #939 compiles but there is a compiler warning about int being converted to unsigned int.  This PR makes that warning go away.

### DIFF
--- a/FluidNC/src/Configuration/Tokenizer.h
+++ b/FluidNC/src/Configuration/Tokenizer.h
@@ -11,64 +11,36 @@
 namespace Configuration {
 
     class Tokenizer {
-        std::string_view current_;
+        std::string_view _remainder;
 
-        void skipToEol();
-
-        void Inc() {
-            if (current_.size() > 0) {
-                current_ = current_.substr(1);
-            }
-        }
-        char Current() const { return Eof() ? '\0' : current_[0]; }
-
-        bool IsAlpha() {
-            char c = Current();
-            return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
-        }
-
-        bool IsSpace() { return Current() == ' '; }
-
-        bool IsWhiteSpace() {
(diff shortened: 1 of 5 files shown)


dymk (approved):
Yep, looks reasonable to me. The length should never overflow a size_t, nor could the value be negative.

FluidNC/src/Configuration/Tokenizer.h
```diff
@@ -11,64 +11,36 @@
...
-        bool IsIdentifierChar() { return IsAlpha() || IsDigit() || Current() == '_'; }
-
-        bool IsEndLine() { return Eof() || Current() == '\n'; }
-
-        bool IsDigit() {
-            char c = Current();
-            return (c >= '0' && c <= '9');
-        }
+        bool isWhiteSpace(char c) { return c == ' ' || c == '\t' || c == '\f' || c == '\r'; }
+        bool isIdentifierChar(char c) { return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_'; }
```
dymk: nit - move implementation into Tokenizer.cpp

dymk (approved):
Looks good to me, although I'm not super familiar with the ins and outs of FluidNC's yaml subset. Looks like the logical changes are confined to Tokenizer - nice to see some simplifications!

FluidNC/src/Configuration/Tokenizer.h
MitchBradley: Done.

MitchBradley:
> I'm not super familiar with the ins and outs of FluidNC's yaml subset

In a nutshell - the only value data type at the tokenizing level is string.  Values are converted from string to another type on a per-key basis.  There are no arrays.  So it's a very restricted subset.

PR #946, opened 2023-06-27
Proxy legacy Grbl settings for sender compatibility

dymk:
Some senders (such as gSender) rely on standard numbered Grbl settings for improved reporting of machine state, and some utility functionality like "Jog to this quadrant of the machine bed" - in particular, reporting of homing status and axes limits.

This adds the relevant numbered settings, derived from `MachineConfig`. They are read-only, and writing to them will result in a `error:162, ReadOnlySetting` error.

### Testing

New output for `$$`
```
Grbl 3.0 [FluidNC v3.0.x (grbl-proxy-settings-446e05d0) (noradio) '$' for help]

> $$
$22=1
$21=0
$20=0
$100=3000.000
$101=3000.000
$102=1000.000
$110=50.000
$111=50.000
$112=200.000
$120=100.000
$121=100.000
$122=100.000
$130=800.000
$131=600.000
$132=100.000
$10=1
ok

> $120 = 1
error:162
```

gSender now renders the quadrant jogging tool (it was not rendered before)
image

Grbl firmware "state" correctly interpreted from sender:
image

### DIFF
--- a/FluidNC/src/Error.h
+++ b/FluidNC/src/Error.h
@@ -78,6 +78,7 @@ enum class Error : uint8_t {
     ConfigurationInvalid        = 152,
     UploadFailed                = 160,
     DownloadFailed              = 161,
+    ReadOnlySetting             = 162,
 };
 
 const char* errorString(Error errorNumber);
--- a/FluidNC/src/NutsBolts.h
+++ b/FluidNC/src/NutsBolts.h
@@ -85,13 +85,6 @@ const char* to_hex(uint32_t n);
 bool  char_is_numeric(char value);
 char* trim(char* value);
 
-template <class T>
-void swap(T& a, T& b) {
-    T c(a);
-    a = b;
-    b = c;
-}
-
 template <typename T>
 T myMap(T x, T in_min, T in_max, T out_min, T out_max) {  // DrawBot_Badge
     return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
--- a/FluidNC/src/Settings.cpp
+++ b/FluidNC/src/Settings.cpp
@@ -5,6 +5,7 @@
 #include "WebUI/Commands.h"     // WebUI::COMMANDS
 #include "System.h"             // sys
 #include "Protocol.h"           // protocol_buffer_synchronize
+#include "Machine/MachineConfig.h"
 
 #include <map>
 #include <limits>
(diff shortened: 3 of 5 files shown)


dymk:
cc @cotepat, who I believe has a fork of FluidNC that does something similar

MitchBradley (commented):
Is there a way to do this without the preprocessor fiddling?

Like, maybe a map from numbers to string names of the relevant tree entry?  Entries like { 122, "/axes/z/acceleration_mm_per_sec2" }

dymk:
I could do something like this: 

```cpp
template <typename T>
using FloatProxyDef = std::tuple<std::string_view,                       // description
                                 std::string_view,                       // grbl numerical name
                                 std::string_view,                       // fluidnc full name,
                                 std::function<T(MachineConfig const&)>  // getter
                                 >;

std::array float_proxy_defs = {
    FloatProxyDef<float> { "Proxy for $130 (X max travel, mm)",
                           "130",
                           "Grbl/MaxTravel/X",
                           [](MachineConfig const& config) { return config._axes->_axis[0]->_maxTravel; } },
    FloatProxyDef<float> { "Proxy for $131 (Y max travel, mm)",
                           "131",
                           "Grbl/MaxTravel/Y",
                           [](MachineConfig const& config) { return config._axes->_axis[1]->_maxTravel; } },
    FloatProxyDef<float> { "Proxy for $132 (Z max travel, mm)",
                           "132",
                           "Grbl/MaxTravel/Z",
                           [](MachineConfig const& config) { return config._axes->_axis[2]->_maxTravel; } },
};
```

versus 

```cpp
FLOAT_PROXY_SETTINGS_XYZ(13, "max travel, mm", "MaxTravel", _maxTravel)
```

and then iterate over `float_proxy_defs` to create the proxy settings. I want to avoid dynamic memory allocation, which I think would be required without a lot of really nasty template metaprogramming.

I agree that the macro version is unfortunate because, well, it's really opaque, but the repetitive nature of the "spelled out" version leaves room for typos. That might be worth the cost though, so the code is more readable.

MitchBradley:
If it were an open-ended problem I would prefer the systematic approach, but since it is finite and very unlikely to be significantly extended, I prefer the explicit approach.  Figuring out how the macros work is just too hard for very little payback.

MitchBradley:
Also I think the descriptions buy very little.  As it turns out, the previous attempts at adding descriptive text have not been particularly useful.  People don't look at them, and they don't add enough additional information beyond the names.  If you don't "get" /axes/x/max_travel_mm, you won't get the same thing spelled differently.

MitchBradley:
What would be really nice, if we could figure out how to do it, is basically to have "130" be an alias for "/axes/x/max_travel_mm".  The tricky thing is that "RuntimeSettings" works the wrong way.  There should be a way to just get the value associated with a path in the tree, instead of having it regurgitate a string like "/axes/x/max_travel_mm=400".

MitchBradley:
But the problem I have right now is too much going on at once.  There is the memory problem, and the "smooth motion" problem, and the websocket stall problem, and the web installer, and what you are doing.

dymk:
Yeah, agreed on it being a bounded problem - I got rid of the macro, and descriptions (just use the full name for it). I haven't tested this version on actual hardware yet, and I'd like to do that first.

Regarding the memory leak, I've got a branch where I'm replacing raw pointer usage / manual `new` / `delete` with RAII containers and `std::unique_ptr` to make memory ownership more explicit. Do you suspect the memory issue is due to a leak, or rather heap fragmentation / something else?

PR #954, opened 2023-07-05
Rotary dial to select NVS namespace

ThomasWagnerJKU:
This pull request adds support for changing the NVS namespace by means of a rotary dial.
This is useful for e.g. having the same firmware on boards used in different machines or being able to switch different networks easily


bdring:
We have pull request guidelines  

MitchBradley:
Closed as not planned, for the reasons cited above.

PR #963, opened 2023-07-13
Add webapp to release files

breiler:
This will add the index.html.gz to the release build to be used by the web installer


MitchBradley:
I will merge this after we fix the problem that is causing the latest release to fall on its face.

PR #972, opened 2023-07-17
Added shared_fault_pin for stopping on driver faults/stalls

MitchBradley:
See http://wiki.fluidnc.com/en/config/axes#axes  shared_fault_pin for more info


bdring:
How does this extend to future faults? Examples...

 - A coolant fault defined under a spindle such as a laser.
 - A global fault at the top level if someone wants to OR all equipment faults (drivers, coolant, etc.)
 
 We don't need to define them now, but I think this feature should be flexible.
 
 A quick fix would be to make this one a top level fault.

MitchBradley:
Maybe it would be better to have /control/fault_pin instead of /axes/shared_fault_pin .  You could also have /control/estop_pin to do the same thing.

The FaultPin class can be instantiated multiple times on different pin, using different config items.  A FaultPin fires a faultPinEvent when it goes active, so that part "just works".  The thing that needs to be added is a way to inquire if any faults are still active.  Currently the code that looks for that just knows about the one fault pin at config->_axes->sharedFaultPin.

bdring:
/control/fault_pin sounds good to me.

If I activate the fault in Idle I can still run motion. Should it activate the alarm in any state?

MitchBradley:
> If I activate the fault in Idle I can still run motion. Should it activate the alarm in any state?

I have been working on that.

MitchBradley:
Possible objections: It would be nice to see the states of fault pins and in fact all control pins in $limits/show.  But that could get out of hand pretty quickly.  Similarly one might want to display those pin states in the ? report; but same can of worms.

MitchBradley:
I intentionally omitted the fault pin from $limits/show because what about the other control pins?  Should it also show the safety door pin and the macro pins and ... ?  And what about the pin report in ?  Big can of worms.  For ?, we would have to come up with a schema for single-letter naming of future pins like spindle faults.

MitchBradley:
Another way to look at pin states is with $gpio/dump.  It will show everything.

bdring:
I added `$GPIO/Dump` to the wiki. 

http://wiki.fluidnc.com/en/features/commands_and_settings#gpiodump

I added it at the top to show a new proposed format of that page. I have been frustrated by the limitations of the table format of that page. 

It would be similar the supported gcodes page. It makes better use of the nav panel.

http://wiki.fluidnc.com/en/features/supported_gcodes

MitchBradley:
Something is wrong with the markup.  The first sentence ends with "$" in the rendered version, omitting the rest of the sentence that is present in the markdown text.

bdring:
Thanks

The renderer in edit mode shows it.

That must have been missing for a while. I need to escape the <> characters in $\<text\>.

PR #981, opened 2023-07-26
New alarms

MitchBradley:
This is a revamp of the alarm handling code, setting the stage for finer-grained alarm states like separate spindle failure and mpos-unknown states.

At present, fine-grained states are not implemented, but the new structure should make it easier to implement them.  The new structure unwinds the previously-byzantine code in mc_reset(), separating the jobs of stopping the system and later responding to a user request to reset.  Alarm events are now queued the same as other protocol events, eliminating the old global variables rtAlarm and rtReset whose behavior and interactions were confusing.

### DIFF
--- a/FluidNC/src/Control.cpp
+++ b/FluidNC/src/Control.cpp
@@ -9,7 +9,7 @@
 Control::Control() {
     // The SafetyDoor pin must be defined first because it is checked explicity in safety_door_ajar()
     _pins.push_back(new ControlPin(&safetyDoorEvent, "safety_door_pin", 'D'));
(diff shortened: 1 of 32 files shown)


MitchBradley:
For G2 and G3 arcs, the soft limit is now tested at the arc level instead of in the individual segments.  That makes it possible to reject the entire motion instead of running the part of the arc that was within the limits.

Furthermore, when a soft limit is detected, any previously-planned valid moves are allowed to complete, instead of issuing a feedhold as soon as the parser notices an out-of-limit GCode block.

bdring:
I was testing soft limits alarms. I did a move to cause alarm. When I cleared the alarm my `startup_line` ran. 

This does not happen with v3.7.4

PR #986, opened 2023-08-05
Create FYSECT_E4_laser.yaml

halaali:
The example on the wiki is highly incomplete and inadequate for a laser cutter. Therefore I created this config for my K40 laser cutter.


bdring:
We are moving the config files. I put your file in this repo

https://github.com/bdring/fluidnc-config-files/tree/main/contributed/FYSETC_E4

PR #994, opened 2023-08-11
P delta update

bdring:
Adding Delta Kinematics

### DIFF
--- a/FluidNC/src/Kinematics/ParallelDelta.cpp
+++ b/FluidNC/src/Kinematics/ParallelDelta.cpp
@@ -77,6 +77,7 @@ namespace Kinematics {
         handler.item("homing_mpos_radians", _homing_mpos);
         handler.item("soft_limits", _softLimits);
         handler.item("max_z_mm", _max_z, -10000.0, 0.0);  //
+        handler.item("use_servos", _use_servos);
     }
 
     void ParallelDelta::init() {
@@ -282,6 +283,11 @@ namespace Kinematics {
     }
 
     bool ParallelDelta::kinematics_homing(AxisMask& axisMask) {
+
+        // only servos use custom homing. Steppers use limit switches
+        if (!_use_servos)
+            false;
+
         auto axes   = config->_axes;
         auto n_axis = axes->_numberAxis;
 
--- a/FluidNC/src/Kinematics/ParallelDelta.cpp
+++ b/FluidNC/src/Kinematics/ParallelDelta.cpp
@@ -81,7 +81,19 @@ namespace Kinematics {
 
     void ParallelDelta::init() {
         // print a startup message to show the kinematics are enabled. Print the offset for reference
-        log_info("Kinematic system: " << name());        
+        log_info("Kinematic system:" << name() << " soft_limits:" << _softLimits);
+
+        auto axes   = config->_axes;
+        auto n_axis = config->_axes->_numberAxis;
+
+        // warn about axissofy limits
+        for (int axis = 0; axis < n_axis; axis++) {
+            if (axes->_axis[axis]->_softLimits) {
+                log_warn(" All soft_limits configured in axes should be false");
+                break;
+            }
+        }
+
         init_position();
     }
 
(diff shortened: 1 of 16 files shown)


FluidNC/src/Kinematics/ParallelDelta.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include "../Limits.h"  // ambiguousLimit()
+#include "../Machine/Homing.h"
+
+#include "../Protocol.h"  // protocol_execute_realtime
+
+#include <cmath>
+
+/*
+  ==================== How it Works ====================================
+  On a delta machine, Grbl axis units are in radians
```
MitchBradley, line 13: I think it would be clearer to say "motor units are in radians".  I'm not used to seeing the phrase "Grbl axis units" so it doesn't convey a clear meaning to me.

FluidNC/src/Kinematics/ParallelDelta.cpp
```diff
@@ -0,0 +1,363 @@
...
+  The kinematics converts the cartesian moves in gcode into
+  the radians to move the arms. The Grbl motion planner never sees
+  the actual cartesian values.
+
+  To make the moves straight and smooth on a delta, the cartesian moves
+  are broken into small segments where the non linearity will not be noticed.
+  This is similar to how Grgl draws arcs.
```
MitchBradley: Grbl not Grgl.

FluidNC/src/Kinematics/ParallelDelta.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+  If you request MPos status it will tell you the position in
+  arm angles. The MPos will report in cartesian values using forward kinematics. 
+  The arm 0 values (angle) are the arms at horizontal.
+
```
MitchBradley, line 25: "For mpos reporting, the motor position in steps is proportional to arm angles in radians, which is then converted to cartesian via the forward kinematics transform.  Arm angle 0 means horizontal."

FluidNC/src/Kinematics/ParallelDelta.cpp
```diff
@@ -0,0 +1,363 @@
...
+    static float last_angle[MAX_N_AXIS]     = { 0.0 };  // A place to save the previous motor angles for distance/feed rate calcs
+    static float last_cartesian[MAX_N_AXIS] = { 0.0 };  // A place to save the previous motor angles for distance/feed rate calcs
+
+    void ParallelDelta::group(Configuration::HandlerBase& handler) {
+        handler.item("crank_mm", rf, 50.0, 500.0);
+        handler.item("base_triangle_mm", f, 20.0, 500.0);
+        handler.item("linkage_mm", re, 20.0, 500.0);
+        handler.item("end_effector_triangle_mm", e, 20.0, 500.0);
+        handler.item("kinematic_segment_len_mm", _kinematic_segment_len_mm, 0.05, 20.0);  //
```
MitchBradley, line 76: ISTM that "kinematic_" is somewhat redundant, so perhaps just "segments_per_mm"?

FluidNC/src/Kinematics/ParallelDelta.cpp
```diff
@@ -0,0 +1,369 @@
...
+        return false;
+    }
+
+    // TO DO. This is not supported yet. Other levels of protection will prevent "damage"
+    bool ParallelDelta::invalid_arc(
+        float* target, plan_line_data_t* pl_data, float* position, float center[3], float radius, size_t caxes[3], bool is_clockwise_arc) {
+        return false;
+    }
+
+    // copied from Cartensian. Needs to be optimized for parallel delta.
```
MitchBradley: Cartesian not Cartensian

FluidNC/src/Kinematics/ParallelDelta.cpp
MitchBradley: So, are the motor units steps_per_mm actually steps_per_radian?  If so, would it make sense to have the motor units in revolutions so the steps_per value come out simpler?

bdring:
Thanks

PR #995, opened 2023-08-13
Added G38.6-9 to force probe units in mm

MitchBradley:


breiler:
Just curious, how will this work from a sender perspective?

If I issue a probe command when it's in G20 modal state, FNC may convert my parameters to mm and report back the probed position in mm if the `forceG21` is set to true?     

How will the sender know this?

bdring:
You should test it. The reporting should always respect the report_inches config value.

MitchBradley:
> Just curious, how will this work from a sender perspective?
> 
> If I issue a probe command when it's in G20 modal state, FNC may convert my parameters to mm and report back the probed position in mm if the `forceG21` is set to true?
> 
> How will the sender know this?

G38.6 through G38.9 are new commands; the old probe commands G38.2-G38.5 are unchanged.  If the sender continues to use the old commands, things will behave as always.  The new commands expect that the parameters on the probe line are given in G21 units, regardless of the modal state.  The reported position is not dependent on the modal state, but rather on the "report_inches" (previously $10) setting.  Therefore the existing sender code that already knows how to interpret positions should "just work".

The idea here is that the old commands will continue to be used in contexts where the G20/G21 mode is explicitly established and known, i.e. within larger GCode programs or macros that establish the modal units and depend on them before and after the embedded probe command.  The new commands can be used for UI probe buttons, where it is desirable to run an isolated probe cycle by pressing a button, without the side effect of changing modal units that might have been previously established by some operation that is unrelated to the button.  If the UI wishes to configure that probe cycle in, say, inches, it is free to do so, but it is responsible for converting the G38.6 arguments to mm.  (Or it can continue to use the old commands and deal with the modal problems in other ways.)

In any case, the reports will continue to depend on "report_inches", not the modal units.

Note that UI attempts to sense the current modal units are inherently unreliable.  A UI agent can know when it sends G20 or G21, but if a different agent on another channel (say WebUI or a pendant) sends one, the knowledge is not available immediately.  If auto-reporting is enabled, the agent will receive a message quickly.  If auto-reporting is not enabled, the agent can send $G to ask, or wait for a periodically-scheduled $G poll.  But regardless of when the resulting [GC: report is received, there will always be a small window of uncertainty because agent2 could send G20 between the time that agent1 determined the modal units and sent the probe command.

All this is an unfortunate interaction between the existence of modal units, the Grbl protocol with its poorly-synchronized command/response/status characteristics and, and the desire to allow multiple UI agents.

breiler:
Thanks for the explanation Mitch. I have been battling probing scenarios with different units in UGS and I was worried that I missed something.

MitchBradley:
So does this new feature help you?  And is there some way it could work even better?  I think that it can be used to solve the problem with WebUI, but perhaps I am missing something.

MitchBradley:
With legacy GRBL you typically do not have the "other UI agent" ambiguity.

breiler:
Yes, so that makes things easier. But I was thinking of incorporating the G20/G21 commands to the probing script to make it more explicit. UGS should be able to restore the parser state once the probing has been finished.

MitchBradley:
Indeed.  I faced the same problem with an auxiliary jog widget that I wrote for cncjs / g2core.  g2core does not have $j so the temporary-mode thing was not available.  Originally I tried to optimize it so that it only sent the G20/1 brackets if necessary, but there were sporadic failures due to wrong or untimely knowledge of the current mode.  I eventually changed it so the initial pre-jog modal was always present and the restore-to-original post-jog modal was optional.  That would sometimes fail to restore the mode, but that failure was less bad than jogging with the wrong units.

Knowing the mode is also complicated by the pipelining of sending a long move, followed by a mode change.  That is rare, but theoretically possible.  Grbl resists efforts to make a totally-reliable UI.

PR #998, opened 2023-08-15
Uniform (and not redundant) treatment of "0x" with to_hex()

vector76:
image

to_hex()  produces strings beginning with "0x" and in almost all usage, it is preceded by "0x", producing double "0x0xA1B2C3D4".

To make coherent, "0x" is removed from the to_hex() function and added in the two places where it is not already present in the caller.


MitchBradley (commented):
I would have been tempted to do it the other way - remove the external 0x and leave it in to_hex

vector76:
That is actually my preference too, but I wasn't sure what was "intended".  I'll update to be that way.

PR #1004, opened 2023-08-24
Add an option to enable/disable SSDP and mDNS to save memory

vector76:
SSDP can momentarily use a considerable amount (11k at least) of heap.

This change adds an option (enabled by default, so no change to default behavior) to allow disabling of SSDP and mDNS for those who want to save memory and go by IP address only.

### DIFF
--- a/FluidNC/src/WebUI/TelnetServer.cpp
+++ b/FluidNC/src/WebUI/TelnetServer.cpp
@@ -9,14 +9,14 @@
 #ifdef ENABLE_WIFI
 
 namespace WebUI {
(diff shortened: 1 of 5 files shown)


bdring:
I think the enum should be Enable and Disabled. The setting would be renamed to make that work.

vector76:
Ok, I think this is what you mean?

MitchBradley:
We already have $Bluetooth/Enable, $Telnet/Enable, and $HTTP/Enable, with values On and Off.  The naming should be similar.  Perhaps $Sta/SSDP/Enable

bdring:
Merged....Thanks @vector76 !

PR #1018, opened 2023-09-07
Store startup log in RTC DRAM

MitchBradley:
... so it can be preserved across panic resets

With this change, $ss will show the current startup messages on a clean startup, but if FluidNC panics and reboots, $ss will show the startup messages that preceded the panic, prefaced with 
```
[MSG:ERR: Showing startup log from previous panic]
```

This patch also reduces FluidNC's RAM usage, since the startup log is no longer holding main RAM.


bdring:
I tested with a bad config and the $ss report looks good

I tested with a missing config and the report show the default config, which it what I would expect.

OK to merge.

PR #1021, opened 2023-09-10
Status outputs

bdring:
Add status (idle, run, hold, alarm) output pins.


MitchBradley:
I think we should add a report_interval_ms config item to Channel subclasses.  That would eliminate the need to send an initial message.  UartChannels on auxiliary UARTs could set it to nonzero and they would not need a TX pin, allowing one-pin status monitoring dongles.

PR #1030, opened 2023-09-23
adds on_idle macro which allows chaining or looping gcode files

vector76:
On transition to Idle when a job finishes, launch the on_idle macro.

A gcode file can assign the macro value to determine its successor, or set it blank if it is the last job in the chain.

This allows looping and chaining files together.

### DIFF
--- a/FluidNC/src/Machine/Macros.cpp
+++ b/FluidNC/src/Machine/Macros.cpp
@@ -20,6 +20,7 @@ Macro Macros::_macro[n_macros]               = { { "macro0" }, { "macro1" }, { "
 Macro Macros::_after_homing                  = { "after_homing" };
 Macro Macros::_after_reset                   = { "after_reset" };
 Macro Macros::_after_unlock                  = { "after_unlock" };
+Macro Macros::_on_idle                       = { "on_idle" };
 
 MacroEvent macro0Event { 0 };
 MacroEvent macro1Event { 1 };
--- a/FluidNC/src/Machine/Macros.h
+++ b/FluidNC/src/Machine/Macros.h
@@ -42,6 +42,7 @@ namespace Machine {
         static Macro _after_homing;
         static Macro _after_reset;
         static Macro _after_unlock;
+        static Macro _on_idle;
 
         Macros() = default;
 
@@ -59,6 +60,7 @@ namespace Machine {
             handler.item(_after_homing._name.c_str(), _after_homing._gcode);
             handler.item(_after_reset._name.c_str(), _after_reset._gcode);
             handler.item(_after_unlock._name.c_str(), _after_unlock._gcode);
+            handler.item(_on_idle._name.c_str(), _on_idle._gcode);
         }
 
         ~Macros() {}
--- a/FluidNC/src/Protocol.cpp
+++ b/FluidNC/src/Protocol.cpp
@@ -779,8 +779,12 @@ void protocol_do_cycle_stop() {
                 sys.suspend.bit.holdComplete = true;
                 sys.state                    = State::SafetyDoor;
             } else {
+                State prev_state  = sys.state;
                 sys.suspend.value = 0;
                 sys.state         = State::Idle;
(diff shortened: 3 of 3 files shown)


V1EngineeringInc:
This would be good for sand tables to keep playing files.

bdring:
The G4 dwell state is idle. That could cause problems.

vector76:
Okay that is weird to me because dwell does not mean "Idle" to me at all, but yes, it appears that G4 dwell (within a job) tries to launch the macro file.  It didn't succeed (not sure why) but it is a possible risk.

Another option (which might be a good idea anyway) is to automatically clear the on_idle macro at the start of any job (launched by macro or not).  Then the default behavior is to terminate and to chain another file the gcode must specify the on_idle macro within the file.

Then if people specify the on_idle macro at the end, they will not trip over G4 doing weird things mid-job.  Someone could always assign the on_idle macro near the beginning and see strange behavior on G4 unless there were some other indicator, like the file stream channel being open.

bdring:
It is what Grbl does. It should be a special state, but that breaks compatibility.

V1EngineeringInc:
I think G4 is normally used for things like letting your spindle spin up before cutting and even some lasers to get up to full power before cutting.

vector76:
I've racked my brain and concluded Grbl is just wrong.  And G4 is not a special state, it's just running.

Idle should mean that the machine is idle, not running, and for example I can change out the cutting bit for the next project.  It's bad if someone were using those new status outputs and thought it was safe and it suddenly came to life unexpectedly.

I could not think of a scenario where a sender would be depending on Grbl's G4=Idle behavior and would break when G4=Run.  Every case I could think of favors G4=Run.  But maybe I am not imaginative enough.

bdring:
It is not moving, so it is idle.

If you turn on the spindle it stays in idle.

I understand your opinion, but gcode never knows what is coming next If anything. 

Lots of tiny idles can occur during sending.

MitchBradley:
I agree that Grbl's state transitions are troublesome. It is a result of the smart sender / dumb controller running on 8 bit micro model.

Idle means that the stepping engine has nothing to do right now. There is no way to know that more work related to the same job will be coming soon. A sender might send some, or not. To fix this there needs to be some form of job bracketing.

MitchBradley:
[The Jobs branch](https://github.com/bdring/FluidNC/tree/Jobs) has my work in progress toward job context management.  Unfortunately it has merge conflicts with Devt so I can't PR it just yet.  The merge conflicts are not severe, but they do require careful analysis.

It should be possible to build directly from the Jobs branch to try it out.  As I recall, it was possible to create a macro that runs a sequence of files and/or other macros.

vector76:
> gcode never knows what is coming next If anything.

> There is no way to know that more work related to the same job will be coming soon.

For macro on transition to idle, I get it, you don't really know if it's truly finished, so idle is somewhat indeterminate.  Perhaps it would make sense if it were "on_sd_file_complete" instead which targets the actual trigger event of interest.

But as for G4 being considered "idle" you describe how it is now but I haven't yet heard whether you prefer for G4 to be "Idle" and not "Running".  You don't have to predict the future to know that a G4 command is executing _right now_.  It's just a choice of what you call it.

Or maybe it's just that Grbl does it that way and Idle vs. Running states are junk anyway so who cares.

MitchBradley:
G4 should continue to work the same as before in the context of a sender, otherwise we will run into problems with senders that try to model grbl state transitions (yes there are some; CNCjs does it in some circumstances).  When we have job contexts, we can fix that when a job is active.  Existing/unmodified senders will not establish a job context so they will not see a difference.

Right now we do not have a proper "Running" state that means what we all want it to mean.  We have GRBL's "Cycle" state which means what it means, entangled with the entire ecosystem of senders.  G4 is not the only place this comes up - M0 and M6 have to be considered too, and there is a possible interaction with GCode errors that can cause trouble.

vector76:
Thank you for taking the time to explain that.  I find it helpful.

PR #1040, opened 2023-10-09
Fixed bug in JSONEncoder when trying to create a member whose value is an object

michmela44:
This fix addresses an issue in the JSONEncoder.

Currently, if you want to create JSON of the form:

```
{
    "data" : {
        "member" : "value"
    }
}
```

The expectation is to create it with code like this:

```
     JSONencoder j(true, &out);
     j.begin();
     j.begin_member("data");
     j.begin_object();
     j.member("member", "value");
     j.end_object();
     j.end();
```
However, this code currently creates JSON with an extra comma, making the output invalid, like this:

```
{
      "data" :, {
          "member" : "value"
      }
}
```

I added a new method called `begin_member_object()`. This method combines what `begin_member() ` and `begin_object()` did, but without the extraneous comma. The new way to create that JSON would be:

```
     JSONencoder j(true, &out);
     j.begin();
     j.begin_member_object("data");
     j.member("member", "value");
     j.end_object();
     j.end();
```

`begin_member()` was never directly called in the FluidNC code, so I made this method private as it only has use as it is used by the JSONEncoder itself. 

This also means that this has no effects on FluidNC currently, however this will come into play creating the more complex data structures required by WebUI 3.

### DIFF
--- a/FluidNC/src/WebUI/JSONEncoder.cpp
+++ b/FluidNC/src/WebUI/JSONEncoder.cpp
@@ -149,6 +149,16 @@ namespace WebUI {
         add(']');
     }
 
+    // Begins the creation of a member whose value is an object.
+    // Call end_object() to close the member
+    void JSONencoder::begin_member_object(const char* tag) {
+        comma_line();
+        quoted(tag);
+        add(':');
+        add('{');
+        inc_level();
+    }
+
     // Starts an object with {.
     // If you need a named object you must call begin_member() first.
     void JSONencoder::begin_object() {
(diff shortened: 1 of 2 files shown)


MitchBradley:
In the future please PR against Devt not main.  I rebased this one for you.

michmela44:
Oh. Sorry. I thought I did. On my machine I made my branch from Devt, maybe I missed something when I created the PR.

MitchBradley:
PRs will hit main unless you click on a different target branch for the compare step.

MitchBradley:
They go (main) < (your branch) regardless of what your branch started from.

michmela44:
Ok, Thanks. I'll make sure to check that in the future.

PR #1042, opened 2023-10-10
Reinstated a macro fix that got lost in a git mixup

MitchBradley:


bdring:
Tested. Works for me on the macro config that was failing before.

PR #1045, opened 2023-10-12
Change hidden password string to abide by minimum length

michmela44:
For editing the WebUI, we currently send a string to hide passwords of length 6, "******".

WebUI 2 doesn't warn about min length constraints, but WebUI3 does.

image

This PR adds two asterisks to abide by the hardcoded length of 8, but in reality the better fix is to use the _minLength field and make sure the fake string meets the constraints.

I tried to do this with std::string(_minLength, '*') but I couldn't get it to end up correct on the web side, so I reverted back to hard-coded. It could just be my lack of C/C++ experience, so maybe someone can change it to something more robust using _minLength.

### DIFF
--- a/FluidNC/src/Settings.cpp
+++ b/FluidNC/src/Settings.cpp
@@ -262,10 +262,10 @@ static bool isPassword(bool (*_checker)(char*)) {
 
 const char* StringSetting::getDefaultString() {
     // If the string is a password do not display it
-    return (_checker && isPassword(_checker)) ? "******" : _defaultValue.c_str();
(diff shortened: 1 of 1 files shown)


V1EngineeringInc:
If I am not mistaken, this is the other half of the changes needed to allow for Esp3dUI V3 with #1044. 

If that gets working, it might help lighten the Dev load of FluidNC? Seems to be getting more and more widely adopted and I am seeing it getting used a lot more.

V1EngineeringInc:
Cool, thanks!

PR #1087, opened 2023-12-15
Update and rename 3axis_v4.yaml to config.yaml

GdadGarage:


MitchBradley:
This is not an appropriate change for the FluidNC repo.  First, it provides no new functionality, but rather changes one trivial machine-specific numeric value.  Second, it renames a file whose name gives information about its meaning to a generic name that gives no clue as to what it is for.  Additions to the example_configs folder should provide important new examples for specific situations that are not already addressed by existing files, with names that give a clue as to the controller and machine configuration that is being addressed.

PR #1095, opened 2023-12-21
New Refr button in Tablet

MitchBradley:
2. Permit datetime field omission in file structure
3. Use jog instead of G0 for probe retraction


bdring:
#3 is nice

PR #1097, opened 2023-12-22
Report file changes

MitchBradley:
Sends [MSG: Files changed] to all channels when something changes on a file system, such as an upload, delete, or rename.  That lets a UI automatically refresh its list of files.  The included version of WebUI responds to that message and refreshes the file selectors.


bdring:
I don't think this is going out on all channels. The problem is similar to the ALARM: X message.

bdring:
It was due to `all_messages: false`

PR #1099, opened 2023-12-30
Better structuring of ESPxxx command response in WebServer

MitchBradley:
... eliminates problems with first-time connection fails Also run the protocol engine during OLED delays, for similar reasons.


MitchBradley:
This is superseded by recent changes in the MSG_JSON branch

PR #1104, opened 2024-01-04
Channel pin extender

MitchBradley:


MitchBradley:
Changelog for this PR:
* Autoreporting handles pin changes better, including the probe pin
* Realtime characters can be UTF-8 encoded, allowing a larger range of realtime events including 16-bit codes
* Channels handle communication with I/O extenders, with flow control of various outgoing messages and events for incoming messages like extender pin state changes
* Message level can be controlled on a per-channel basis via message_level: in the channel section.  The global $message/level sets the maximum level for all channels, and an individual channel can further restrict the level.
* Added new commands $Msg/Uart0, $Msg/Uart1, $Log/Msg, $Log/Error, $Log/Warn, $Log/Info, $log/Debug, $Log/Verbose to send messages at various levels.  Useful for testing.

*

PR #1115, opened 2024-01-20
Clean up before fall through to ap mode

BarbourSmith:
This pull request causes the ESP to clean up and reset the wifi after failing to connect to a wifi network in STA mode before switching to AP mode. 

Without this change ESP32-S3 would report that a AP network had been created, but no network would actually be created. 

More information is available here in the discord thread:

https://discord.com/channels/780079161460916227/1073102734020067328

Tested to not impact the behavior of the ESP32 which was already working correctly before this PR.


BarbourSmith:
While testing this PR I came across what I think is an unrelated issue, but it is close enough that I think it's worth mentioning. When changing values in the settings I get` Error500: No Response` as a popup. The settings seem to be saved correctly. 

image

MitchBradley:
It turns out that am looking into that No response thing right now.  There are some super-gnarly issues surrounding the way that webui sends commands over HTTP and gets back responses.  The crux of the problem is that some Grbl commands are synchronous and some are not, which makes it really hard to do the right thing in all cases, considering that webui uses synchronous HTTP GET for some things, and expects other responses to come back over a websocket.  It is so complicated that I am writing a wiki page to describe the problem in all its gory.

BarbourSmith:
I've flirted around edges of messing with the synchronous vs asynchronous GRBL commands and getting everything to work right there is TRICKY. I can see how something like that could cause the behavior of showing a popup, but also working correctly. I was 99% sure that it was unrelated, but I wanted to mention it just in the off chance that I was missing something and I broke it somehow 😛

MitchBradley:
The code in this PR was incorporated into FluidNC via a different PR.  Thanks @BarbourSmith .

PR #1101, opened 2023-12-31
WebUI-v3 support

michmela44:
This pull request contains the core changes required for WebUI-v3 to work. This adds support for WebUI-v3, but does not remove anything related to the current WebUI

ESP3D WebUI does not yet contain the code to run correctly for FluidNC. The changes required for this can be found at https://github.com/michmela44/ESP3D-WEBUI/tree/3.0-FluidNCDev. Interface work still in progress.

PR #1118, opened 2024-01-26
Web ui3

MitchBradley:


michmela44:
The changes in commit [3be3ae5](https://github.com/bdring/FluidNC/pull/1101/commits/3be3ae58a157f305ec3906c7a200602fe495bca4) are important as well.

Without them, the About page will not load correctly

PR #1128, opened 2024-02-07
Support for Danfoss VLT 2800 VFD

whosmatt:
Fully tested except CCW direction since I do not have reversing enabled in my VFD.


whosmatt:
I should mention that the code contains type punning which may not be supported when the c++ standard is changed, although it seems to be supported almost everywhere

MitchBradley:
@whosmatt We merged your VFD PR.  Could you please update the FluidNC Wiki to add information about this VFD in [this section](http://wiki.fluidnc.com/en/config/config_spindles#variable-frequency-drives-vfd-controlled-with-rs485) .  Information about wiki contributions is [here](http://wiki.fluidnc.com/en/wiki_contributions)

Issue #1130, opened 2024-02-11
Problem: "Toolsetter Pin" init-message is never displayed if "Probe Pin" is configured.


nicolas-lang:
### Wiki Search Terms

I did not check the wiki, I already know the problem and it's solution and it's in the sourcecode.

### Controller Board

-

### Machine Description

-

### Input Circuits

```yaml
-
```

### Configuration file

```yaml
-
```

### Startup Messages

```text
-
```

### What happened?

The info [message](https://github.com/bdring/FluidNC/blob/09fd57261d65e2bf7d60834c7f3af60a2e0c569f/FluidNC/src/Probe.cpp#L28) for the toolsetter pin is never displayed if the [message](https://github.com/bdring/FluidNC/blob/09fd57261d65e2bf7d60834c7f3af60a2e0c569f/FluidNC/src/Probe.cpp#L18) for the probe pin is displayed.

### Other Information

It is caused by a redundant `show_init_msg = false;` in [L19](https://github.com/bdring/FluidNC/blob/09fd57261d65e2bf7d60834c7f3af60a2e0c569f/FluidNC/src/Probe.cpp#L19)

PR #1131, opened 2024-02-11
fixes "Toolsetter Pin" init-messages

nicolas-lang:
The info [message](https://github.com/bdring/FluidNC/blob/09fd57261d65e2bf7d60834c7f3af60a2e0c569f/FluidNC/src/Probe.cpp#L28) for the toolsetter pin is never displayed if the [message](https://github.com/bdring/FluidNC/blob/09fd57261d65e2bf7d60834c7f3af60a2e0c569f/FluidNC/src/Probe.cpp#L18) for the probe pin is displayed.

The issue is caused by a redundant show_init_msg = false; in [L19](https://github.com/bdring/FluidNC/blob/09fd57261d65e2bf7d60834c7f3af60a2e0c569f/FluidNC/src/Probe.cpp#L19)

This PR removed the duplicate.

### DIFF
--- a/FluidNC/src/Probe.cpp
+++ b/FluidNC/src/Probe.cpp
@@ -8,14 +8,13 @@
 
 // Probe pin initialization routine.
 void Probe::init() {
-    static bool show_init_msg = true;  // used to show message only once.
+    static bool show_init_msg = true;  // used to show messages only once.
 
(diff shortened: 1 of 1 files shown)


nicolas-lang:
Fixes #1130

PR #1142, opened 2024-02-20
Check for config

bdring:
Gives a config error and message if not configured.

### DIFF
--- a/FluidNC/src/WebUI/WebSettings.cpp
+++ b/FluidNC/src/WebUI/WebSettings.cpp
@@ -658,12 +658,8 @@ namespace WebUI {
             return Error::InvalidStatement;
         }
 
-        if (!parameter || *parameter == '\0' || !strchr(parameter, '>')) {
-            return Error::InvalidValue;
-        }
-
         auto opath = strchr(parameter, '>');
-        if (*opath == '\0') {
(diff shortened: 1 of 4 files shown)


MitchBradley (commented):
Wouldn't it be better to move the is configured check down deeper so it automatically applies to anything that tries to use the sd?

MitchBradley (commented):
And "primary sd:none" in ShowFwInfo()

FluidNC/src/WebUI/WebSettings.cpp
```diff
@@ -633,6 +659,10 @@ namespace WebUI {
     }
 
     static Error renameObject(const char* fs, char* parameter, AuthenticationLevel auth_level, Channel& out) {
+        if (fs == sdName && !config->_sdCard->is_configured(out)) {
+            return Error::InvalidStatement;
+        }
+
```
MitchBradley, line 660: Another reason for putting the check in FluidPath is that the fs argument to renameObject and friends is just the default, which can be overridden with an explicit prefix like /localfs/ or /sd/ .  Thus checking based on fs == sdName will fail to catch some uses of SD, and will wrongly complain if the sd default is overridden via an explicit /localfs/ prefix in the parameter.

PR #1146, opened 2024-02-27
Revert ackwait for JSON over serial.  It was a bad idea.

MitchBradley:


MitchBradley:
This doesn't work for pendants

PR #1149, opened 2024-02-29
Create devcontainer.json

JamieKaybg:

PR #1150, opened 2024-02-29
Merge hook1

JamieKaybg:


MitchBradley:
Does this replace #1149?  What is it for?

oliof:
the submitters profile smells of fake. I would suggest to report and block.

MitchBradley:
Possibly bogus.  Submitter has not answered questions.

PR #1175, opened 2024-03-23
Prevent underflow when recalculating an empty planner buffer

pentacular:
planner_recalculate did not check for an empty planner buffer.

This change adds an early exit guard.

### DIFF
--- a/FluidNC/src/Planner.cpp
+++ b/FluidNC/src/Planner.cpp
@@ -121,6 +121,10 @@ static uint8_t plan_prev_block_index(uint8_t block_index) {
 
 */
 static void planner_recalculate() {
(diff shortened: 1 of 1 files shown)


MitchBradley (commented):
This looks good on the face of it.  Has it been stress-tested?

pentacular:
I discovered the issue by calling protocol_send_event(&feedOverrideEvent, -FeedOverride::CoarseIncrement) every 10ms resulting in Stepper::get_realtime_rate() returning NaN due to a negative max_entry_speed_sqr being used in replanning.

The extent of the stress testing I have done is to run with the fix for about ten minutes.

PR #1188, opened 2024-04-13
Add flag to prevent unneeded delays and confusing debug messages

bdring:


FluidNC/src/Machine/Axes.cpp
```diff
@@ -66,10 +70,13 @@ namespace Machine {
 
         _sharedStepperDisable.synchronousWrite(disable);
 
-        if (!disable && config->_stepping->_disableDelayUsecs) {  // wait for the enable delay
```
MitchBradley, line 69: I think this might be clearer as
```
if (!disable && disabled) {
    disabled = false;
    if (config->_stepping->_disableDelayUsecs) {  // wait for the enable delay
            delay_us(config->_stepping->_disableDelayUsecs);
    }
}
```
Less redundancy and we probably don't need the log message anymore.

PR #1211, opened 2024-05-06
feat: WebUI pops up a dialog to report startup errors

MitchBradley:
You will need to update the index.html.gz in your local filesystem to enable this


craiglink:
looks good to me. 
would it make sense to have an uncompress version of index.html checked in and have the build do the gzip compression? 
it would be easier to track html changes

MitchBradley:
> looks good to me. would it make sense to have an uncompress version of index.html checked in and have the build do the gzip compression? it would be easier to track html changes

It might be but
1. I don't want to do the work
2. It would make it harder for someone to just download the index.html.gz from the repo.  I don't know that people do that - but I also don't know that they don't.

Ideally there would be better versioning of the WebUI.  Version management is a lot of boring work, and the halfhearted switchover to WebUI3 threatens to make any work on WebUI2 obsolete.  I'm having a hard time getting motivated to do much work on WebUI2.  The startup error thing seems worthwhile based on the number of people who ask for help on Discord after having totally missed or ignored startup messages that could have solved their problem.

PR #1212, opened 2024-05-06
define Events as constant 

craiglink:
- Well-known Events are defined as constant reducing some use of RAM
- Various xxxPin and Macro classes are update to support the constant Event pointers
- Pin classes are more objected-orientated by providing accessors to values now made private or protected
- Simple Pin assignment constructors are moved to the header files so the compiler can optimize and inline the assignments

[env:wifi]
RAM:   [===       ]  27.3% (used 89400 bytes from 327680 bytes)
Flash: [========= ]  91.1% (used 1791125 bytes from 1966080 bytes)

### DIFF
--- a/FluidNC/src/Control.cpp
+++ b/FluidNC/src/Control.cpp
@@ -28,7 +28,7 @@ void Control::init() {
 
 void Control::group(Configuration::HandlerBase& handler) {
     for (auto pin : _pins) {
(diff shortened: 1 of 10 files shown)


MitchBradley:
I think C++ is poorly named.  It should be "const C const + const +"

PR #1214, opened 2024-05-06
remove dependency on the global Serial object

craiglink:
remove dependency on the global Serial object which was only used in OLEDDisplay to output a deprecated message.

PR to original OLEDDisplay repo has been requested as well: https://github.com/ThingPulse/esp8266-oled-ssd1306/pull/408/files

Reduces RAM and Memory usage

[env:wifi] // before
RAM:   [===       ]  27.3% (used 89400 bytes from 327680 bytes)
Flash: [========= ]  91.1% (used 1791125 bytes from 1966080 bytes)

[env:wifi] // after
RAM:   [===       ]  27.2% (used 89072 bytes from 327680 bytes)
Flash: [========= ]  90.9% (used 1786609 bytes from 1966080 bytes)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -29,7 +29,7 @@ build_flags =
 	-Wno-unused-function
 lib_deps =
 	TMCStepper@>=0.7.0,<1.0.0
-	thingpulse/ESP8266 and ESP32 OLED driver for SSD1306 displays@^4.2.1
+	ESP8266 and ESP32 OLED driver for SSD1306 displays=https://github.com/craiglink/esp8266-oled-ssd1306#NO_GLOBAL_SERIAL
 bt_deps =
 	BluetoothSerial
 wifi_deps =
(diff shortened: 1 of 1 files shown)


MitchBradley:
Let's wait to see if they take the PR.  I prefer not to use private repos unless they provide a compelling advantage.  In this case, eliminating the global Serial saves some memory but not enough to get excited about.

MitchBradley:
What I *would* like is a version of the oled lib that does not use String.  That would be worth a private repo.

bdring:
I would not be sad if the internally coded OLED was deprecated.

MitchBradley:
I took a swag at eliminating String from the lib. It took about 10 minutes

MitchBradley:
The easiest solution would be to just pin the library version to 4.4.1, the version before the change that introduced the global Serial dependency.  The changes from 4.4.1 to 4.5.0 look irrelevant to us.  We can declare that our built-in OLED support is not going to be improved going forward and it is what it is.

robercy:
> I would not be sad if the internally coded OLED was deprecated.

Hi, with fluid dial on desk, this is a good idea.

MitchBradley:
FluidDial cannot connect until after the WiFi connection attempt completes (or fails).  I have some systems where OLED shows the status of the WiFi connection attempt, which is useful information when you are waiting to see if FluidDial is going to connect.

MitchBradley:
Superseded by #1215

PR #1237, opened 2024-06-10
Catch overcurrent only while jogging or running files

BarbourSmith:
This error at other times is usually a false reading and it's tripping folks up


BarbourSmith:
Sorry, wrong target 🤦‍♂️😔

PR #1248, opened 2024-07-04
enable conditional compilation to reduce image size

craiglink:
If you are open to it, I'd like to have these conditional compilation settings merged.  For those of us with 6Pack External or 6x CNC Controllers there won't be any other stepper drivers.  Additionally there isn't an OLED display easily added.  Thus to save some space and allow room for other experimentation, I'd like to have the option to not include the supporting code and 2 external libraries in the build.

I don't see there being a need to build and share this binary for others to download, but I'll leave that decision to you :)

- enable conditional compilation for only External Drivers
- enable conditional compilation for OLED Display

The size of the wifi-external build is
RAM:   [==        ]  19.5% (used 63772 bytes from 327680 bytes)
Flash: [========= ]  86.6% (used 1703465 bytes from 1966080 bytes)

The size of the original wifi build
RAM:   [==        ]  19.7% (used 64564 bytes from 327680 bytes)
Flash: [========= ]  91.5% (used 1799137 bytes from 1966080 bytes)

### DIFF
--- a/FluidNC/esp32/tmc_spi.cpp
+++ b/FluidNC/esp32/tmc_spi.cpp
@@ -1,3 +1,4 @@
+#ifndef EXTERNAL_STEPPERS_ONLY
 // Copyright (c) 2022 Mitch Bradley
 // Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
 
@@ -92,3 +93,4 @@ uint32_t TMC2130Stepper::read(uint8_t reg) {
 
     return data;
 }
+#endif
--- a/FluidNC/src/Motors/Dynamixel2.cpp
+++ b/FluidNC/src/Motors/Dynamixel2.cpp
@@ -1,6 +1,6 @@
 // Copyright (c) 2020 -	Bart Dring
 // Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
-
+#ifndef EXTERNAL_STEPPERS_ONLY
 /*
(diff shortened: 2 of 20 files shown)


MitchBradley (commented):
I am hesitant about this because where does it end?  People will want to remove spindles and kinematics systems and ...

Have you tried just excluding the files using build_src_filter in platformio.ini ?

craiglink:
I’m not familiar with build_src_filter. I’ll take a look at it

MitchBradley:
You can make a custom env like this
```ini
[env:fewer_motors]
extends = env:wifi
build_src_filter = ${env:wifi.build_src_filter} -<src/Motors/Solenoid.cpp> -<src/Motors/RcServo.cpp>
```
Then you can stash that change to platformio.ini and apply it anytime.  That technique can be used to exclude any set of files that you want, without needing to bracket files with ifdefs or define the ifdef targets in a build rule.

craiglink:
That works
`build_src_filter = ${common_esp32_base.build_src_filter}
  -<esp32/tmc*.c*>
  -<src/Motors/Dynamixel2.cpp> 
  -<src/Motors/*Servo*.cpp> 
  -<src/Motors/Solenoid.cpp> 
  -<src/Motors/StepStick.cpp> 
  -<src/Motors/TMC*.cpp>
  -<src/Motors/Trinamic*.cpp> 
  -<src/Motors/UnipolarMotor.cpp>`

PR #1251, opened 2024-07-05
standarized on one delay(ms) call

craiglink:
The code base currently has 3 calls to delay for X number of milliseconds.

This change standarized on the default delay(ms) call which just calls vTaskDelay() under the covers.  It also updates the x86 support to do the same.  

The custom delay_ms(ms) in the code base has the same implementation as delay(ms), so those calls have been updated to just use delay(ms).

### DIFF
--- a/FluidNC/src/Motors/Dynamixel2.cpp
+++ b/FluidNC/src/Motors/Dynamixel2.cpp
@@ -73,7 +73,7 @@ namespace MotorDrivers {
 
         // servos will blink in axis order for reference
         LED_on(true);
-        vTaskDelay(100);
+        delay(100);
         LED_on(false);
     }
 
--- a/FluidNC/src/Motors/RcServo.cpp
+++ b/FluidNC/src/Motors/RcServo.cpp
@@ -23,8 +23,6 @@
 #include "../Limits.h"  // limitsMaxPosition
 #include "RcServoSettings.h"
 
-#include <freertos/task.h>  // vTaskDelay
-
 namespace MotorDrivers {
     // RcServo::RcServo(Pin pwm_pin) : Servo(), _pwm_pin(pwm_pin) {}
 
@@ -92,8 +90,8 @@ namespace MotorDrivers {
             float home_time_sec = (axisConfig->_maxTravel / axisConfig->_maxRate * 60 * 1.1);  // 1.1 fudge factor for accell time.
 
             _disabled = false;
-            set_location();                    // force the PWM to update now
-            vTaskDelay(home_time_sec * 1000);  // give time to move
+            set_location();               // force the PWM to update now
+            delay(home_time_sec * 1000);  // give time to move
         }
         return false;  // Cannot be homed in the conventional way
     }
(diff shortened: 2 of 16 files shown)


FluidNC/src/Motors/Dynamixel2.cpp
```diff
@@ -73,7 +73,7 @@ namespace MotorDrivers {
 
         // servos will blink in axis order for reference
         LED_on(true);
-        vTaskDelay(100);
```
craiglink, line 76: this was probably a previous minor bug that should have been `vTaskDelay(100 / portTICK_PERIOD_MS);`

FluidNC/src/Motors/RcServo.cpp
```diff
@@ -92,8 +90,8 @@ namespace MotorDrivers {
             float home_time_sec = (axisConfig->_maxTravel / axisConfig->_maxRate * 60 * 1.1);  // 1.1 fudge factor for accell time.
 
             _disabled = false;
-            set_location();                    // force the PWM to update now
-            vTaskDelay(home_time_sec * 1000);  // give time to move
```
craiglink, line 96: this was probably a previous minor bug that should have been `vTaskDelay(home_time_sec * 1000 / portTICK_PERIOD_MS);`

FluidNC/src/WebUI/WebServer.cpp
```diff
@@ -860,7 +860,7 @@ namespace WebUI {
                     //Upload write
                     //**************
                 } else if (upload.status == UPLOAD_FILE_WRITE) {
-                    vTaskDelay(1 / portTICK_RATE_MS);
```
craiglink, line 863: this previously would evaluation to 1 or 0 depending on the hardware's clock rate.  Was that intentional?

FluidNC/src/Protocol.cpp
```diff
@@ -124,10 +124,10 @@ char activeLine[Channel::maxLine];
 bool pollingPaused = false;
 void polling_loop(void* unused) {
     // Poll the input sources waiting for a complete line to arrive
-    for (; true; /*feedLoopWDT(), */ vTaskDelay(0)) {
+    for (; true; /*feedLoopWDT(), */ delay(0)) {
         // Polling is paused when xmodem is using a channel for binary upload
         if (pollingPaused) {
-            vTaskDelay(100);
```
craiglink, line 130: this was probably a previous minor bug that should have been `vTaskDelay(100 / portTICK_PERIOD_MS);`

MitchBradley (commented):
This is a good idea but did you see my discord post where I analyzed all the delays and suggested a different approach?  I would rather use an internally-defined delay function instead of going directly to the Arduino framework's delay(), for future portability in case we are finally able to ditch Arduino.

craiglink:
I was debating that as well. I chose to go with the Arduino one because of the number of 3rd party Arduino based libraries included in the project. But I’d be fine other other way as well. Want me to do that with this PR?

craiglink:
The only discord comments I saw from you are the ones I responded too in discord.

MitchBradley:
https://discord.com/channels/780079161460916227/1258511678287646730/1258511678287646730

If you could implement that plan, it would be great.

craiglink:
looks like you beat me to it here - https://github.com/bdring/FluidNC/pull/1255/files

PR #1252, opened 2024-07-05
update library dependency finder to use chain+

craiglink:
Set dependency finder to use [chain+](https://docs.platformio.org/en/latest/librarymanager/ldf.html) so that only needed libraries pulled in.

### DIFF
--- a/FluidNC/src/WebUI/TelnetServer.cpp
+++ b/FluidNC/src/WebUI/TelnetServer.cpp
@@ -1,12 +1,12 @@
 // Copyright (c) 2014 Luc Lebosse. All rights reserved.
 // Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
-#include <ESPmDNS.h>
 #include "../Machine/MachineConfig.h"
 #include "TelnetClient.h"
 #include "TelnetServer.h"
 #include "WebSettings.h"
 
 #ifdef ENABLE_WIFI
+#include <ESPmDNS.h>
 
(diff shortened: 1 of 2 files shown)


FluidNC/src/WebUI/TelnetServer.cpp
```diff
@@ -1,12 +1,12 @@
 // Copyright (c) 2014 Luc Lebosse. All rights reserved.
 // Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
-#include <ESPmDNS.h>
 #include "../Machine/MachineConfig.h"
 #include "TelnetClient.h"
 #include "TelnetServer.h"
 #include "WebSettings.h"
 
 #ifdef ENABLE_WIFI
+#include <ESPmDNS.h>
```
craiglink, line 9: ESPmDNS was being pulled in for bluetooth only and noradio builds

MitchBradley:
Is there a problem with this?  It seems like a good idea in principle.

FluidNC/src/WebUI/TelnetServer.cpp
craiglink: the change in this include makes sense

craiglink:
This chain+ setting worked for a number of builds, but failed on a few additional ones after testing.  Specifically it wasn't finding wifi.h when compiling the custom websockets code.    not sure why so closed the MR until I can figure it out

PR #1258, opened 2024-07-12
support developers having custom builds in platform_override.ini

craiglink:
... and prevent the override from being commited to the project.

This follows other platformio conventions so the developers may have their own custom builds without having to regularly deal with conficts in platformio.ini

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ FluidNC.vcxproj.filters
 *.suo
 packages/
 pio_machine.h
+platformio_override.ini
 project.checksum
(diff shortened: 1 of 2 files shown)


bdring:
Is the `platformio_override.ini` file supposed to be in this PR?

craiglink:
No. That is the point of the line in the .gitignore. Developers can have local configs in their platformio_override.ini which should never be checked in and thus will not conflict with outer platformio.ini changes

MitchBradley:
I tested this and it works as advertised.

PR #1259, opened 2024-07-12
conditional compilation / inclusion of OLED support

craiglink:
- because most boards don't have an OLED display
- the OLED module is retired
- the OLED library uses a fair amount of ROM
- and it uses the String class

... provide an option to not included it.

**build_src_filter** is not an option because of other dependencies in the source code on the OLED class existence.

### DIFF
--- a/FluidNC/src/OLED.cpp
+++ b/FluidNC/src/OLED.cpp
@@ -1,3 +1,4 @@
+#ifndef NO_OLED
 #include "OLED.h"
 
 #include "Machine/MachineConfig.h"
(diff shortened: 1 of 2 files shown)


MitchBradley:
Instead of adding more ifdefs I am looking into eliminating the current ones so all inclusion/exclusion can be done via build_src_filter.  It appears that a combination of weak links and use of the registration pattern may do the trick.  The methods that need to be called from registered modules appear to be init(), poll(), make_settings(), and reset_settings().  Possibly something related to reporting too.

MitchBradley:
This is moot as a result of #1274.  You can now omit many different features by adding, e.g. -<src/OLED.cpp> , to platformio.ini .  OLED is now a "module" that can be excluded that way. Most of the WiFi/network protocols are similarly modularized.

PR #1276, opened 2024-07-30
Fixed $WiFi/ListAPs.

MitchBradley:


bdring:
This works for me, but takes a longer than before.

MitchBradley:
It takes longer the first time, but subsequent times are fast because the results from the last scan are cached.  The longer time is a tradeoff.  In the past, at wifi startup, FluidNC would initiate a scan and cache the results so they are immediately available to listaps.  While the scan was happening, attempts to start WebUI would stall and sometimes fail because the WiFi core was busy with the scan.  The upside, though, was that the scan data was already available when you asked for it.

It seems to me that it is preferable to optimize for the WebUI startup time, since that is typically done on every boot, compared to wifi scan which is only needed during one-time setup.

PR #1282, opened 2024-08-03
Wifi power savings

bdring:
Allows user to set wifi power saving mode.


MitchBradley (approved):

MitchBradley:
It appears that the changes may have been added to the pre-Modules version of WifiConfig.cpp.  I think this needs to be redone atop the current version.  As-is, the merge conflict is severe.

PR #1287, opened 2024-08-05
Atc class

bdring:
New ATC system
- You can use the original method where changing tools just changes spindles.
- You can use an m6_macro
- You can use custom ATC classes


FluidNC/src/Spindles/Spindle.cpp
```diff
@@ -108,6 +120,43 @@ namespace Spindles {
...
     }
 
+    const char* Spindle::atc_info() {  // this can be used in the startup response
+        std::string atc_info = "";
+        if (_atc != NULL) {
+            atc_info = " ATC:" + _atc_name;
+        } else if (!_m6_macro._gcode.empty()) {
+            atc_info = " m6_macro:";
+        }
+        return atc_info.c_str();
```
MitchBradley: This isn't safe.  The atc_info string object goes out of scope and is freed when atc_info() returns, and the c_str() pointer is then pointing to memory that is now in the heap and could be reused at any time.  You could make it `std::string Spindle::atc_info()`

FluidNC/src/Spindles/Spindle.cpp
```diff
@@ -108,6 +120,43 @@ namespace Spindles {
...
+    }
+
+    // pre_select is generally ignored except for machines that need to get a tool ready
+    // set_tool is just used to tell the atc what is already installed.
+    bool Spindle::tool_change(uint32_t tool_number, bool pre_select, bool set_tool) {
```
MitchBradley, line 135: Perhaps call it "current_tool" instead of "set_tool".  "set" as a prefix is usually used as a verb, as in set_ and get_

FluidNC/src/ToolChangers/atc_manual.cpp
```diff
@@ -0,0 +1,215 @@
...
+                return false;
+            }
+
+            float tlo = offset_probe[2] - _tool_setter_offset;
+
+            //log_info("Set TLO:" << tlo);
+            gc_exec_linef(false, Uart0, "G43.1Z%0.3f", tlo);
+
+            move_to_save_z();
+            gc_exec_linef(false, Uart0, "G0X%0.3fY%0.3f", starting_wpos[0], starting_wpos[1]);
```
MitchBradley: I worry that these gc_exec_linef()s could block waiting for the planner, possibly resulting in deadlocks

FluidNC/src/Spindles/Spindle.cpp
bdring: `bool set_tool` means it is coming from a stand alone T<num> rather than an M6 T<num>

Maybe `select_tool` is better?

https://linuxcnc.org/docs/2.8/html/gcode/other-code.html#sec:select-tool

PR #1288, opened 2024-08-07
Bring back current threshold panic but make the threshold really high

BarbourSmith:
We now need to get 150 over current warnings in a row to trigger the alarm instead of just 15 hopefully preventing false alarms


BarbourSmith:
Sorry! Wrong target 😞

PR #1297, opened 2024-08-17
Basic automated ATC implementations (Draft/WIP)

nicolas-lang:
My intention is mostly to offer my implementation for review and maybe as contribution for further automated ATC implementations.
The code is untested but seems to compile, unless I am doing the compile incorrectly.
I am slightly scared in uploading it, as I don't want to brick my ATC-hardware 😉
*Since this PR is not intended to be merged, I'll just target main.*

### Description
The idea is to have the ATC to move to the tool's xy position, and then use a tool changer specific macro to perform any pickup/return moves, operating required outputs to the ATC with M62/63.
This way it should be rather compatible with different hardware only requiring yaml-cfg changes.
The only downside I see with this behaviour right now, would be exposing the ATC-I/Os to the Gcode - but I would consider this acceptable, since GCode already has millions of ways of crashing your machine anyways.

for a simple vertical ATC the pickup macro could look like this:
```gcode
G91 (or alternatively use G90 moves and the variables set up by the ATC: G0 Z#<_tc_tool_z>)
M62 P6 (Triggers air valve to open the ATC )
G4 P0 1.0 (wait for ATC to settle)
G0 Z-50.0 (go to tool-pickup location)
M63 P6 (Triggers air valve to close the ATC )
G4 P0 1.0 (wait for ATC to settle)
G0 Z50.0 (lift tool from fixture)
```
It shouldn't be difficult to adapt the macro for a fork-holder atc that is accessed from the side, or run the spindle for something like the rapidchange ATC.

P.S.
The general implementation of "using macros" is mostly guesswork since i haven't found any related documentation in the wiki, Especially for the macro variables I am not confident at all.

### DIFF
--- a/FluidNC/src/ToolChangers/atc_basic.cpp
+++ b/FluidNC/src/ToolChangers/atc_basic.cpp
@@ -15,11 +15,41 @@ namespace ATCs {
 
     bool Basic_ATC::tool_change(uint8_t new_tool, bool pre_select, bool set_tool) {
         protocol_buffer_synchronize();  // wait for all motion to complete
-        _macro.erase();             // clear previous gcode
-        
-        if (pre_select) { // not implemented
+        _macro.erase();                 // clear previous gcode
+
+        if (pre_select) {  // not implemented
             return true;
         }
+
+        // save current location, so we can return after the tool change.
+        _macro.addf("#<start_x >= #<_x>");
+        _macro.addf("#<start_y >= #<_y>");
+        _macro.addf("#<start_z >= #<_z>");
+
+        //save machine states
+        Macro set_state;
+        Macro restore_state;
+        set_state.addf("M9");  // Disable coolant
+        if (gc_state.modal.coolant.Mist) {
+            restore_state.addf("M7");
+        }
+        if (gc_state.modal.coolant.Flood) {
+            restore_state.addf("M8");
+        }
+        set_state.addf("G21");  // become Metric
+        if (gc_state.modal.units == Units::Inches) {
+            restore_state.addf("G20");
+        }
+        set_state.addf("G90");  // become absolute
+        if (gc_state.modal.distance != Distance::Absolute) {
+            restore_state.addf("G91");
+        }
+        set_state.addf("M5");  // turn off the spindle
+        if (gc_state.modal.spindle != SpindleState::Disable) {
+            restore_state.addf("M3");
+        }
+
+        _macro.addf(set_state._gcode.c_str());
 
         // set_tool is used to update the current tool and reset the TLO to 0
         if (set_tool) {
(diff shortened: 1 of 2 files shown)


nicolas-lang:
Just a short update: 
I have tested a bit and unfortunately it doesn't work yet, I was sooooo confident with my implementation ;-/

I have tested with this config:
```yaml
atc_basic:
  safe_z_mpos_mm: -1.000000
  probe_seek_rate_mm_per_min: 400.000000
  probe_feed_rate_mm_per_min: 40.000000
  ets_mpos_mm: -72.500 -7.500 -45.000
  ets_rapid_z_mpos_mm: 0.000000
  toolreturn_macro: G91&G0 Z-10.0&M62 P6&G4 P0 1.0&M63 P6&G4 P0 1.0&G0 Z10.0
  toolpickup_macro: G91&M62 P6&G4 P0 1.0&G0 Z-10.0&M63 P6&G4 P0 1.0&G0 Z10.0
  tool1_mpos_mm: -100.000 -7.500 -15.000
  tool2_mpos_mm: -120.000 -7.500 -15.000
  tool3_mpos_mm: -140.000 -7.500 -15.000
  tool4_mpos_mm: -160.000 -7.500 -15.000
  tool5_mpos_mm: -180.000 -7.500 -15.000
  tool6_mpos_mm: -200.000 -7.500 -15.000
  tool7_mpos_mm: -220.000 -7.500 -15.000
  tool8_mpos_mm: -240.000 -7.500 -15.000
```

but all I get when `M6 T2` is 
```txt
[MSG:INFO: Relay spindle changed to tool:2 using ATC:atc_basic]
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.
Core  1 register dump:
PC      : 0x40108f20  PS      : 0x00060e30  A0      : 0x801093d9  A1      : 0x3ffb1fd0
A2      : 0x3ffaf510  A3      : 0x00000000  A4      : 0x3ffae99c  A5      : 0x3ffc74c8
A6      : 0x0000000f  A7      : 0x00000038  A8      : 0x00000c24  A9      : 0x3ffb1fb0
A10     : 0x3ffae8ec  A11     : 0x00000000  A12     : 0x3ffe5280  A13     : 0x00000039
A14     : 0x3ffe52c6  A15     : 0x3ffe5280  SAR     : 0x00000008  EXCCAUSE: 0x0000001c
EXCVADDR: 0x00000004  LBEG    : 0x4008bf75  LEND    : 0x4008bf97  LCOUNT  : 0xffffffff
Backtrace: 0x40108f1d:0x3ffb1fd0 0x401093d6:0x3ffb2000 0x40106499:0x3ffb20b0 0x400e1e7f:0x3ffb20f0 0x400fce1d:0x3ffb21d0 0x400fde21:0x3ffb2210 0x400edff0:0x3ffb2250 0x40129929:0x3ffb2290
ELF file SHA256: f0f5be5fabded9f2
```

bdring:
If you use the vscode serial monitor, it will decode the backtrace.

FluidNC/src/ToolChangers/atc_basic.cpp
```diff
@@ -0,0 +1,151 @@
...
+            if (_prev_tool > 0) {
+                // return tool
+                move_to_tool_position(_prev_tool);
+                _macro.addf(_toolreturn_macro._gcode.c_str()); // use macro with G91 movements or the _tc_tool_* variables to to return tool, operating the ATC using M62 & M63
+                _macro.addf(set_state._gcode.c_str()); // ensure the previous user macro didn't change modes
+            }
+
+            if (new_tool > 0) {
+                //pickup tool, if this is the 1st pickup ever, we also probe the tool_setter_offset
+                move_to_tool_position(_prev_tool);
```
nicolas-lang: this needs to use `new_tool`

nicolas-lang:
> If you use the vscode serial monitor, it will decode the backtrace.

that helped a lot ! Thanks for the suggestion.

After a bit of debugging it already looks quite promising.
It looks like Tools are successfull being measured and changed.

*I still have to double check probing with dial gauge and debug some strange observations...
(it behaves a bit like my dual Z is loosing it's secondary offset when probing a tool, at least it looks like there is a slight jerk in Y2 )*

nicolas-lang:
Added a small description with macro and cfg examples [here](https://github.com/nicolas-lang/FluidNC-Dev/blob/Basic_ATC/FluidNC/src/ToolChangers/atc_basic.md)

PR #1324, opened 2024-09-19
VFD rework

atlaste:
Per request of @MitchBradley I did some rework on the VFD code. Basically I moved all the VFD specific code that can reside in a task to said task, removing the requirement for IRAM there. That should free up some RAM... This implementation should be backward compatible.

Needs more people testing this! I'm happy to test it out on my H2A VFD of course.


bdring:
It works on my Huanyang.

bdring:
@atlaste Did you test on the H2A VFD yet?

atlaste:
> @atlaste Did you test on the H2A VFD yet?

Not yet; I wanted to do some proper unit testing and crash testing first, so I ended up spending a few evenings writing test code for that. That test succeeded, which means that I'm fairly sure that the RS485 RX/TX traffic of the code is the same for the spindles that I've tested.
The other test that I wanted to do is force a crash on the system, so we don't get into the IRAM mess again. I know how Mitch did that, so I'm going to hack something together in the code to mimick that behavior as well. I just haven't had time for that yet.

atlaste:
The H2A also works. Just need the crash test after the IRAM changes land, and then I'm confident it's fine to pull.

atlaste:
Now that I know why my crash test crashed the CPU, I just moved the LD scripts back to DRAM to do the crash test. The VFD code itself wasn't the cause.

I haven't been able to crash the code ever since. I'm confident enough this works. Feel free to pull.

PR #1330, opened 2024-09-24
Implement M66 Immediate mode

dymk:
This is a partial implementation of [M66 - Wait On Input](https://linuxcnc.org/docs/2.8/html/gcode/m-code.html#mcode:m66), as defined by LinuxCNC.

Adds a few things that needed to be in place - 
- UserInputs class and configuration. This is similar to UserOutputs, and defines pins to use for reading digital and analog inputs. Do note that there's not support for reading actual analog values yet, this is a partial implementation. There would need to be further changes to support actual analog reads of pins, which as far as I can tell isn't yet done in the codebase.
- A new Error to indicate that one of a gcode command's value word is invalid
- Change the M62-M68 modal groups to be M5 to reflect LinuxCNC
- Enum to represent M66 wait mode

And finally, the M66 implementation itself. This is a partial implementation only, and supports only Immediate wait mode (0). It will read a user defined input and set numbered parameter `#5399` to that value. That parameter can be used in subsequent conditional branching.

This is to support implementing safe pneumatic ATC operation, which relies on reading sensor inputs to detect tool presence. That being said, I'm sure there are plenty of usecases for reading sensor values and inputs.

~Now working on testing this on real hardware.~

Tested on ATC with an NPN NC inductive sensor.

Relevant log lines after the controller is reset
```
...
[MSG:INFO: User Digital Output: 3 on Pin:gpio.12]
[MSG:INFO: User Digital Input: digital0_pin on Pin gpio.36:low]
[MSG:INFO: Axis count 3]
...
```

`config.yml` snippet configuring digital0_pin as an input - 
```yaml
# ...
user_inputs:
  digital0_pin: gpio.36:low
# ...
```

Using the M66 command -
```
; A tool is in the ATC holder
M66 P0 L0
[MSG:DBG: M66: digital_input0 result=1]
ok
(print,#5399)
[MSG:INFO: PRINT,1.000000]
ok
; Tool is removed from holder (should still be 1, no M66 issued yet)
(print,#5399)
[MSG:INFO: PRINT,1.000000]
ok
M66 P0 L0
[MSG:DBG: M66: digital_input0 result=0]
ok
(print,#5399)
[MSG:INFO: PRINT,0.000000]
ok
```


MitchBradley (commented):
I haven't tested it so I can't say if it works, but I don't see any problems with the code.

bdring:
Nice work. I can do some testing.

bdring:
I am getting some `[MSG:ERR: Gcode unused words]` messages. I think you need some `clear_bitnum(value_words, GCodeWord::L);` type statement to remove the words as they are used.

dymk:
Thanks for the fast testing @bdring - I just tested this on my CNC, and added fixes for those errors. It's working on my hardware now.

bdring:
It works better for me now.

I am not sure how this should work.

#5399 does not exist until you do an M66. The `[EXISTS[...]]` function does not work on numbered parameters.

```
Grbl 3.8 [FluidNC v3.8.2 (pr/1330-3eda31c0) (noradio) '$' for help]
[MSG:INFO: ALARM: Unhomed]
ALARM:14
$H
[MSG:Homed:Z]
[MSG:Homed:XY]
ok
D#5399
[MSG:INFO: param #5399 is not found]
error:2
[MSG:ERR: Bad GCode number format]
M66 P0 L0
ok
D#5399
[MSG:INFO: Value is 0.000]
ok
```

dymk:
The LinuxCNC documentation doesn't explicitly state what system defined numbered parameters should return in situations like this, but I think a reasonable default is to initialize `5399` to `0.0`.

dymk:
Changed so that `#5399` is initialized to 0 on controller boot. Added a gcode fixture test.

```
# => ./config.yaml /littlefs/config.yaml
# restart the machine to clear all variables
-> $Bye
<... * Grbl 3.8*
# parameter should be default initialized to 0
-> (print,#5399)
<- [MSG:INFO: PRINT,0.000000]
-> M66 P0 L0
<- ok
-> (print,#5399)
<- ok
<- [MSG:INFO: PRINT,0.000000]
```

PR #1400, opened 2024-12-11
Added $32 read only proxy

bdring:


bdring:
Here is some test data. 

```
M6T10
[MSG:INFO: Changed to spindle:Laser]
ok
$32
$32=1
ok
M6T1
[MSG:INFO: Changed to spindle:PWM]
ok
$32
$32=0
ok
```

PR #1425, opened 2025-01-10
Better m6 macros

bdring:
## Control of the current tool number

If you are using m6_macros, you must set the current tool after the tool change with M61Qx. This allows the macro to control what happens if there is an issue preventing a successful tool change.

If you are not using m6_macros, FluidNC will change the tool number.

## Using Multiple Spindles

If you are using multiple spindles and want to change to a new spindle, but do something to the current spindle first, send an M6Tx to the current spindle then send M6Tx to the new spindle.

For example: If you want to change from an ATC spindle (tool range 0-9) to a laser (tool range 10-255), but want to put away the current tool first. Send this via console or include in your gcode file.

```gcode
M6T0 ; put away current tool
M6T11 ; switch to laser spindle 
```

Your macros should use the selected_tool and current_tool to determine what to do. In the case of secondary spindles the tool range is not accessible to the macro. You should hard code variables for these numbers.

The tool numbers have no special significance in FluidNC other than determining the current spindle. Your macro determines if Tool 0 means no tool or if another tool has special meaning like a touch probe. 

## M2 and M30 

Fixed some issue with M2 and M30. This returns from the current job or macro. Be careful using in macro because they reset the modal values like G54, G21, etc

## % character support 

The us of % in gcode files is now compliant with LinuxCNC. We don't suggest its use though. There is no real benefit.

## Alarm/Send in macros

A race condition was fixed so alarms stop all gcode processing

### DIFF
--- a/FluidNC/src/Channel.h
+++ b/FluidNC/src/Channel.h
@@ -81,7 +81,8 @@ protected:
 
     UTF8 _utf8;
 
-    bool _ended = false;
+    bool _ended   = false;
+    bool _percent = false;
 
 protected:
     bool _active = true;
@@ -180,6 +181,7 @@ public:
     void push(const std::string& s) { push(reinterpret_cast<const uint8_t*>(s.c_str()), s.length()); }
 
     void end() { _ended = true; }
+    void percent() { _percent = true; }
 
     // Pin extender functions
     virtual void out(const char* s, const char* tag);
--- a/FluidNC/src/GCode.cpp
+++ b/FluidNC/src/GCode.cpp
@@ -1605,7 +1605,7 @@ Error gc_execute_line(char* line) {
     // NOTE: Pass zero spindle speed for all restricted laser motions.
     if (!disableLaser) {
         pl_data->spindle_speed = gc_state.spindle_speed;  // Record data for planner use.
-    }                                                     // else { pl_data->spindle_speed = 0.0; } // Initialized as zero already.
+    }  // else { pl_data->spindle_speed = 0.0; } // Initialized as zero already.
     // [5. Select tool ]: NOT SUPPORTED. Only tracks tool value.
     // [M6. Change tool ]:
     if (gc_block.modal.tool_change == ToolChange::Enable) {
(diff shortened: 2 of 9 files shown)


bdring:
Here is some test data. This is using a separate macro for each spindle. They both set M61 as required. The PWM spindle macro sends an alarm if the tool is 9.

```
Grbl 3.9 [FluidNC v3.9.4 (BetterM6Macros-84f6d403) (wifi) '$' for help]
(print, Selected tool:%d#<_selected_tool>, Current tool:%d#<_current_tool>)
[MSG:INFO: PRINT, Selected tool:0 Current tool:-1]
ok
M61Q1
ok
(print, Selected tool:%d#<_selected_tool>, Current tool:%d#<_current_tool>)
[MSG:INFO: PRINT, Selected tool:1 Current tool:1]
ok
T2
ok
(print, Selected tool:%d#<_selected_tool>, Current tool:%d#<_current_tool>)
[MSG:INFO: PRINT, Selected tool:2 Current tool:1]
ok
M6
[MSG:INFO: Sel:2 Cur:1]
ok
(print, Selected tool:%d#<_selected_tool>, Current tool:%d#<_current_tool>)
[MSG:INFO: PRINT, Selected tool:2 Current tool:2]
ok
M6T3
[MSG:INFO: Sel:3 Cur:2]
ok
(print, Selected tool:%d#<_selected_tool>, Current tool:%d#<_current_tool>)
[MSG:INFO: PRINT, Selected tool:3 Current tool:3]
ok
M6T0
[MSG:INFO: Sel:0 Cur:3]
ok
[MSG:INFO: PRINT, Deactivate PWM]
(print, Selected tool:%d#<_selected_tool>, Current tool:%d#<_current_tool>)
[MSG:INFO: PRINT, Selected tool:0 Current tool:0]
ok
M6T11
[MSG:INFO: Changed to spindle:Laser]
[MSG:INFO: Sel:11 Cur:0]
ok
(print, Selected tool:%d#<_selected_tool>, Current tool:%d#<_current_tool>)
[MSG:INFO: PRINT, Selected tool:11 Current tool:11]
ok
M6T10
[MSG:INFO: Sel:10 Cur:11]
ok
[MSG:INFO: PRINT, Deactivate laser]
M6T9
[MSG:INFO: Changed to spindle:PWM]
[MSG:INFO: Sel:9 Cur:10]
ok
[MSG:INFO: ALARM: Abort Cycle]
ALARM:3
```

PR #1431, opened 2025-01-21
Generic vfd improvements

MitchBradley:
Added predefined ModbusVFD types that can be selected just by setting the model: string.


MitchBradley:
Fixes #1454

PR #1457, opened 2025-02-24
State Machine Part 1

BarbourSmith:
This adds the basic framework for the calibration process to be run by a single unified state machine instead of a patchwork of functions and state variables which can be hard to understand and unpredictable.


BarbourSmith:
Sorry wrong target

PR #1458, opened 2025-03-06
BESC signal duty at turn on

bdring:
Initial signal duty should be `min_pulse_us` not 0

### DIFF
--- a/FluidNC/src/Spindles/BESCSpindle.cpp
+++ b/FluidNC/src/Spindles/BESCSpindle.cpp
@@ -58,6 +58,7 @@ namespace Spindles {
         stop();
         init_atc();
         config_message();
+        set_besc_duty(0);
     }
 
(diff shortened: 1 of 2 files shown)


FluidNC/src/Spindles/BESCSpindle.cpp
```diff
@@ -58,6 +58,7 @@ namespace Spindles {
         stop();
         init_atc();
         config_message();
+        set_besc_duty(0);
     }
```
MitchBradley, line 62: Why not just call set_output(0) here instead of introducing the new set_besc_duty() factor?

MitchBradley:
It appears to me that we could simplify this greatly by setting up the speed map correctly, so all the conversion is done by that existing mechanism.  I am going to take a stab at it.  I can't test with a real BESC, but I can look at waveforms.

PR #1491, opened 2025-05-04
Trinamic Diag0 output on 2130 and 5160 drivers

bdring:
This allows diag0 to be used for driver error reporting.

### DIFF
--- a/FluidNC/src/Motors/TMC5160Driver.cpp
+++ b/FluidNC/src/Motors/TMC5160Driver.cpp
@@ -54,6 +54,10 @@ namespace MotorDrivers {
 
         tmc5160->tpfd(_tpfd);
 
+        tmc5160->diag0_error(_diag0_error);
+        tmc5160->diag0_otpw(_diag0_otpw);
+        tmc5160->diag0_int_pushpull(_diag0_int_pushpull);
+
         switch (_mode) {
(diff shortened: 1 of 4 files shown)


FluidNC/src/Motors/TMC5160Driver.cpp
```diff
@@ -54,6 +54,10 @@ namespace MotorDrivers {
 
         tmc5160->tpfd(_tpfd);
 
+        tmc5160->diag0_error(_diag0_error);
+        tmc5160->diag0_otpw(_diag0_otpw);
+        tmc5160->diag0_int_pushpull(_diag0_int_pushpull);
```
MitchBradley, line 59: What does the "int" in diag0_int_pushpull signify?  Internal? Integer? Something else?

bdring: That is the name of the register on the chip. The name is also used in the library.
It apppears to mean internal.

MitchBradley (approved):

FluidNC/src/Motors/TMC5160Driver.cpp
MitchBradley: I suppose it could also mean "interrupt".

PR #1494, opened 2025-05-06
Bridge mode v2

Benedict-Carling:
Currently in draft

### DIFF
--- a/FluidNC/src/Bridge.cpp
+++ b/FluidNC/src/Bridge.cpp
@@ -9,8 +9,8 @@
  * - A UART connection to an external device (like a pendant)
  *
  * The bridge mode can be entered using the $BRIDGE command and exited using:
- * 1. $NORMAL or $$NORMAL command from USB (computer side)
- * 2. $$EXIT_BRIDGE command from the pendant
+ * 1. $NORMAL command from USB (computer side)
+ * 2. $EXIT_BRIDGE command from the pendant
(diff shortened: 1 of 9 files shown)


MitchBradley:
* Needs some justification/rationale text
* Can the existing passthrough mode be used for this purpose, or modified to serve both purposes?

Benedict-Carling:
> * Needs some justification/rationale text
> * Can the existing passthrough mode be used for this purpose, or modified to serve both purposes?

Hi @MitchBradley thanks for the quick response, it is a great project.

I'm looking into ways to update / flash a fluidNC pendant (esp32-s3) via the fluidNC itself (by having the fluidNC act as a USB to UART bridge) and thought a new state might be the easiest solution, thanks for the idea to look into using / expanding the pass though mode, I had not thought to use that.

I'll send you a review request once I am happy with the implementation e.g. deciding to use pass through mode (it may be that no code change is required in which case I'll close this PR) or continue via the initial approach, the new bridge state method

All the best,
Ben

MitchBradley:
The purpose of the passthrough mode is to allow flashing of an expander that is connected to FluidNC, via an external program that talks to the main FluidNC serial port.  FluidTerm has a built-in STM32 download program that uses the facility.  You can see how it works at https://github.com/MitchBradley/FluidTerm2

In principle it should work with a ESP32 pendant too, but you would probably need to put the pendant into download mode manually, and possibly force the esptool com port via the command line.

PR #1496, opened 2025-05-11
Probe crash

MitchBradley:
1. Fixed sporadic crash while probing that was caused by reentering report_recompute_pin_string().
2. Separated the reporting of probe (with P) and toolsetter (with T) pins.


bdring:
Tested both changes.

PR #1530, opened 2025-07-24
Increase default max feedrates for axis

BarbourSmith:
Changes the default maslow.yaml file to increase the max feedrate to 2200mm/min


BarbourSmith:
Sorry, wrong target 🤦‍♂️

PR #1535, opened 2025-08-02
Make sure that during long moves the belts stay synchronized

BarbourSmith:
The new kinematics system was correctly computing the lengths of the belts at the beginning and end of each movement, but it was using a linear interpolation between those points. This works on traditional CNC machines, but for maslow this was leading to the belts not staying tight during long moves. 

This PR breaks long moves into smaller segments retaining the correct lengths of each belt over the course of long moves.


BarbourSmith:
Sorry, wrong target :/

PR #1552, opened 2025-08-24
Add UVW Linear Axis

Skorpeo:
Add UVW Linear Axis as 7th 8th and 9th Axis

### DIFF
--- a/FluidNC/src/Config.h
+++ b/FluidNC/src/Config.h
@@ -23,7 +23,7 @@ Some features should not be changed. See notes below.
 // machine.h is #included below, after some definitions
 // that the machine file might choose to undefine.
 
-const int MAX_N_AXIS = 6;
+const int MAX_N_AXIS = 9;
 
 const int MAX_MESSAGE_LINE = 256;
 
@@ -36,13 +36,16 @@ const int Z_AXIS = 2;
 const int A_AXIS = 3;
 const int B_AXIS = 4;
 const int C_AXIS = 5;
+const int U_AXIS = 6;
+const int V_AXIS = 7;
+const int W_AXIS = 8;
 
-const int X2_AXIS = (X_AXIS + MAX_N_AXIS);
-const int Y2_AXIS = (Y_AXIS + MAX_N_AXIS);
-const int Z2_AXIS = (Z_AXIS + MAX_N_AXIS);
-const int A2_AXIS = (A_AXIS + MAX_N_AXIS);
-const int B2_AXIS = (B_AXIS + MAX_N_AXIS);
-const int C2_AXIS = (C_AXIS + MAX_N_AXIS);
+// const int X2_AXIS = (X_AXIS + MAX_N_AXIS + 1);
+// const int Y2_AXIS = (Y_AXIS + MAX_N_AXIS + 1);
+// const int Z2_AXIS = (Z_AXIS + MAX_N_AXIS + 1);
+// const int A2_AXIS = (A_AXIS + MAX_N_AXIS + 1);
+// const int B2_AXIS = (B_AXIS + MAX_N_AXIS + 1);
+// const int C2_AXIS = (C_AXIS + MAX_N_AXIS + 1);
 
 const int SUPPORT_TASK_CORE = 0;  // Reference: CONFIG_ARDUINO_RUNNING_CORE = 1
 
--- a/FluidNC/src/Parameters.cpp
+++ b/FluidNC/src/Parameters.cpp
@@ -112,7 +112,7 @@ bool ngc_param_is_rw(ngc_param_id_t id) {
 }
 
 static bool is_axis(int axis) {
-    return axis >= 0 && axis < MAX_N_AXIS;
+    return axis >= 0 && axis < MAX_N_AXIS - 1;
 }
 static bool is_rotary(int axis) {
     return axis >= A_AXIS && axis <= C_AXIS;
--- a/FluidNC/src/ProcessSettings.cpp
+++ b/FluidNC/src/ProcessSettings.cpp
@@ -479,13 +479,22 @@ static Error home_b(const char* value, AuthenticationLevel auth_level, Channel& 
 static Error home_c(const char* value, AuthenticationLevel auth_level, Channel& out) {
     return home(bitnum_to_mask(C_AXIS), out);
 }
+static Error home_u(const char* value, AuthenticationLevel auth_level, Channel& out) {
+    return home(bitnum_to_mask(U_AXIS), out);
+}
+static Error home_v(const char* value, AuthenticationLevel auth_level, Channel& out) {
+    return home(bitnum_to_mask(V_AXIS), out);
+}
+static Error home_w(const char* value, AuthenticationLevel auth_level, Channel& out) {
+    return home(bitnum_to_mask(W_AXIS), out);
+}
 static std::string limit_set(uint32_t mask) {
-    const char* motor0AxisName = "xyzabc";
+    const char* motor0AxisName = "xyzabcuvw";
     std::string s;
     for (int axis = 0; axis < MAX_N_AXIS; axis++) {
         s += bitnum_is_true(mask, Machine::Axes::motor_bit(axis, 0)) ? char(motor0AxisName[axis]) : ' ';
     }
-    const char* motor1AxisName = "XYZABC";
+    const char* motor1AxisName = "XYZABCUVW";
     for (int axis = 0; axis < MAX_N_AXIS; axis++) {
         s += bitnum_is_true(mask, Machine::Axes::motor_bit(axis, 1)) ? char(motor1AxisName[axis]) : ' ';
     }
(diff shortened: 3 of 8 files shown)


Skorpeo:
Hi, I just saw I should open a issue first.  Let me know if I can leave this here.  Also, I already did the work for additional rotary axis so wasn't sure if we can repurpose "D" command to bring in D E F rotary axis and use a different letter instead?

MitchBradley:
You can leave this PR open and refer to an issue that you create after the fact.

D, E, F are not available.  https://linuxcnc.org/docs/html/gcode/overview.html#_word

The only real difference between linear and rotary axes is that linear axes are scaled in G20 (inches) mode, while rotaries aren't.  If you only work in one G21 (mm) mode, there will not be a problem

Skorpeo:
[https://github.com/bdring/FluidNC/issues/1553](https://github.com/bdring/FluidNC/issues/1553)

MitchBradley:
Next steps, if the PR is accepted:
* Edit the wiki to acknowledge the new axes
* Check to see if the new axes break WebUI and WebUI-3 and possibly fix them
* Check to see if various senders (http://wiki.fluidnc.com/en/support/gcode_senders) either support or choke on the new axes and document the results in the wiki.

Skorpeo:
ok i already tested on tinybee and webui shows the new axis in settings and gcode works.   I only tested webui and UGS - UGS doesn't show the UVW in the GUI (same with webui) but manually entering gcode works.   I will look for the wiki docs.

bdring:
I added a W axis to my XYZ machine config. This created ABCUV to fill in the missing axes.

I can move all axes except for the W

```
Grbl 3.9 [FluidNC v3.9.8 (pr/1552-414e4d4c) (wifi) '$' for help]
$X
[MSG:INFO: Caution: Unlocked]
ok
G0 x10 y10 z10 a10 b10 c10 u10 v10 w10
ok
<Idle|MPos:10.000,10.000,10.000,10.000,10.000,10.000,10.000,10.000,0.000|FS:0,0|Pn:Z|WCO:0.000,0.000,0.00
```

If I add V as the last axis, V does move.

Have you tried all axes?

Skorpeo:
Hi, admittedly I hadn't tested W axis yet and I can reproduce your issue, W also doesn't work for me.

Skorpeo:
I tried various things but I can't figure out why its not working :/

MitchBradley:
Is there a uint8_t being used as an axis bitmask somewhere?

MitchBradley:
./Machine/Axes.h:77:        static void step(uint8_t step_mask, uint8_t dir_mask);
./Stepper.cpp:34:    uint8_t  direction_bits;
./Stepper.cpp:70:    uint8_t  step_bits;     // Stores out_bits output to complete the step pulse delay
./Stepper.cpp:71:    uint8_t  execute_step;  // Flags step execution for each interrupt.
./Stepper.cpp:72:    uint8_t  step_outbits;  // The next stepping-bits to be output
./Stepper.cpp:73:    uint8_t  dir_outbits;
./Stepping.cpp:123:void IRAM_ATTR Stepping::step(uint8_t step_mask, uint8_t dir_mask) {
./Stepping.cpp:126:    static uint8_t previous_dir_mask = 255;  // should never be this value
./Stepping.h:76:        static void step(uint8_t step_mask, uint8_t dir_mask);

Skorpeo:
I'm not going to pretend like I know what I'm doing but it seems to work now

FluidNC/src/Stepper.cpp
```diff
@@ -67,11 +67,11 @@ typedef struct {
...
-    uint8_t  step_bits;     // Stores out_bits output to complete the step pulse delay
-    uint8_t  execute_step;  // Flags step execution for each interrupt.
-    uint8_t  step_outbits;  // The next stepping-bits to be output
-    uint8_t  dir_outbits;
-    uint32_t steps[MAX_N_AXIS];
+    uint16_t  step_bits;     // Stores out_bits output to complete the step pulse delay
+    uint16_t  execute_step;  // Flags step execution for each interrupt.
+    uint16_t  step_outbits;  // The next stepping-bits to be output
+    uint16_t  dir_outbits;
+    uint16_t steps[MAX_N_AXIS];
```
MitchBradley, line 75: Changing steps[] from 32 to 16 is bad.

Skorpeo: sorry sloppy

bdring:
I did some more testing. I made a XYUW machine. I tested homing with various cycle configs. So far, all work fine.

Skorpeo:
i had to revert a error i made

bdring:
Changing MAX_N_AXIS causes issues with saved values (coordinate systems). It seems to work when upgrading, but caused a crash when reverting. I had to force a reinstall of a clear partition in order for old versions to work again. 

We probably need some sort of version control of the settings. Grbl does that with ...

```cpp
// Version of the EEPROM data. Will be used to migrate existing data from older versions of Grbl
// when firmware is upgraded. Always stored in byte 0 of eeprom
#define SETTINGS_VERSION 10  // NOTE: Check settings_reset() when moving to next version.
```

We would need to figure out how to deal with existing releases.

We might consider adding [other values](http://wiki.fluidnc.com/en/development/open_issues#add-more-non-volatile-things) as well. (Tool Num and TLO)

FluidNC/src/ProcessSettings.cpp
```diff
@@ -490,18 +490,16 @@ static Error home_w(const char* value, AuthenticationLevel auth_level, Channel&
     return home(bitnum_to_mask(W_AXIS), out);
 }
 static std::string limit_set(uint32_t mask) {
-    // const char* motor0AxisName = "xyzabcuvw";
     const char* motor0AxisName = AXES_LABELS_LOWER;
```
Skorpeo, line 493: this can also be deleted  const char* motor0AxisName = AXES_LABELS_LOWER;

FluidNC/src/Parameters.cpp
```diff
@@ -80,9 +80,9 @@ const std::map<const std::string, int> work_positions = {
     { "_a", 3 },
     { "_b", 4 },
     { "_c", 5 },
-    //    { "_u", 0},
-    //    { "_v", 0},
-    //    { "_w", 0},
+    { "_u", 0},
```
Skorpeo, line 83: s/b 6,7,8?

FluidNC/src/Settings.cpp
```diff
@@ -420,31 +420,54 @@ Error UserCommand::action(const char* value, AuthenticationLevel auth_level, Cha
...
             // We could issue a warning message if we were so inclined.
-            return true;
+            break;
         case ESP_ERR_NVS_INVALID_NAME:
         case ESP_ERR_NVS_INVALID_HANDLE:
         default:
             return false;
     }
+#if MAX_N_AXIS == 9
+    len = 3 * sizeof(float);
```
Skorpeo, line 441: MAX_N_AXIS * sizeof(float); ?

MitchBradley: It is for the 3 axes that are not already in the other NVS blob.  Getting them all at once causes a buffer overrun crash if you downgrade.

FluidNC/src/ProcessSettings.cpp
MitchBradley: Delete it

FluidNC/src/Parameters.cpp
MitchBradley: Please fix it and test.  I am busy with other stuff.

MitchBradley:
I have been doing a lot of type sanitization in the S3OldCompiler branch.  I got rid of most bare "int" types in favor of more explicit types, to avoid complaints from modern compilers.  Then I changed a lot of things that used uint8_t "just because" into finer-grained types like axis_t, motor_t, AxisMask, MotorMask, objnum_t and pinnum_t.   axis_t, which is now an enum, was particularly invasive, and resulted in finding quite a number of potentially-dubious usages.  After making those changes, I incorporated the UVW extensions into the S3OldCompiler branch.

FluidNC/src/Config.h
```diff
@@ -23,26 +23,26 @@ Some features should not be changed. See notes below.
...
 const int Z_AXIS = 2;
 const int A_AXIS = 3;
 const int B_AXIS = 4;
 const int C_AXIS = 5;
+const int U_AXIS = 6;
+const int V_AXIS = 7;
+const int W_AXIS = 8;
+
+const std::unordered_set<int> LINEAR_AXES = { X_AXIS, Y_AXIS, Z_AXIS, U_AXIS, V_AXIS, W_AXIS };
+const std::unordered_set<int> ROTARY_AXES = { A_AXIS, B_AXIS, C_AXIS };
```
MitchBradley, line 44: These unordered_set definitions in a file that is included by nearly ever .cpp file increases the size of the image by 60K.  We cannot afford that.

MitchBradley:
This feature got added in commit 6cdf85a

Skorpeo:
Hi, thanks for all your help and efforts! Apologies, I wasn't able to do more.  This was actually my first time writing anything in c++ and I haven't had access to equipment last few months.

PR #1570, opened 2025-09-20
Changed atc_manual to G53 for all moves

bdring:
This makes it safe from G91 mode. It will now alarm if not in G17 mode.

 - Tested with G90 (works)
 - Tested with G91 (works)
 - Tested with G18 (alarmed)


qinMOTU:
I'll try again once my probe tip arrives. Thank you.

MitchBradley (approved):

PR #1574, opened 2025-09-24
Async webui

MitchBradley:


FluidNC/include/Driver/sdspi.h
```diff
@@ -4,4 +4,4 @@ bool sd_init_slot(uint32_t freq_hz, int cs_pin, int cd_pin = -1, int wp_pin = -1
 void sd_unmount();
 void sd_deinit_slot();
 
-std::error_code sd_mount(int max_files = 1);
+std::error_code sd_mount(int max_files = 3);
```
MitchBradley, line 7: This is a big concern for memory usage.  Each open file uses a large buffer (or maybe it is two of them) for the data sector and maybe also the cluster map.  The larger the SDCard allocation unit, the larger the buffers.  I set max_files at 1 and carefully arranged things to work with only one file open at a time, because multiple files open tended to use up all the memory and cause crashes.  We are pushing ESP32 right up to its limits, which is one reason I have been so resistant to add big features.

FluidNC/src/WebUI/WebClient.cpp
```diff
@@ -2,64 +2,178 @@
...
-    WebClient webClient;
+    QueueHandle_t WebClients::_background_task_queue=nullptr;
+    TaskHandle_t  WebClients::_background_task_handle=nullptr;
+
+    void WebClients::background_task(void* pvParameters) {
+        std::string cmd;
+        while (true) {
+            WebClient *_webClient;
+            if(xQueueReceive(_background_task_queue, &_webClient, portMAX_DELAY)==pdTRUE){
+                _webClient->_xBufferLock.lock();
```
MitchBradley: There are a lot of cases where the formatting does not match our standards.  We use clang-format, along with the .clang-format file in the repo, to ensure formatting consistency.  We have our editors set to run clang-format on save.

FluidNC/src/WebUI/WebClient.cpp
```diff
@@ -2,64 +2,178 @@
...
+        }
+    }
 
-    WebClient::WebClient() : Channel("webclient") {}
+    WebClient::WebClient() : Channel("webclient") {
+        //Uart0.printf("WebClient::WebClient()\n");
+        if(WebClients::_background_task_queue==nullptr) // If we are the first instanciation ever, create the event queue
+            WebClients::_background_task_queue = xQueueCreate(64, sizeof(WebClient*)); 
+        if(WebClients::_background_task_handle == nullptr){ // Same here, create the unique background task
+            xTaskCreatePinnedToCore(WebClients::background_task,                        // task
```
MitchBradley: Tasks are also consumers of substantial chunks of memory, so we are trying not to add more of them.

FluidNC/src/WebUI/WebClient.cpp
Gerporgl: Ok, I will look into it. I only realized midway that I was not aligned on the standard and left that for later on...

FluidNC/include/Driver/sdspi.h
Gerporgl: The reason is because with async, you may have more than one request happening to download a document at the same time, and I had seen this causing a crash because of the 1 limit. This fixed it.
Since while in motion we don't allow access (or long access), I was hopping this will not be an issue most of the time... or at least in my case this became more stable because of that change.

Edit: also files are always closed right after they are served/used.

FluidNC/src/WebUI/WebClient.cpp
Gerporgl: I know, I did not like that idea originally (and still not the biggest fan). In fact at that point I had almost given up on async as I was facing challenge responding to the ESP400 command specifically....

I could not find another memory efficient way, you see, this task actually saved a lot of memory, otherwise the previous solution was to hold everything into memory, and commands such as [ESP400] return a huge amount of data (33KB for me). This is kind of one drawback of working with the async library, at least the way I understood it, it wouldn't send back the data to the client if you are inside the request, but perhaps there is another way.

The task is stopped most of the time, as it waits forever on a new command, and while in motion, most commands are not allowed. I would have taught that a paused task with a small function should not use much resource.

Also, now the webClient is a new object that is free at the end of each request who needs to run a synchronous command... instead of one global. Again, this is because technically having only one webclient (like I had at the beginning) could cause 2 (or more) requests to steal each others responses (or something like that). Now a webclient live only for the time of one request, and is freed after.
The task is global as I thought it would be more efficient. But earlier one there was a task per webclient and terminated at the end...

MitchBradley: The static memory usage of the task is for its stack.  Tasks often need at least 4K of stack, especially if they call into library functions that can nest deeply into network stacks and whatnot.

Gerporgl: Yes, you are right, and I had increased it as well... because I think the lower initial stack size value did not work, but this may be different now.
I'll take further look at other memory usage to see if we can decrease it without decreasing actual stability.

FluidNC/src/WebUI/WebClient.cpp
Gerporgl: I committed a first pass of clang-format ting. Also changed some public variable names.

MitchBradley:
There is something I do not quite understand; perhaps you can explain the problem.

As background, [ESP... commands, notably ESP400, are synchronous in that the response data is returned via the HTTP transaction.  It uses WebClient as the Channel for routing the command output to the client.  In principle, it seems to me that the async web server ought to be able to handle something like that in a streaming fashion, perhaps by using chunked encoding.

What am I missing?

Gerporgl:
Yes, I know, at first I was also thinking this should be simple to do, especially with async it shouldn't be a disadvantage.
But I cannot call "send chunked" or send data... it is a callback form the AsyncWebServer that fires as it is ready to send more chunks of data to the client.

If I could have called beginSynchronouseCommand, then called a kind of pullSynchronouseCommandChunk from within the chunked response callback passing the maxlen multiple times that would have worked.

When you call settings_execute_line(...), it does not return until all of it is completed (even if internally it pushes it "somewhere"), and if you are inside the request handler when calling it (which is the case), then the callbacks cannot be called, so you have to find a place to hold the entire buffer, which is >32KB. It makes sense, they are only async, not multi threaded necessarily.

Other large payloads serving documents (like index.html.gz) work just fine, because you can read from a stream in chunks as the async stack is ready to send when the send buffer is good.

So in fact, if there was a version of executing a command with a response (and ideally with the json serializer included) as a stream that does things in real time that would work. I considered this idea, but that part of the FluidNC code seemed too much to change for me, and wasn't sure it was realistic to do.

So the task in background to "fork" the settings_execute_line seemed like the easiest that I could test, at least to see if it fixed my memory issue.

I even tried another idea to write the settings_execute_line into a temporary file, then stream it back as a filestream... it worked too, but I did not really like writing data as a file on flash, of course. To have to store a temporary 33KB json on the local FS, and well the SD was tempting but that would make even less sense.

Again, perhaps I am missing an easier approach.

MitchBradley:
Yuck.

FluidNC/src/WebUI/WebUIServer.cpp
```diff
@@ -133,33 +128,26 @@ namespace WebUI {
...
-        //_websocketserverv3 = new AsyncWebServer(_port+2);
-        _socket_serverv3 = new AsyncWebSocket("/"); //new WebSocketsServer(_port + 2, "", "webui-v3");
+        // The only major difference with websockets for v2 webui vs v3 seems to be the currentID vs CURRENT_ID and activeID vs ACTIVE_ID
+        // In order to only have one websocket server (for simplicity and maintability reasons) we could:
+        // 1 - Send both messages types all the time
+        // 2 - Don't do anything and just send v3 payloads, since at this point it seems we don't rely on this pageId mechanism anymore with
+        // our async and cookie session implementation
+        // 3 - Remove all of these active and current IDs altogether since again, we may have no need for this anymore
+        // 4 - Potentially check for a difference in requests headers of v2 vs v3 to dynamically send the proper payload in the same handler
+        // For now, I've settled with #3
```
michmela44, line 128: I would be curious to see how removing the CurrentID and ActiveID completely affects the WebUI v3, especially in the case of multiple connections. (I haven't used v2 in a while, and don't remember how it was handled)

At least in v3, I thought I remembered these being part of a handshake that was required to fully establish a connection. It was also used to manage which open browser is the "active" connection, since having multiple websocket clients wasn't supported.

Without that broadcast, I suspect we'd be left with a stale web client that isn't actually receiving data without knowing it.

Unless we also have removed that restriction in FluidNC as well, but I believe we had more memory/crashing problems related to multiple WebUI connections than anything else earlier on.

Gerporgl: Currently in the new async approach, I changed it to rely on session cookies that the esp sends to the web ui when index.html is requested (even for 304s) and are automatically handled by the browser (so I did not have to change the webui index.html at all to do this).

I tested most of the time on webui3, but also used webui2,  using the same websocket server and everything on a single port 80 and this method works pretty well for both.

The only drawback is that within one browser "process", on the same computer, all tabs share the same session cookie of a given site. So if you open multiple tabs, only the last one would be sent the websocket messages, since 2 sessions of the same ID exists and we pick the last websocket connected (that is just how these cookie works). I find this was acceptable. If you open another browser on another device, or even a new browser tab in a private session, you would not have this problem as they will get different session IDs.

I concluded that to solve this problem properly the webui should generate the session ID itself after the page load (perhaps this is what the pageid of webui2 was?), and always pass this (as a cookie or header ideally).

As for stale websocket clients, they will always end up having either a graceful disconnection by the browser, or a timeout by the async tcp stack, so I have never been able to see any issue with that.

I  agree that the general guideline for FluidNC should be one user session only, and the multiple browser sessions is just for when you need to switch from one computer to another, since memory is limited and we can't handle 10,000 concurrent connections like a modern PC web server. We could limit the websockets count, but currently it is the default, which I can't recall what it is.

Also, if you run the code you will see the number of websockets instanciated in the INFO logs, their client ID and session ID, and it is easy to see that things gets closed and freed. And for testing purpose (or perhaps this could stay like that) the auto reporting is currently forced to 50ms for all websocket connections, and I did not see any problem other than if there is a very slow client on a mobile phone for example (over cell network on a vpn) it may get dropped because their websocket queue gets full, but that does not impact the stability of the ESP or make it hang in any way, free memory still seems to be acceptable and well above the warning threshold.

michmela44: > As for stale websocket clients, they will always end up having either a graceful disconnection by the browser, or a timeout by the async tcp stack, so I have never been able to see any issue with that.

There is code within the WebUI itself that relies on those messages that lets the user know that a connection was made from another browser instance, and that their instance is inactive until they click the Reconnect button, which will kick off the other user. 

> As for stale websocket clients, they will always end up having either a graceful disconnection by the browser, or a timeout by the async tcp stack, so I have never been able to see any issue with that.

It I remember correctly, if a disconnect is received without this broadcast happening, then the Webui v3 instance will try to recover the connection automatically.

You will either be left with 2 instances fighting each other for connection, causing unnecessary processing on the ESP with connection handling over and over, or left with a browser instance that looks like it's working, but maybe doesn't. I don't think either of those are very good experiences.

> So if you open multiple tabs, only the last one would be sent the websocket messages, since 2 sessions of the same ID exists and we pick the last websocket connected (that is just how these cookie works). I find this was acceptable.

I currently find that users unknowingly will have multiple browser tabs open on their machine. This has caused issues in the past for lots of users we are aware of. v3 made this better by enforcing only one active tab at a time using these IDs. This sounds like we are asking for trouble, and maybe a support nightmare, if we allow many tabs to be open with inconsistent states, when users may unknowingly pick an old one over the newer one when navigating back and forth between tabs.

> (or perhaps this could stay like that) the auto reporting is currently forced to 50ms for all websocket connections,

I'd have to defer to @MitchBradley on that one, but 50ms feels unnecessarily fast for a default.

Gerporgl: > You will either be left with 2 instances fighting each other for connection, causing unnecessary processing on the ESP with connection handling over and over, or left with a browser instance that looks like it's working, but maybe doesn't. I don't think either of those are very good experiences.

I did not really see this fighting for each other, especially on the ESP side I did not see it getting reconnections or unwanted requests, even when looking at the browser network dev tools. It was more that one side did not get websockets messages anymore.

> I currently find that users unknowingly will have multiple browser tabs open on their machine. This has caused issues in the past for lots of users we are aware of. v3 made this better by enforcing only one active tab at a time using these IDs. This sounds like we are asking for trouble, and maybe a support nightmare, if we allow many tabs to be open with inconsistent states, when users may unknowingly pick an old one over the newer one when navigating back and forth between tabs.

Let me take a look at this, I can probably disconnect the first browser session of the same session ID, I just have to find what exactly webui3 need to show that disconnected page.

Again, this should not be too hard to fix also if we wanted to support this properly, by chaning the webui code, but I prefer to not go there myself.

Gerporgl:
The last commits should now properly kick/disconnect a previous browser session on the same browser on a different tab, this should avoid any confusion if one tab did not receive message updates in the previous approach where multiple clients would be allowed to stay connected. 

Reused the current id & active id logic that was there before, but limited to cases where it could cause confusion.

Tested briefly with webui3 and webui2, seems to work.

FluidNC/src/WebUI/WebUIServer.cpp
Gerporgl: > I'd have to defer to @MitchBradley on that one, but 50ms feels unnecessarily fast for a default.

I agree with that, it was mostly for stress testing.

It does look great though to see a fast refresh rate on the webui while a job is running, or also while jogging. The previous non async implementation was very fragile to using auto refresh since it could cause the job to slow down and hang, which was the whole reason for this async implementation.

The main downside I have seen with a very fast auto report is not that it would crash or hang, but it is the fact that the current websocket queue is set to 32, and while you are printing a job, it may be possible to get messages dropped if the wifi connection (or just the connection) happen to be slow (either on the ESP32 wifi or the client side is slow), and the client may get disconnected.

Perhaps having it on by default may not be a bad idea (100ms perhaps), or a config. Webui3 does not have a client side config for this I think, or I could not find it... I know webui2 does.

Also, even though I decreased the websocket queue from 64 to 32, there was no practical good reason for that, other that I was trying to get the lowest memory usage possible, but honestly, in practice, and under stress, I don't think there was any issue even at 64, and even higher.
I think 64 may be a more robust setting based on my previous testings. Anyway, this is a compile time switch, easy to change (WS_MAX_QUEUED_MESSAGE).

MitchBradley:
50 ms is probably too fast.  It is hard to read numbers that change faster than, say, 10 times per second.  5/second is probably as fast as you could reasonably use.

FluidNC/src/WebUI/WebUIServer.cpp
michmela44: I tend to run mine with auto report somewhere in the 250-500ms range.

The FluidNC wiki says

> When the machine is not moving, reports are sent when some reported value changes.  When moving, reports are sent at the time interval that the sender requested.

So the Reporting interval only seems to affect reporting while the machine is in motion, in which case, half second updates seem more than enough for the WebUI.

> Webui3 does not have a client side config for this I think, or I could not find it...

You configure the polling commands yourself in the settings.

If you want to have Automatic Reporting enabled by default, add a polling command for _**$Report/Interval=N**_ and set the polling time to 0. When the time is set to 0, WebUI v3 will only run the command once when the WebSocket connection is established.

MitchBradley:
For me, the most valuable thing about auto-reporting is the immediate reporting of state changes, e.g. from Run to Idle, and the reports of GCode mode changes.

Gerporgl:
> 50 ms is probably too fast. It is hard to read numbers that change faster than, say, 10 times per second. 5/second is probably as fast as you could reasonably use.

I'll change it to 200

FluidNC/src/WebUI/WebUIServer.cpp
Gerporgl: > If you want to have Automatic Reporting enabled by default, add a polling command for $Report/Interval=N and set the polling time to 0. When the time is set to 0, WebUI v3 will only run the command once when the WebSocket connection is established.

I see! I never really went far enough to understand that panel, very good to know. For me as soon as I read "polling" I was no this is not I want, I want updates to be pushed automatically. But indeed doing $ri= is what I was doing initially but didn't know I can do it automatically. So we can leave it slow by default, and if users want they can request a faster rate with that method, no need for extra config

I think having it on by default is nice as a default user experience so you get all the extra machine status and active modes as Mitch mentioned.

michmela44: The FluidNC wiki says 

> Automatic reporting is off by default, for compatibility with senders that do not expect it.

This is for Mitch and/or Bart to decide if they want to change that stance

Gerporgl: > Automatic reporting is off by default, for compatibility with senders that do not expect it.

I only turned it on by default for websockets when they connect, so it doesn't affect the senders.

michmela44: @MitchBradley I assume this comment was meant for here instead of the other PR

> On by default at the FluidNC level would potentially break one or more of the approximately one million senders that people use.
> 
> It would, though, be possible to have WebUI 2 and/or 3 default to auto-reporting.

I do agree that I've always found it odd that the WebUIs (I think both 2 and 3) require some manual intervention to get reporting working initially.

But I also agree that it's a dangerous change to enable this on the firmware side by default. 

It should be a pretty easy change in the WebUI v3 to enable some better defaults to help new users out. I'll have a look at making that change. Perhaps even just defaulting to auto-reporting even for existing users that have no polling commands would probably be ok. The WebUI is pretty useless if you don't get any feedback at all.

FluidNC/src/WebUI/WebClient.cpp
```diff
@@ -1,52 +1,156 @@
...
+            xBufferLock.lock();
+        }
+        xBufferLock.unlock();
+    }
+
+    int WebClient::copyBufferSafe(uint8_t* dest_buffer, size_t maxLen, size_t total) {
+        xBufferLock.lock();
+        if (_buflen > 0) {
+            int bytes = min(_buflen, maxLen);
+            memcpy(dest_buffer, _buffer, bytes);
```
MitchBradley, line 97: Instead of explicit C-style buffer management, perhaps it would be better to use std::string and let the compiler/library handle the allocation details.  What do you think?

Gerporgl: Originally I was just trying to reuse the existing buffer that was in there in that class before, which was also using C style memcpy, but I guess I changed a lot the way this buffer was used (and the entire class itself).

I also had the same thought afterward that it would probably have been easier to manage this with a strings, after I got everything sorted out and stable with the buffer approach. Also I wasn't sure about any potential performance impact of std::string and any memory overhead, so I left it like that.

We do need a size limit for that buffer to not use too much memory (worst case I've seen so far is [esp400] returning over 30KB of json data), but that can be done also with std::string

MitchBradley:
@Gerporgl I have been testing the async stuff in my S3Async branch.  It mostly works well but I have experienced the following problem a few times.  If I restart the FluidNC controller, a WebUI 2 instance often has trouble either establishing - or perhaps keeping open - the websocket.  Sometimes I have to reload the WebUI page several times before it will work right.

I have only limited debugging information about this, but in one instance I was able to look via Developer Tools.  I saw this request "http://fluidnc.local/command?commandText=G20&PAGEID=" fail with error 500.  It happened multiple times, then started working, perhaps after I established a connection on another device.

MitchBradley:
Diving into the code a bit, it seems like 500 would happen if getWSChannel(session) fails.

... and in turn, that would mean that _webWsChannels is empty.

Gerporgl:
I'll do more test with webui2 and try to reproduce this issue.
I have been testing mostly with webui3, and only did brief tests with webui2.

MitchBradley:
I am investigating pings, or the lack thereof, to webui 2.  It seems to have something to do with the distinction between _wsChannels and _webWsChannels

I don't see any place that ever adds an entry to _webWsChannels.

MitchBradley:
I think that the session id stuff needs to be made to interoperate with WebUI2's PAGEID/CURRENT_ID/ACTIVE_ID scheme.  Or else WebUI2 needs to be made aware of the new scheme.

Gerporgl:
> _webWsChannels

That is interesting that nothing is being added to this list... I am reviewing that code. There is probably some cleanup I did but I missed some dependent parts. I see also some older functions that are no longer used and will see if I can clean that up so that the code is less misleading. I did miss this sendPing, and thought it wasn't really used anymore, but it is, but currently probably does not send any.

I am still trying to reproduce this issue, I am now testing with webui2 but did not have any luck so far but I'll leave it running longer and test more.

If you have the web browser debugger open when it happens, it may help to check the cookie request header and session value passed to the web socket connection page to be sure something is passed. Also having the webui in DBG level can help to see the Uart0 logs for the web socket connections and their relevant ID and session string, ex.:
[MSG:DBG: WebSocket connect cid#1 from 192.168.96.11 uri / session VGGRYHKX total 1]
Example browser request header: "Cookie sessionId=VGGRYHKX"

I see also that you are using the mdns name fluidnc.local, I don't know if you have more than 1 device running locally, otherwise I am not sure what IP it will resolve to or if it could even change between requests. I use IP addresses on my side since I occasionally have multiple devices running.

Gerporgl:
I just saw a case of webui2 sending a request like command?commandText=$G&PAGEID= which returned a 500 error (websocket dead), but that is because that request came before the websocket connection was established on page load, and those type of commandText are websocket type of commands, so there is nothing really that can be done, other than perhaps changing the logic to revert to a synchronous command if that happens... but I'm not sure if the UI would expect that. It did not seem to cause any problem though, the ui cotinued to load fine. It seems a bit random, it does not always happen.

MitchBradley:
The problem I have seen seems to happen after the controller reboots, not on a long running session.  When it happens, WebUI appears to load successfully, but commands that you run from buttons and whatnot just don't get executed.

Gerporgl:
> The problem I have seen seems to happen after the controller reboots, not on a long running session. When it happens, WebUI appears to load successfully, but commands that you run from buttons and whatnot just don't get executed.

Yes, I am seeing the same thing as you were about to post this.

Webui2 does not seem to detect websocket disconnections as well as webui3 did in all my tests, webui3 also has a frequent ping (every 5 seconds) on the client side (which is probably why it did not really need the sendPing at all from the server) and it is working pretty flawlessly for reconnecting to fluidnc. It is very quick to detect websocket disconnections and reconnect after a reboot.
Eventually, I just saw that webui2 will reconnect, but it takes a very long time.

I'll revisit this to see of I can find a solution and look at the old code to see if there was a reliable mechanism to fix that.

I also just tested the logic to return an synchronous response when there is no websocket connection, and it does seem to fix (or mask) the issue, and does not return a 500 dead web socket, and send back the command response directly in the http request, and webui2 appear to take those responses and show them if needed, but this may break other things in the webui, I am not sure.

When you reboot the device, if you refresh the page in the browser, do you see issues after that? On my side it seemed ok. I know this is not a solution, but at least this will narrow the issue to only that case.

Gerporgl:
There are other XML format errors in the web browser console that I never looked at before, I'll have to look at that to see if that is also a problem.

Gerporgl:
I tested with the main branch of FluidNC (from bdring repo) and I  see the same 500 errors from command requests when websockets are not connected, also if I reboot the device it also does not reconnect automatically unless I wait for a very long time (10 minutes in my case), or refresh the whole page. During that time, if I send commands I don't see any of the responses.

I see that the XML errors are not there as much in the original version, so I'll dig on that issue next.

MitchBradley:
I am looking into some WebUI2 improvements that might help.  Fixing sendPing helps somewhat, but more testing is needed.

Gerporgl:
The XML error was easy, that is because (I think) the previous webserver library would always set Content-Type to text/html by default, and now by default there is no content-type at all if you don't set one. So I've just set it to text/html and the webui seems to be happy with that, although these commands are a mix of text and json sometimes, ideally we should set the proper content-type but I doubt that was the case even before, by looking at the code.

I am just surprised that it wasn't causing more issues than that, it almost looks like webui was still able to use the data of those commands, but perhaps not.

MitchBradley:
Push the fix to your branch and I will apply the patch to my WIP

Gerporgl:
Ok, I pushed my changes.

Gerporgl:
At the end, there isn't a lot of differences, the main thing I think was the content-type for synchronous commands, and the ping that wasn't working at all before, which I decided, at least for now after some testings, to ping with the same content to all clients, it could be changed easily.
I did not see on my side a measurable improvements to add the ping however, perhaps it was causing webui2 to consider it was disconnected after 10 minutes causing a reconnection, but I did not have the time to test and confirm this.

Previously the ping was passing the pageId (ex.: PING:1), which now would be the client ID (I think...), it is just that I am not too sure what to pass as an ID and what the webui would do with that anyway... since it seems to work without it, I would rather just send it empty. I'll see if I can find the code of webui2 and take a look at how it uses the ping...

I also got rid of that old ws list that wasn't really used anymore, after attempting to fix it, I thought it could cause more issues that benefits. So if everything looks good I would just remove that code that is now commented.

I also removed a function override that wasn't being used anymore.

MitchBradley:
I will take it from here.

MitchBradley:
It appears that async doesn't failover to AP mode if STA connect fails.

Gerporgl:
I haven't had a chance to test any of the AP connectivity, so that is possible it behaves differently.

Do you think it could be that line that was supposed to be only a temporary test I did:
https://github.com/Gerporgl/FluidNC/blob/async_webui/FluidNC/src/WebUI/WifiConfig.cpp#L792
It is supposed to be `return ConnectSTA2AP();` normally, it was for an earlier testing where I did not want to wait on wifi on startup (you may have noticed that different behavior compared to before on startup).

So on startup, we normally wait for wifi connection, and I guess if it does not work it uses AP, then perhaps the async webserver is only created after and binds to this instead, but it may not be able to do so if the server was already created.... anyway, it is probably worth a try to revert this, it wasn't supposed to be there.

MitchBradley:
I'm getting really confused by session ids.  There is one place that creates a random string, then there is getSession() that uses IP and port, but does not seem to be called.  And there is the ESPSESSIONID thing that seems to be related to authentication.

Can you explain what is going on?

Gerporgl:
The getSession that works with ip:port was a left over from an earlier test, I am removing it, I did not notice this was still there. It's not used anywhere.

As for ESPSESSIONID, this seems related to the authentication that is disabled by default (which I haven't done or touched), and I don't think it would work if we tried to enable it. I do not really know that part and how it works and or if it even work. I think the authentication can be done by a reverse proxy web server to save memory in the ESP.

And for the session cookies (random string), this is the approach I am using now to tie each websockets with web requests. It works by having the FluidNC webuiserver send a random unique cookie header (Set-Cookie) when the index.html is requested, but only if no cookie was already present from the client request header. Then automatically, this cookie is saved by the browser (for this "site" on this computer and browser) and always passed on any subsequent http or websocket request to tie the http web requests with the corresponding session websocket. It also does not require any special implementation in the webui itself since the cookie mechanism is a standard web browser mechanism.

MitchBradley:
I have a version of WebUI2 that uses your session string as the page_id value.  It seems to work okay.

I wonder if it would be better to change the session ID every time that index.html.gz is requested?  That would allow restarts to be tracked, and allow multiple tabs in the same browser.  Seems like that would result in more precise identification of websockets.

Thoughts?

Gerporgl:
Originally I had the session token generated on every page load, but since the cookies are shared with all tabs, this was causing only the last page to load to receive websocket communication. So I changed to stay the same, and find duplicated sessions and ask them to disconnect by sending nonmatching activeid vs page id, it worked pretty well, but was strange that you could only have one tab open on a given computer, although this also saved some resources by disconnecting old sessions.

But I guess my ideal approach has always been that the webui could generate the sessionid on index page load and keep it in memory and pass it always to each call (websocket calls).

So there is definitely room for improvement, I was trying to reduce my original scope by not changing the webui, so that I can focus on the async implementation part.

In your new approach, do you still use cookies at all?

MitchBradley:
My latest code, which includes small changes to WebUI2, seems to work well.  Here's how it works.  Most of this explanation should be familiar to you since it is basically your scheme.  I am writing it out to clarify it in my own mind.

* When a root request for index.html.gz arrives, a new random session ID is generated as a response cookie, regardless of whether the page was cached.  This is similar to your formula, except that it doesn't depend on whether there was already a cookie; every page reload creates a new session. Each WebUI instance is a separate session, even multiple tabs in the same browser.  
* (Same as yours) When a websocket is created, the middleware extracts the session ID cookie and uses it to tag the websocket.
* (Same) When a later HTTP request that needs to send an asynchronous response, the session ID cookie is used to lookup the correct websocket for that response.
* Every 5 seconds (used to be 10), a PING is sent on all the websockets.  If WebUI doesn't receive a ping within 10 seconds, it disconnects and displays an alert.

MitchBradley:
@Gerporgl I figured out how to return large JSON responses without collecting everything into a big string!  The key is to use AsyncResponseStream.  I added a variant to JSONencoder that invokes a callback function on flush(), then passed a lambda `[response](const char* s) { response->print(s); }`.  Works like a charm.

Gerporgl:
> @Gerporgl I figured out how to return large JSON responses without collecting everything into a big string! The key is to use AsyncResponseStream. I added a variant to JSONencoder that invokes a callback function on flush(), then passed a lambda `[response](const char* s) { response->print(s); }`. Works like a charm.

This is good news, I'm glad you figured this out!

I originally wanted to do something like that, but wasn't sure how the json encoder was working and where to hook this up in the code.

MitchBradley:
Instead of closing down one WebUI instance when another is opened on the same domain, instances are now independent so multiple ones Just Work.  It required adding PAGEID to WebUI3.

Gerporgl:
Yes, webui3 and pageid was my issue when I did the implementation. There were comments in the code that pageid wasn't used by webui3, and was relying on the last websocket connected to send back to the last opened page only, which wasn't a good user experience in my mind. Also, the pageid was confusing to me just because of the name itself (page), I thought it was related to keeping track of the different pages/tabs in the webui front-end (like when you go from the main dashboard to the settings page, etc...). It is only later on that I realized that pageid was probably some kind of what I call session id.

MitchBradley:
Superseded by #1601

PR #1575, opened 2025-09-24
Fix typos

luzpaz:
Found via `codespell -q 3 -L ascript,bu,datas,mot,nd,ot,ser`

Note: there are 2 source typos that were omitted from this PR in order not to block its merge. 
./FluidNC/src/Limits.h:21: cordinate ==> coordinate
./FluidNC/src/Pins/PinDetail.h:39: frequencey ==> frequency

### DIFF
--- a/CodingStyle.md
+++ b/CodingStyle.md
@@ -32,7 +32,7 @@ clang-format:
    problems, you should be fixing your includes, instead of disabling
    clang-format.
 2. Preprocessor commands are not formatted nicely in some cases.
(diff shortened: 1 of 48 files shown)


MitchBradley:
Normally I would be happy to take a patch like this, but I am in the middle of an invasive change to support ESP32-S3 and I don't want to have two patches that both affect a lot of files.  You could try rebasing against the S3OldCompiler branch

luzpaz:
Superseded by #1576

PR #1586, opened 2025-10-02
Fix invalid_arc routine to check for full rotations in G2/G3 arc commands

Popeyef5:
Simple fixes to account for the fact that if G2/G3 commands include Pn with n > 1, arcs perform at least one full rotation in the working plane before arriving at the final position. 

Before, the invalid_arc routine did not contemplate this but rather calculated the min/max positions along the commanded path by considering a partial arc between the current and desired position. 

With these changes, the routine correctly contemplates the case where n > 1 and calculates the min/max coordinates in the working plane along the commanded path correctly.


bdring:
I recalled seeing some tough arc gcode on this gcode viewer. It seems to work.

https://nraynaud.github.io/webgcode/

It has complex arcs with z motion and plane changes.

Do you have examples of arcs that failed before and work now that I can add to this test code?

image

Popeyef5:
Yes, the one in the original issue. I tested it late at night so please repeat the experiment yourself but 

Home axis
G1 X10 F1000
G3 X0 Y10 I-10 P2 F500

Went through before (https://youtube.com/shorts/Sco4Mq2YMPY?si=h7ExzpqSOneOAKIz) and correctly reports an error now.

PR #1587, opened 2025-10-02
Added a probe hard limit option

bdring:
This optionally gives an ExecAlarm::ProbeHardLimit alarm if the probe is tripped during homing, jogging or any motion cycle other than probing.


bdring:
This has been added to the main branch. It will be in the next release.

https://github.com/bdring/FluidNC/pull/1587

PR #1594, opened 2025-10-13
TangentialKnife from jejmule work now (tested) with inifinite C axe

xerio123:
now Tangential Knife works well I tested it on my CNC
Plus it can handle infinite C axes. That means it can do more than 0 to 360 deg. It choice the smallest angle between an angle and and his opposite. no matter if more than 360 degree


Ams-Laser:
how does this works ?

xerio123:
Very simple, it follows the path and keeps the angular A axe Tangential i can make a video if needed

PR #1604, opened 2025-11-03
S3 fixes

atlaste:
I made an update of the S3 async branch, fixing the bugs I could find. 

I've tried to make individual commits easy to follow.

These are just the compiler and linker bugs from IDF, with a relatively high warning level. Some of them are compiler errors (initialization order, etc), some of them are real bugs (like the string compares in the async web server). My next step is to upload and test these changes; this is still a work in progress.

### DIFF
--- a/FluidNC/src/Planner.cpp
+++ b/FluidNC/src/Planner.cpp
@@ -38,7 +38,7 @@ typedef struct {
 static planner_t pl;
 
 // Returns the index of the next block in the ring buffer. Also called by stepper segment buffer.
-uint8_t plan_next_block_index(uint8_t block_index) {
+static uint8_t plan_next_block_index(uint8_t block_index) {
     block_index++;
     if (block_index == config->_planner_blocks) {
         block_index = 0;
--- a/FluidNC/src/Spindles/VFDSpindle.cpp
+++ b/FluidNC/src/Spindles/VFDSpindle.cpp
@@ -147,7 +147,6 @@ namespace Spindles {
         int unchanged = 0;
         //            const int limit     = 150;  // 15 sec / 100 ms
         const int limit = 100;
-        // auto      last  = _sync_dev_speed; -> unused.
 
         if (_debug > 1 && _sync_dev_speed != UINT32_MAX) {
             log_info("Syncing to " << int(dev_speed));
--- a/FluidNC/esp32/esp32s3/USBCDCChannel_IDF.cpp
+++ b/FluidNC/esp32/esp32s3/USBCDCChannel_IDF.cpp
@@ -87,13 +87,6 @@ void USBCDCChannel::handle_line_state(bool dtr, bool rts) {
     _state |= ((!!rts) << 1) + (!!dtr);
     _state &= 0xfff;
 
-    #if 0
-    ::putchar(((_state >> 8) & 0xf) + '0');
-    ::putchar(((_state >> 4) & 0xf) + '0');
-    ::putchar((_state & 0xf) + '0');
-    ::putchar('\n');
-    #endif
-
     // A sequence of transitions from R1D1 to R0D0 to R1D0 triggers a persist restart
     if (_state == 0x302) {
         // Persistent restart - maintains USB connection
--- a/FluidNC/src/Planner.h
+++ b/FluidNC/src/Planner.h
@@ -93,9 +93,6 @@ plan_block_t* plan_get_system_motion_block();
 // Gets the current block. Returns NULL if buffer empty
 plan_block_t* plan_get_current_block();
 
-// Increment block index with wrap-around
-uint8_t plan_next_block_index(uint8_t block_index);
-
 // Called by step segment buffer when computing executing block velocity profile.
 float plan_get_exec_block_exit_speed_sqr();
 


FluidNC/esp32/esp32s3/USBCDCChannel.cpp
```diff
@@ -3,6 +3,8 @@
 
 #include "USBCDCChannel.h"
 
+#ifdef CONFIG_ESP_CONSOLE_USB_CDC
+
```
MitchBradley, line 7: Does CONFIG_ESP_CONSOLE_USB_CDC  enable CDC early in the bootloader, using the "hardware" USB stack?  That is incompatible with the new NVS config item that lets you control whether the tinyusb-based USB stack is enabled.  When disabled, the USB pins FPIO 19 and 20 can be used for other purposes.

I was really trying to avoid ifdefs for this, and to avoid doubling the number of released images for USB support.

FluidNC/esp32/esp32s3/USBCDCChannel.cpp
```diff
@@ -202,32 +213,34 @@ void USBCDCChannel::flushRx() {
...
+    while (timeout) {
+        if (_cdc.available()) {
+            int res = int(_cdc.read(buffer, length));
+            // If res < 0, no bytes were read
+            return res < 0 ? 0 : res;
         }
-    }
 
-    return length - remlen;
+        delay_ms(1);
```
MitchBradley: Have you tested this with CTRL-U upload in FluidTerm and WebInstaller.  I fiddled with this code a lot to get that to work right.  Is there a situation where the replaced code fails?

FluidNC/include/Driver/fluidnc_uart.h
```diff
@@ -8,7 +8,7 @@ class InputPin;
 void uart_init(uint32_t uart_num);
 void uart_mode(uint32_t uart_num, uint32_t baud, UartData dataBits, UartParity parity, UartStop stopBits);
 bool uart_half_duplex(uint32_t uart_num);
-int  uart_read(uint32_t uart_num, uint8_t* buf, size_t len, uint32_t timeout_ms);
+int  uart_read(uint32_t uart_num, uint8_t* buf, uint32_t len, uint32_t timeout_ms);
```
MitchBradley, line 11: Why uint32_t?  size_t is typically used for len arguments to POSIX read() functions.

FluidNC/src/Spindles/VFDSpindle.cpp
```diff
@@ -147,7 +147,7 @@ namespace Spindles {
         int unchanged = 0;
         //            const int limit     = 150;  // 15 sec / 100 ms
         const int limit = 100;
-        auto      last  = _sync_dev_speed;
+        // auto      last  = _sync_dev_speed; -> unused.
```
MitchBradley: Just delete it

FluidNC/src/WebUI/WifiConfig.cpp
```diff
@@ -442,6 +499,17 @@ namespace WebUI {
                     log_stream(out, "Gateway: " << IP_string(IPAddress(WiFi.softAPIP())));
                     log_stream(out, "Mask: 255.255.255.0");
 
+#ifdef IDFBUILD
```
MitchBradley, line 505: Could this be based on the IDF version instead of IDFBUILD?

FluidNC/src/OLED.cpp
```diff
@@ -323,6 +323,7 @@ void OLED::parse_status_report() {
         }
         if (tag == "A") {
             // SCFM
+            /* Unused.
```
MitchBradley, line 326: I don't understand the reason for commenting this out.

FluidNC/src/Planner.cpp
```diff
@@ -38,7 +38,7 @@ typedef struct {
 static planner_t pl;
 
 // Returns the index of the next block in the ring buffer. Also called by stepper segment buffer.
-static uint8_t plan_next_block_index(uint8_t block_index) {
+uint8_t plan_next_block_index(uint8_t block_index) {
```
MitchBradley: I don't see any external uses of plan_next_block_index().  Wouldn't it be better just to delete the declaration in Planner.h?

FluidNC/esp32/esp32s3/USBCDCChannel.cpp
atlaste: It basically disables the bootloader CDC/Console support used for various things. The CONFIG_ESP_CONSOLE_USB_CDC is just an sdkconfig setting. 

"I was really trying to avoid ifdefs for this, and to avoid doubling the number of released images for USB support." -- I don't think you need to double releases (in any event); I totally understand why that's undesirable. 

I think I can just get rid of the ifdef and get it to compile. This file gave quite a few errors in the beginning so I figured this was the root cause and added the check. That might have been too premature on my end. I'll get to the bottom later; I have some more pressing issues first.

FluidNC/src/WebUI/WifiConfig.cpp
atlaste: No, i tried that. Arduino also compiles using idf at some point, so it defines the same symbol during compilation. I can check the version as well, but must add a -D to cmakelists to be able to differentiate between the frameworks.

MitchBradley: I still don't see why this wouldn't work:
```c++
#include "esp_idf_version.h"
#if  ESP_IDF_VERSION_MAJOR >= 5
...
```
the point being that this change, and a few others, are needed because you are building with IDF 5.5 vs 4.7, not because you are building with IDF vs Arduino.  Switching on version will help later when Arduino gets to IDF 5.

atlaste: Thats fair. Ill walk through the ifdefs to see whats what, its not much anyways.

FluidNC/src/Protocol.cpp
```diff
@@ -122,8 +123,11 @@ char activeLine[Channel::maxLine];
 
 bool pollingPaused = false;
 void polling_loop(void* unused) {
+    // Add current task to the watchdog
```
MitchBradley: Ouch, just realized that this creates an ESP32 dependency in code that should be CPU agnostic.  Hmmm.

FluidNC/src/WebUI/WifiConfig.cpp
atlaste: Changed what I could. Commit 84ae213bcd6cdd2b1fc4f320d1ed984b3539efa3

FluidNC/src/OLED.cpp
atlaste: I didn't delete it because it seems like a bug. The code assigns some variables, but never uses them. I could just delete the whole tag - but that seems wrong to me. Not sure what the idea was here.

FluidNC/include/Driver/fluidnc_uart.h
atlaste: You're right. But both `fnc_uart_read_bytes` and `uart_read_bytes` (IDF) use uint32_t for len. If I modify this, it would need a type cast. I figured it's better to just pass it along in this specific case.

FluidNC/esp32/esp32s3/USBCDCChannel.cpp
atlaste: I solved this a different way. The USBCDCChannel.cpp code is the same as the original. See above.

FluidNC/esp32/esp32s3/USBCDCChannel.cpp
atlaste: Unfortunately this turned out to be a nasty one. ESP-IDF and Arduino are fighting over TinyUSB, because Arduino apparently depends on an old version of some github project, and Espressif decided to port the latest TinyUSB to their framework. At TinyUSB they don't seem to be bothered with backward compatibility.

So, I decided to just bite the bullet and do it properly. The new implementation in commit 7d4a4e0579ed7417732f1c212c112e3ff7626b05 has an USBCDCChannel_IDF cpp / h file, which implements the TinyUSB using USB/ACM without Arduino. I had to override Arduino's SDKConfig to get rid of the conflict, because Arduino didn't play nice with dependencies here. That solves it.

The implementation is pretty much the same as the USBCDCChannel, but uses the new TinyUSB API. I still do need to properly test it (by uploading a file etc), so I'll leave this conversation open.

MitchBradley: That's very interesting.  I will look into using your Arduino-free channel, in order to cut out the middleman.  I suppose that I might need to do some versioning work if the TinyUSB API has changed.

What aspects of SDKConfig were troublesome?

MitchBradley:
I incorporated all of this into S3Async patch by patch, fixing conflicts that had arisen.  The current version of S3Async, which is about to be promoted to main, has all of this stuff, although the ifdefs in the USBCDC area are somewhat different.

FluidNC/esp32/esp32s3/USBCDCChannel.cpp
anvirao: Hello, how can I make this condition true? If I directly remove this macro, it will cause a compilation error. Thank you!~
ESP_IDF_VERSION_MAJOR >= 5

PR #1608, opened 2025-11-09
Fix Bug when using Airedale / Expansion Pin as a User Input

pionex:
When using an Expansion Pin as a User Input, M66 always read 0.  This fix changes the pin type in the UserInput class to InputPin and uses the get() method.  M66 now appears to work.  

I have not yet regression tested to see how this affects non expansion inputs.  I just wanted to post this in case it helps someone with a better understanding of the input classes.  There is an issue where UserInput creates a std::array of the pins, but InputPin requires the pin name to be provided in the ctor.  It would require a number of changes to make this work (array of InputPin* maybe), so right now this just puts a generic pin name / legend so the array can be created during class initialization.

### DIFF
--- a/FluidNC/src/Machine/UserInputs.cpp
+++ b/FluidNC/src/Machine/UserInputs.cpp
@@ -24,13 +24,13 @@ namespace Machine {
     void UserInputs::init() {
         for (auto& input : _analogInput) {
             if (input.pin.defined()) {
-                input.pin.setAttr(Pin::Attr::Input);
+                input.pin.init();  // This will register for events
                 log_info("User Analog Input: " << input.name << " on Pin " << input.pin.name());
             }
         }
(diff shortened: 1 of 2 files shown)


MitchBradley:
Thanks for pointing out these problems.  The M66 implementation was provided by a user and apparently I did not review it carefully enough.  The adverse interaction with the configuration system goes fairly deep.

alpha1mm:
I’m not sure if this is expected behavior but

M66 P0 L0 D#5399 

on a single line will print the old value before M66 was called

but when the GCodes are on separate lines, it works as expected.

MitchBradley:
D is evaluated immediately before the rest of the block is processed.

I am working on a fix for M66

MitchBradley:
The S3Async branch contains a comprehensive fix for this problem

PR #1624, opened 2025-12-07
add little endian support to `ModbusVFD`

ajmirsky:
Certain spindle controllers transmit and receive RPM information as little endian:

a spindle speed of 1000 (0x00C8) needs to be sent as `01 06 80 05 C8 00 E7 CB`

the configuration now supports the `le` modifier in addition to the existing scaling of `[*][/][%]`:

`set_rpm_cmd: 06 80 05 le rpm > echo`

and

` get_rpm_cmd: 03 80 18 00 01 > 03 02 le rpm*20/4`

### DIFF
--- a/FluidNC/src/Spindles/Spindle.cpp
+++ b/FluidNC/src/Spindles/Spindle.cpp
@@ -185,6 +185,12 @@ namespace Spindles {
                 break;
             }
         }
+
+        // if the offset is that max value of uint32, then the offset was never set. therefore, bypass the mapping process
+        if(_speeds[i].offset == -1) {
+            return speed;
+        }
+
         uint32_t dev_speed = _speeds[i].offset;
 
         // If the requested speed is greater than the maximum map speed,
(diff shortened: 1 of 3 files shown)


ajmirsky:
Support for Stepper Online's Brushless Driver BLD-510B

image

ajmirsky:
once this pull request is accepted, I can update the wiki with the following:

updates for the wiki [Modbus VFD](http://wiki.fluidnc.com/en/config/config_spindles#generic-modbusvfd-rs485) section

> Speeds are given in RPM, converted to and from device-specific units via scaling specifiers in the command templates. Most VFDs specify speeds in units of Hz, possibly scaled by a factor of 10 or 100. To convert from Hz to RPM, you would multiply by 60 (seconds per minute). If the VFD uses deciHz (Hz*10), you would multiply by 60 and divide by 10 - or you could just multiply by 6. If the VFD transmits the values in little endian format, add the `le` keyword prior to the rpm (or minrpm/maxrpm) keyword.

and add another section following Scaling Specifiers:

> The named data items  "rpm", "minrpm" or "maxrpm" can be sent in little endian format where the lowest order byte of the value is sent first, followed by the high order byte:

image

> In this case, modbus commands should add the `le` specifier prior to the scaling operations:

image

- for **get**... commands, the two-byte value will be decoded from little endian to an int before applying the scaling
- for **set**... commands, the int will be scaled and then encoded as little endian before sending

PR #1626, opened 2025-12-08
Fix S3 chip issues

bdring:


install_scripts/posix/install-fs.sh
```diff
@@ -6,6 +6,9 @@ if ! . ./tools.sh; then exit 1; fi
 
 if ! check_security; then exit 1; fi
 
+chip="esp32"
+export chip
+
```
MitchBradley, line 10: You can do this on one line as
```sh
export chip="esp32"
```

install_scripts/posix/tools.sh
```diff
@@ -89,8 +97,14 @@ run_fluidterm() {
 }
 
 install() {
-    if ! check_security; then
-        exit
+    BaseArgs="--chip $chip --baud 230400"
+
+    echo $BaseArgs
+    
+    if [ "$chip" != "esp32s3"]; then
```
MitchBradley: This would be better as 
```sh
if [ "$chip" == "esp32"]; then
```
because the checksecurity thing is only valid for esp32, not for anything that is not esp32s3

install_scripts/posix/tools.sh
```diff
@@ -38,6 +38,7 @@ EsptoolPath=esptool.py
 
 BaseArgs="--chip esp32 --baud 230400"
```
MitchBradley: Should this be 
```
BaseArgs="--chip $chip --baud 230400"
```
then omit the redefinition of BaseArgs several places below?

install_scripts/posix/tools.sh
```diff
@@ -73,11 +77,15 @@ check_security() {
...
 
 esptool_erase() {
-    if ! check_security; then
-        deactivate
-        return 1
-    fi
-    esptool-basic erase-flash
+    BaseArgs="--chip $chip --baud 230400"
+    if [ "$chip" != "esp32s3"]; then
+        if ! check_security; then
```
MitchBradley, line 82: See comment on line 104

install_scripts/posix/tools.sh
bdring: I thought that would force chips other than ESP32 and ESP32S3 to fail early.

MitchBradley: If the chip is neither 32 or 32s3, the first esptool operation that you try will fail because --chip does not match the actual chip type.  It seems to me that it is less confusing for an operation that you actually want to perform to fail, instead of failing checksecurity which you do not really want to do on say, an S2.

PR #1639, opened 2026-01-04
Uninitialized variable broke WebUI3

MitchBradley:
[ESP800]json=yes was reporting ??? instead of XYZ due to an uninitialized variable

### DIFF
--- a/FluidNC/src/WebUI/WifiConfig.cpp
+++ b/FluidNC/src/WebUI/WifiConfig.cpp
@@ -604,7 +604,7 @@ namespace WebUI {
                 j.member("HostPath", "/");
                 j.member("Time", "none");
                 std::string axisLetters;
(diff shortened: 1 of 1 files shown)


FluidNC/src/WebUI/WifiConfig.cpp
```diff
@@ -604,7 +604,7 @@ namespace WebUI {
                 j.member("HostPath", "/");
                 j.member("Time", "none");
                 std::string axisLetters;
-                for (axis_t axis; axis < Axes::_numberAxis; axis++) {
+                for (axis_t axis = X_AXIS; axis < Axes::_numberAxis; axis++) {
```
540lyle, line 607: Consider adding unit test

MitchBradley: If you would add a full suite of unit tests, it would be greatly appreciated.

PR #1641, opened 2026-01-04
Add $HTTP command for outgoing HTTP requests from GCode

appositeit:
## Summary

This PR adds a `$HTTP` command that enables CNC machines to communicate with external systems during job execution. The goal is to support turning on air/water/etc automatically as part of the headers in gcode:

- **HTTP/HTTPS support** - Works with both protocols (HTTPS uses insecure mode for self-signed certs)
- **Custom headers and body** - Full control over request configuration
- **JSON value extraction** - Parse numeric values from JSON responses into GCode parameters
- **State-aware** - Blocks requests during unsafe states (Homing, Jog, Sleep, etc.)
- **Configurable error handling** - `"fail":false` option allows non-critical requests to continue on error

JSON value extraction can be used to return FLOAT values and store them in gcode params. I have no specific use case for this, but I think it could allow some clever automations.

### Limitations

This command is blocked from being used while the spindle/laser is on or the head is moving. There are no guarantees about overheads associated with HTTP requests so I don't want to risk interfering in a potentially destructive operation.
 
### Example Usage

```gcode
; Simple GET request
$HTTP=http://server/api/status

; POST with headers and body
$HTTP=https://homeassistant:8123/api/services/switch/turn_on{"headers":{"Authorization":"Bearer TOKEN"},"body":"{\"entity_id\":\"switch.laser_water\"}"}

; Extract values from JSON response
$HTTP=http://sensor/api{"extract":{"_temp":"temperature","_humid":"humidity"}}
; Now #<_temp> and #<_humid> contain the extracted values

; Non-blocking request (continues on error)
$HTTP=http://metrics.local/log{"fail":false}
```

### Files Changed

- `FluidNC/src/WebUI/HttpCommand.cpp` - Main implementation
- `FluidNC/src/WebUI/HttpCommand.h` - Header file
- `FluidNC/src/Parameters.cpp` - Fixed function signature mismatch
- `doc/overview.md` - Documentation
- `platformio.ini` - Added json-streaming-parser dependency

## Test Plan

- [x] Build succeeds for wifi environment
- [x] HTTP requests to local server work
- [x] HTTPS requests to external servers work (google.com, httpbin.org)
- [x] JSON value extraction works correctly
- [x] Parameters `#<_HTTP_STATUS>` and `#<_HTTP_RESPONSE_LEN>` are set
- [x] Test on actual CNC hardware with Home Assistant integration

### DIFF
--- a/examples/http_command/README_HTTP.md
+++ b/examples/http_command/README_HTTP.md
@@ -27,7 +27,7 @@ A comprehensive settings file showing all HttpCommand features:
   - Tool management: `fetch_tool_offsets`
   - Different HTTP methods: PUT (`put_job_status`), DELETE (`delete_old_logs`)
 
-### `http_example.ngc`
+### `http_example.gc`
 A G-code example program demonstrating how to use the commands and tokens from `http_settings.json`:
 
 - **Simple command usage** - Basic command substitution with `@command_name`
@@ -45,7 +45,7 @@ A G-code example program demonstrating how to use the commands and tokens from `
 
 1. Copy `http_settings.json` to `/localfs/http_settings.json` on your FluidNC device/SD card
 2. Update the URLs and tokens with your actual API endpoints and credentials
-3. Use `http_example.ngc` as a reference for your own G-code programs
+3. Use `http_example.gc` as a reference for your own G-code programs
 
 ### Editing http_settings.json
 
@@ -249,7 +249,7 @@ cp http_settings.json /mnt/sd_card/localfs/
 
 #### 3. Run test G-code
 
-Using `http_example.ngc` or your own test program:
+Using `http_example.gc` or your own test program:
 
 ```gcode
 ; Simple extraction test


MitchBradley:
Would it be feasible to use https://github.com/MitchBradley/json-streaming-parser , as used in FluidDial, for the JSON parsing?  I would prefer to maintain one thing instead of two.

appositeit:
Yes, for sure. I'm all for maintainability!

Jenna

appositeit:
* I've changed over to json-streaming-parser
* I finally got around to testing this on my actual laser machine (which required some upgrades and my machine is super crammed so it was awkard!)

I added a little proxy because I didn't want to futz with the 256 character limit. I figure this is a fairly common use case, I'm but I'm also happy to remove it.

To be transparent: Claude Code wrote this change. I just had a lot of input. I won't be offended if you don't like anything and want it changed.


I think the documentation on or linked from http://wiki.fluidnc.com/en/features/commands_and_settings should read something like this, I'm happy to update at the appropriate moment.

$HTTP

  Makes outgoing HTTP requests from GCode to external services like Home Assistant, webhooks, or REST APIs.

  Syntax:
  $HTTP=url
  $HTTP=url{json_options}

  The optional JSON object can include:
  - method - HTTP method (GET, POST, PUT, DELETE). Default: GET, or POST if body is present
  - timeout - Request timeout in milliseconds. Default: 5000, Maximum: 10000
  - body - Request body as a string. JSON must be escaped: "{\"key\":\"value\"}"
  - headers - Object of custom HTTP headers, e.g. {"Authorization":"Bearer xyz"}
  - extract - Object mapping GCode parameter names to JSON response keys
  - halt_on_error - If true (default), errors halt GCode. If false, execution continues

  After each request, these global parameters are set:
  - #<_HTTP_STATUS> - HTTP status code (200, 404, etc). Set to 0 if connection failed
  - #<_HTTP_RESPONSE_LEN> - Length of response body

  Parameters defined in the extract option are also set. All GCode parameters are floats, so only numeric JSON values can be extracted.

  Examples:

  Simple GET request:
  $HTTP=http://192.168.1.100:5050/status

  POST with JSON body:
  $HTTP=http://192.168.1.100/api{"method":"POST","body":"{\"action\":\"start\"}"}

  Non-critical request (continue on failure):
  $HTTP=http://192.168.1.100/log{"halt_on_error":false}

  Extract values from JSON response and use in GCode:
  ; Server returns: {"temperature": 25.5, "humidity": 65.2}
  $HTTP=http://192.168.1.100/sensors{"extract":{"_temp":"temperature","_humid":"humidity"}}

  ; Use extracted values in conditional logic
  #<_safe> = [#<_temp> LT 50]
  o100 if [#<_safe> EQ 0]
    M0 (Temperature too high!)
  o100 endif

  Limitations:
  - Blocks GCode line processing (stepper motion continues independently)
  - Maximum timeout: 10 seconds
  - Response body truncated to 256 characters
  - Only works when WiFi is connected in STA mode
  - Commands limited to 255 characters total
  - HTTPS supported but certificates not validated

  Working with long authentication tokens:

  Services like Home Assistant require JWT tokens that exceed the 255-character line limit. Use a proxy server with short URLs. See examples/ha-proxy/ in the FluidNC repository.

MitchBradley:
The code looks good.  I wonder if the proxy is a barrier to use?  It creates the requirement for an external machine to run the proxy.  Is there a way to store the long tokens in a file on localfs?  They could be cached in RAM on init if the localfs access during execution is a problem.

appositeit:
Oh, that's a good idea. How about this:


┌────────────────┬────────────────────────────────────────────────────────┐
  │    Feature     │                      Description
│

├────────────────┼────────────────────────────────────────────────────────┤
  │ Token file     │ /localfs/.http_tokens - key=value format, one per line
│

├────────────────┼────────────────────────────────────────────────────────┤
  │ Substitution   │ ${token_name} replaced in URL, headers, and body
│

├────────────────┼────────────────────────────────────────────────────────┤
  │ Reload command │ $HTTP/Tokens/Reload - reload tokens without reboot
│

├────────────────┼────────────────────────────────────────────────────────┤
  │ Documentation  │ Updated header comments with examples
 │

└────────────────┴────────────────────────────────────────────────────────┘
  Example usage:

  Token file (/localfs/.http_tokens):
  ha_token=Bearer eyJhbGciOiJIUzI1NiIs...

  GCode:
  $HTTP=http://homeassistant:8123/api/services/switch/turn_on{
"headers":{"Authorization":"${ha_token}"},"body":"{\"entity_id\":\"switch.laser_water\"}"}

  This eliminates the need for the proxy in simple auth scenarios.

I guess I should plumb a setting for the .http_tokens file?

MitchBradley:
How about /localfs/http_settings.json

No . prefix and not configurable.

appositeit:
Good with that, I'd prefer not to have to make it configurable!  😅

appositeit:
Based on Mitch's suggestion (which I agree with :)!):
* Tokens are now stored /localfs/http_settings.json
* Tokens are a sub-dict allowing us to add configuration in the future if necessary
* I've removed the proxy from the PR since it's now unnecesary.
* Tested on my a test ESP32 and my K40.

appositeit:
I think this ready for another round of review.

MitchBradley:
I will get to it soon, I hope.  I am fighting with MacOS's WebDAVFS which is requiring a lot of focus.

appositeit:
Np problem at all. There's no hurry, other than my general excitement. For
my purposes I'm already running the code.

Thank you- best of luck with webdavfs!

monken:
Is there any risk that this could get abused by malicious NC files? Maybe put it behind a feature flag that is disabled by default?

FluidNC/src/Parameters.cpp
```diff
@@ -482,7 +482,7 @@ bool get_param_ref(const char* line, size_t& pos, param_ref_t& param_ref) {
     }
 }
 
-bool set_named_param(const std::string& name, float value) {
+bool set_named_param(const char* name, float value) {
```
MitchBradley, line 498: I am trying to weed out C strings, so a better change is to fix the call signature in Parameters.h to use std::string& instead of const char*.  That was an omission that the compiler did not catch because there were no external uses of set_named_params(), and the internal use in Parameters.cpp picked up the std::string one.

FluidNC/src/WebUI/HttpCommand.cpp
```diff
@@ -0,0 +1,710 @@
...
+        return Error::Ok;
+    }
+
+    // Module for registering the HTTP command during system initialization
+    class HttpCommandModule : public Module {
+    public:
+        explicit HttpCommandModule(const char* name) : Module(name) {}
+
+        void init() override {
+            new UserCommand("HTTP", "Custom/HTTP", http_command_handler, http_state_check, WG);
```
MitchBradley: Let's name it HTTP/GET.  There are a number of other commands that start with HTTP/ and it seems presumptuous to take over the root of that name space.

FluidNC/src/WebUI/HttpCommand.cpp
```diff
@@ -0,0 +1,710 @@
...
+//
+// Token substitution:
+//   Long tokens (like JWT auth tokens) can be stored in /localfs/http_settings.json and
+//   referenced using ${token_name} syntax. This avoids the 255-char GCode line limit.
+//
+//   Settings file format (JSON):
+//     {"tokens":{"ha_token":"Bearer eyJhbGciOiJIUzI1NiIs...","api_key":"sk-abc123"}}
+//
+//   Usage in GCode:
+//     $HTTP=http://ha:8123/api{"headers":{"Authorization":"${ha_token}"}}
```
MitchBradley: Are we at risk of overflowing the command line even with the tokens delegated to files?  I could imagine people wanting to use longer response maps.  Perhaps it would be better to delegate even more of the command to the settings file.

FluidNC/src/WebUI/HttpCommand.cpp
```diff
@@ -0,0 +1,710 @@
...
+//
+//   Tokens are loaded at startup. Use $HTTP/Tokens/Reload to reload without reboot.
+//
+// Examples:
+//   $HTTP=http://example.com/api
```
MitchBradley: I wonder if there is some public server that could be used for testing this.  If not, perhaps there could be a simple Python script to run on a local machine to test against.

MitchBradley: I am ready to merge this if we can figure out a good way to test/exercise it.

MitchBradley:
@appositeit 

I think I have addressed all of my concerns (thanks to Claude for making it go faster) ...

* Created a Python test server to respond to a range of example requests
* The command name is now $HTTP/GET AKA $HG
* There is now a "commands" section of the http_settings.json file, with keys that are aliases for longer command strings.  If you say $HTTP/GET=@foo it will look for a "foo" key and substitute the value.
* There is no longer a response length limit; it just pushes the response bytes through the streaming parser as they come in.

Would it be okay with you if I push my changes directly to this PR?

appositeit:
Please, go ahead. Apologies for not being very responsive.

Jenna

MitchBradley:
@copilot please change "M118 message" to "(PRINT, message)"

MitchBradley:
@copilot Ignore the previous message; the fix has already been done.

MitchBradley:
It would be interesting to see how much of the code in NotificationsService.cpp could be subsumed by this new feature.  Perhaps everything except the Email service.  NotificationsService is a bit of a sore spot since it has several functions for specialized services that are rarely used and perhaps even obsolete.

PR #1644, opened 2026-01-07
Fixed spindle accel delay on spin down

bdring:
- Added a debug report of the delay

### DIFF
--- a/FluidNC/src/Spindles/Spindle.cpp
+++ b/FluidNC/src/Spindles/Spindle.cpp
@@ -272,12 +272,12 @@ namespace Spindles {
         int32_t dwell = 0;
         if (down) {
             dwell = (down < maxSpeed() ? _spindown_ms * down / maxSpeed() : _spindown_ms);
+            log_debug("Spin down delay ms:" << dwell);
+            dwell_ms((uint32_t)dwell, DwellMode::SysSuspend);
         }
         if (up) {
             dwell = (up < maxSpeed() ? _spinup_ms * up / maxSpeed() : _spinup_ms);
-        }
-        if (dwell) {
-            log_debug("Spindle delay ms:" << dwell);
+            log_debug("Spin up delay ms:" << dwell);
             dwell_ms((uint32_t)dwell, DwellMode::SysSuspend);
         }
 


FluidNC/src/Spindles/Spindle.cpp
```diff
@@ -269,12 +269,18 @@ namespace Spindles {
...
         }
+        int32_t dwell = 0;
         if (down) {
-            dwell_ms(down < maxSpeed() ? _spindown_ms * down / maxSpeed() : _spindown_ms, DwellMode::SysSuspend);
+            dwell = (down < maxSpeed() ? _spindown_ms * down / maxSpeed() : _spindown_ms);
         }
         if (up) {
-            dwell_ms(up < maxSpeed() ? _spinup_ms * up / maxSpeed() : _spinup_ms, DwellMode::SysSuspend);
+            dwell = (up < maxSpeed() ? _spinup_ms * up / maxSpeed() : _spinup_ms);
+        }
```
MitchBradley: This changes the logic.  Before, you could have two separate delays, one for spinning down and one for spinning up.  Now there is only the one delay and up does it all, not accounting for the spindown time.  You could probably fix that by saying dwell += (up ...

bdring:
Sorry, accidentally clicked resolve.

I don't follow. What do you mean by separate delays. Are you talking about the 2 config values? I made down double what up is. Down now takes twice as long for the same rpm delta.

```
/10v:
  forward_pin: gpio.7
  reverse_pin: gpio.8
  pwm_hz: 5000
  direction_pin: NO PIN
  output_pin: gpio.6
  enable_pin: NO PIN
  spinup_ms: 5000
  spindown_ms: 5000
  tool_num: 0
  speed_map: 0=0.00% 0=33.33% 8000=33.33% 24000=100.00%
  off_on_alarm: false
  atc:
  m6_macro:
  s0_with_disable: true
  disable_with_s0: false
ok
$10v/spindown_ms=10000
ok
S2000
[MSG:DBG: Spindle delay ms:416]
ok
S0
[MSG:DBG: Spindle delay ms:833]
ok
s2000
[MSG:DBG: Spindle delay ms:416]
ok
```

MitchBradley:
No, I mean when switching directions.  Going from CCW to CW spins down then up, with a separate delay value for each of down and up.  That is why the preceding switch block is so complicated.  Otherwise you would not need to consider direction; you would just difference the start and end speeds.

bdring:
OK, I forgot that people do stupid things.

PR #1650, opened 2026-01-13
Revamp of filesystem volume name handling

bdring:
Simplifies the path canonicalization and eliminates a likely cause of crashes due to a non-thread-safe static char buffer.

### DIFF
--- a/FluidNC/src/Error.cpp
+++ b/FluidNC/src/Error.cpp
@@ -20,7 +20,7 @@ const std::map<Error, const char*> ErrorNames = {
     { Error::Overflow, "Line too long" },
     { Error::MaxStepRateExceeded, "Max step rate exceeded" },
     { Error::CheckDoor, "Check door" },
-    { Error::LineLengthExceeded, "Startup line too long" },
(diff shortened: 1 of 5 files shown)


bdring:
I created a pull request so I can add test results.

It seems to work ok with the standard commands used with a serial terminal. 

The Web Installer is still OK, but I don't see any subdirectory support. I think that was always the case.

What was the general intent of this change?  Basically, what should I do to test the new code.

MitchBradley:
Thanks for testing it.  The motivation was that mislabeled Discord issue that was caused by browser-induced crashes.  The crash dumps showed a memory corruption problem that was due to a static char buffer in the path canonicalization routine.  That static buffer was not thread safe and I think that the async web server was either reentering that routine, or else something in the std::filesystem::path code was holding a reference to that buffer instead of copying it to a safe place.  Either way, the contents of the buffer was changing at a strange time and causing the crash.
Fixing it entailed very careful use of std::string, and created a cascade of data type changes.

bdring:
Thanks.

I'll check the WebDAV too.

bdring:
WebDAV works fine for me

PR #1657, opened 2026-01-27
Feature/regression test suite

540lyle:
# Add Unit Test Infrastructure with Code Coverage

## Overview

This PR establishes a **cross-platform unit test infrastructure** with code coverage tooling for FluidNC. The focus is on **testable, pure utility code** with minimal dependencies, providing a foundation for incremental test expansion as code becomes more testable.

**Current coverage: 88.1% line coverage (258/293 lines) on tested modules** | **270 tests passing** ✅

### Cross-Platform Coverage Infrastructure
- **coverage.py** - Automated coverage script (works on Windows, macOS, Linux)
- **coverage_build.py** - PlatformIO extra script for gcov linking
- **tests_coverage** environment in platformio.ini
- HTML coverage reports with file-level and function-level detail

### New Test Module
- **FluidErrorTest.cpp** (5 tests, 100% coverage) - Error category system and message lookup
- **StringUtilTest.cpp** (69 tests, 98.8% coverage) - String trimming, parsing, comparison
- **UTF8Test.cpp** (59 tests, 95.4% coverage) - UTF-8 encoding/decoding with round-trip verification
- **UtilityTest.cpp** (32 tests) - Template utilities and conversion constants
- **RegexprTest.cpp** (72 tests, 100% coverage) - Pattern matching for setting names
- **ErrorBehaviorTest.cpp** (9 tests) - Error enum values and message lookup
- **StateTest.cpp** (7 tests) - Machine state values and ranges
- **CommandCompletionTest.cpp** (10 tests, 85.4% coverage) - Command completion logic
- **FluidErrorTest.cpp** (5 tests, 100% coverage) - Error category system
- **PinOptionsParserTest.cpp** (7 tests, 100% coverage) - Pin option parsing

### Documentation
- **FluidNC/test/UnitTests.md** - Updated with:
  - How to run tests (`pio test -e tests`)
  - Coverage setup and usage
  - Troubleshooting guide

### Pre-commit Hook
- Updated to run `tests` environment (fast, no coverage overhead)
- Skips on empty commits, provides escape hatch via `FLUIDNC_SKIP_TESTS=1`

## Coverage Reality Check

⚠️ **Important:** Coverage metrics reflect **only the 11 source files currently under test**. FluidNC has ~100+ source files total.

**What's tested (293 lines total):**
- ✅ Pure utility code (string parsing, UTF-8, regex)
- ✅ Error categorization system
- ✅ Command completion logic
- ✅ Pin option parsing

**What's NOT tested (vast majority of codebase):**
- ❌ GCode parser and execution engine
- ❌ Motion planning and kinematics
- ❌ Spindle, coolant, and tool control
- ❌ Homing and probing sequences
- ❌ Configuration system (YAML parsing)
- ❌ Communication protocols (Serial, WebSocket, Telnet)
- ❌ File system operations

**Why the gap?** Most FluidNC code has tight coupling to:
- FreeRTOS kernel (tasks, queues, timers)
- Hardware abstractions (GPIO, I2C, SPI, UART)
- System globals (machine state, configuration)

These require either integration tests on real hardware or significant refactoring to enable mocking/stubbing.

### Approaches Tried and Reverted

1. **❌ Shallow enum validation tests** 
   - Initially added `GCodeParserTest.cpp` and `AlarmSystemTest.cpp` that only checked enum integer values
   - **Why reverted:** Zero behavior coverage, tests only caught protocol constant changes (should be code review)
   - **Lesson:** Tests must exercise actual code paths, not just validate compile-time constants

2. **❌ TestCategories.h framework**
   - Created macro framework for categorizing tests (SMOKE_TEST, SANITY_TEST, etc.)
   - **Why reverted:** Broken includes, undefined macros, duplicated constants from headers, never actually used
   - **Lesson:** Don't build infrastructure before you need it

3. **❌ Testing heavily-coupled modules**
   - Attempted to test `FluidPath.cpp` (115 LOC) and `NutsBolts.cpp` (240 LOC)
   - **Why reverted:** Deep FreeRTOS/filesystem dependencies required extensive mocking
   - **Lesson:** Focus on pure utility code first; integration tests or refactoring needed for coupled code

### What Actually Worked

✅ **Coverage tooling first** - `coverage.py` + `coverage_build.py` made gaps immediately visible  
✅ **Test pure utilities** - String parsing, UTF-8, regex have zero external dependencies  
✅ **FluidErrorTest.cpp** - Small module (19 LOC), pure utility, quick win at 100% coverage  
✅ **Fast feedback loop** - 8-second test runs encourage frequent execution  
✅ **Cross-platform from day one** - Works on Windows/macOS/Linux without special setup

### Foundation for Quality
- Provides **infrastructure** for adding tests as code gets refactored
- **Coverage visibility** shows what's tested vs. untested (currently: 293 lines of ~10,000+)
- **Fast feedback** - tests run in ~8 seconds

### Practical Scope
This PR focuses on **pure utility code** that can be unit tested without mocking. The 88% coverage number is **only for these 11 testable files**, not the entire codebase.

### Coverage Highlights (Tested Files Only)
- **100% coverage**: FluidError.cpp, Regexpr.cpp, PinOptionsParser.cpp
- **98.8% coverage**: string_util.cpp
- **95.4% coverage**: UTF8.cpp
- **88.1% overall**: 258 of 293 lines in modules included in test build

## How to Use

```bash
# Run tests (recommended for CI/pre-commit)
pio test -e tests

# Run tests with coverage instrumentation
pio test -e tests_coverage

# Generate text coverage report
python coverage.py

# Generate interactive HTML coverage report
python coverage.py --html

# Generate detailed coverage with branch analysis
python coverage.py --html --verbose

### DIFF
--- a/FluidNC/capture/freertos/task.h
+++ b/FluidNC/capture/freertos/task.h
@@ -2,6 +2,7 @@
 
 #include "FreeRTOS.h"
 #include "FreeRTOSTypes.h"
+#include <climits>
 
 void vTaskDelay(const TickType_t xTicksToDelay);
 

### CODE REVIEW

File: FluidNC/test/UnitTests.md [hunk 1]
```diff
@@ -15,6 +15,20 @@ Running unit tests is done by:
 
 `pio test -e native`
 
+## Test suites
+
+The native unit test environment (`-e native`) builds a reduced subset of the firmware along with a set of GoogleTest suites.
```
540lyle: Native is redundant. Should make a call on whether we want it to be default test env or just remove it for more explicitly named common test env

MitchBradley:
The function tests - stringutil, utf8, utility, regex - look good, but the tests that just regurgitate enum values seem like they are too shallow to be of much value.  I would argue that, if a test could be performed by looking only at a single .h file, then it is pretty trivial.

MitchBradley:
I just submitted PR #1658 which fixes a deficiency in command completion.  It would be nice to have some tests for that, but I don't understand the test scaffolding well enough to add such tests, which would involve sending characters to the command interpreter and parsing what is echoed.

I suppose that the test could be done at a lower level by calling num_initial_matches() and looking at the results.  That would test the base completion mechanism without involving the presentation layer.

FluidNC/test/UnitTests.md
MitchBradley: We don't need the native env

MitchBradley:
It's getting better ...

I tried to add a command completion test that calls num_initial_matches() on the string "/axes/", expecting there to be the same number of matches as there are configured axes.  It didn't work because of include problems.  The inclusion structure tried to pull in files that weren't on the path, or that conflicted with other things.

Thinking about how things should work to be truly useful:
1. The test build should be relatively complete, akin to the windows_x86 build but with more generic versions of stuff - like the windows serial I/O could just be putchar() and whatnot.
2. It should load the builtin default config file and check the startup message output.
3. Actually it should do most of the init process in the existing main.

As it is, the targeted unit tests on isolated routines is helpful, but not hugely so.

540lyle:
> It's getting better ...
> 
> I tried to add a command completion test that calls num_initial_matches() on the string "/axes/", expecting there to be the same number of matches as there are configured axes. It didn't work because of include problems. The inclusion structure tried to pull in files that weren't on the path, or that conflicted with other things.
> 
> Thinking about how things should work to be truly useful:
> 
> 1. The test build should be relatively complete, akin to the windows_x86 build but with more generic versions of stuff - like the windows serial I/O could just be putchar() and whatnot.
> 2. It should load the builtin default config file and check the startup message output.
> 3. Actually it should do most of the init process in the existing main.
> 
> As it is, the targeted unit tests on isolated routines is helpful, but not hugely so.

I've cleaned this up, added an option for code coverage and reproting, but Its going to be tricky to get deeper penetration in unit tests and maintain their isolation. What you're asking for is more of a functional or integration test suite. Consider introducing a mocking framework to get _meaninful_ code coverage via unit tests and then expanding the next level of testing infrastructure. 

image

MitchBradley:
I resolved some merge conflicts with recent changes to main and merged the result, so this PR is effectively done.  Thanks for doing it.  There is now a posix build env that is pretty comprehensive - in addition to the windows_x86 branch that is nearly identical except for the different Console driver implementations.  They could be used for integration tests.  Their common code includes a VFD simulator that implements the Huanyang protocol. so it is possible to test the VFD code.

PR #1663, opened 2026-01-28
Most VFDs are now implemented with GenericProtocol

MitchBradley:
but you still declare them with, e.g. "Huanyang:"

Also the No response message can no be suppressed by setting debug: to 0 or 1 in the vfd section.


FluidNC/src/Spindles/VFD/H100.cpp
```diff
@@ -0,0 +1,24 @@
...
+                                0xffffffff,                              // min_rpm
+                                0xffffffff,                              // max_rpm
+                                "05 00 49 ff 00 > echo",                 // cw
+                                "05 00 4A ff 00 > echo",                 // ccw
+                                "05 00 4B ff 00 > echo",                 // off
+                                "06 02 01 rpm%*4 > echo",                // set_rpm
+                                "04 00 00 00 02 > 04 04 rpm%*4 ignore",  // get_rpm
+                                "03 00 0B 00 01 > 03 02 minrpm*60",      // get_min_rpm
+                                "03 00 05 00 01 > 03 02 maxrpm*60"       // get_max_rpm
+                ) {}
```
AxoInsanit, line 18: @MitchBradley are you sure about these values? I have it working with these:
```
ModbusVFD:
  model: H100
  uart_num: 1
  modbus_id: 1
  off_on_alarm: true
  cw_cmd:  05 00 49 ff 00 > echo
  ccw_cmd: 05 00 4a ff 00 > echo
  off_cmd: 05 00 4b ff 00 > echo
  set_rpm_cmd: 06 02 01 rpm*10/60 > echo                                #different
  get_rpm_cmd: 04 00 00 00 02 > 04 04 rpm*60/10 ignore        #different
  get_min_rpm_cmd: 03 00 0b 00 01 > 03 02 minrpm*60/10     #different
  get_max_rpm_cmd: 03 00 05 00 01 > 03 02 maxrpm*60/10   #different
  s0_with_disable: true                         #needed for safety_polling
```

and `bool safety_polling() const override { return true; }` on GenericProtocol.h

FluidNC/src/Spindles/VFD/GenericProtocol.h
AxoInsanit, line 39: shouldn't this be true so we get speed monitoring?

FluidNC/src/Spindles/VFD/H100.cpp
MitchBradley: How would you suggest handling the fact that the speed resolution is configurable?
image

MitchBradley: I do not have an H100 VFD myself.  Different people have reported using different commands, so I am not sure what to believe.  Perhaps you could try the different variants, along with different settings for F169, and add your results to the wiki.  It would be great if someone who actually has an H100 could assume responsibility for documentation, support, and maintenance.

FluidNC/src/Spindles/VFD/GenericProtocol.h
MitchBradley: Probably, but since it applies to every VFD, it would need a lot of testing that we are not set up to do directly.  Would you be able to coordinate user testing?

AxoInsanit: It's working really well for me for the H100 with the config I added on other comment, I did some crazy stress test on M5s to M3s on a loop that the previous H100 driver failed consistently.
I can try helping with the user testing, I'm on discord as axo6666
Maybe moving it to a YML config entry per spindle would make it more manageable

FluidNC/src/Spindles/VFD/H100.cpp
AxoInsanit: That's a good point, mine has F169 on 0 (the default), and it's working great with this config, not sure there's any reason to  have 2 bit decimal
I can take over the h100 on the wiki and document my h100 settings, I have a gpenny 1k5/h100 combo that I think it's very popular in AE

MitchBradley: Okay I will change to your version after I take care of a couple of other things I am doing.  Meanwhile, could you add the info about F169 to the wiki?  I agree that hundredths of a hz is pointless

AxoInsanit: will do, just registered to the wiki but kinda crashed trying to send me the verification email

bdring: The wiki does not have an email server. I verified you and gave you edit rights.

AxoInsanit: thanks, just updated

AxoInsanit: added my H100 functions settings as well

PR #1685, opened 2026-02-26
Improve VFD spindle speed synchronization handling and abort condition

monken:
Occasionally, FluidNC 4.0.1 will go into a loop reading the spindle speed after a job finishes.

My first PR, let me know if I missed anything obvious.

* `unchanged` is now properly counted: tracks how many consecutive readings returned the same speed value. After 100 unchanged readings (limit = 100, ~25 seconds at default 250ms poll rate), the loop exits.

* `sys.abort()` is checked each iteration: if a reset fires while syncing, _syncing is cleared and the function returns immediately.


bdring:
I pulled your changes into this PR and tested. It was easier than try to sync the PRs and main.

https://github.com/bdring/FluidNC/pull/1663

I'll close this PR soon.

Thanks for the work!

monken:
No worries, glad it was somewhat helpful!

Issue #1143, opened 2024-02-22
Feature: Allow any .htm(l)(.gz) to be viewed as webpage


woutervddn:
### Machine Context

I'm automating my tablesaw, basically it has 3 motors: fence, angle and sawblade height.

The interesting thing here is that the useful blade height is dependent on the angle.

So for this machine, the interface that makes most sense is one that only shows the relevant input parameters for the user and calculates the correct motor positions from there.

I made an example interface in 20 minutes that can also automatically subtract blade width or any other form of math in the input field. Or if you enter an angle that is below 44° that it knows that you actually want the compound angle (since the blade only goes from 92° to 44°)

I've got a video of the (demo) interface in question:

https://github.com/bdring/FluidNC/assets/1514131/44ad78ed-588f-418d-88a2-569acd190f18

### Feature Description

Allowing for multiple html pages to be viewed greatly reduces the barrier for machines to have specific interfaces. Only the needed buttons or elements would be available and the rest would be hidden.

Currently the tablet page is in a separate tab but making changes to that interface requires you to download the upstream project, understand the entire UI system, make the changes, compile and upload. Which is way harder than just having to create a webpage with basic fetch calls to the command interface.

The separate interface I built is (granted without the fetch calls being implemented) is 3.1KB

The functionality for this is also already there (in a way) as you can upload a custom 404 page which is my current workaround.

### Other Approaches

A custom interface could be added to a tab in the existing index page. But that requires you to understand it all and compile for each change.

Another approach would be to just make your interface and upload that as index.html.gz. This would result in the user needing to recreate all (or a lot) of the index.html.gz functionality (like file upload etc) which seems unrealistic. Or go without that functionality, but even replacing the index.html.gz back to the old version for a configuration change or maintenance would require more hoops to jump through.

Having a separate UI for 'daily use' and the default index.html.gz for configuration+maintenance just makes more sense in that way.

### How I Can Help

The change is small and I'm pretty sure that I can make it on my own, no experience writing tests though... Help in that regard and a second pair of eyes to verify what I wrote would be appreciated.

woutervddn:
Would the above WIP pull request be sufficient to solve this?

jeyeager65:
In FluidNC v4.0, you can load HTML using the WebDAV address.  So, http://fluidnc.local/flash/mycustomui.html will serve instead of download mycustomui.html or mycustomui.html.gz

PR #1695, opened 2026-03-12
Serve .html files instead of download

jeyeager65:
FluidNC will serve the index.html(.gz) file from the flash storage as the default file when accessing the root URL.  However, other .html(.gz) files are instead downloaded by the browser.  This will allow any .html(.gz) file to be rendered by the browser instead of downloaded.

This would be useful for custom interfaces.  The alternatives are to completely replace the WebUI index.html.gz file or to create a WebUI v3 extension.  Replacing the full WebUI is a lot of work since it allows for FluidNC configuration.  A WebUI v3 extension is ok but there are some formatting limitations.  You also cannot access a page extension by URL.

If you wanted to make it easier to access an HTML page from WebUI v3, it works to create a macro of Type "URL address" and an Action of "test.html".

This was tested by creating a basic HTML page as test.html.gz, saving to flash, and accessing via http://fluidnc.local/test.html.  Additionally, WebUI v3 extensions were also validated to still load properly since they are also .html.gz files.

A potential concern is that this means any standalone UI would have its own websocket connection, however, FluidNC v4.0 has better handling for this.

Since this still calls myStreamFile, it is still subject to all the normal handling such as block during motion, HTTP 304 responses, and .gz files.

Existing issue #1143 requests this and provides a draft PR.  However, that code is out of date for v4.0 and untested.  This provides an updated and simplified version.

### DIFF
--- a/FluidNC/src/WebUI/WebUIServer.cpp
+++ b/FluidNC/src/WebUI/WebUIServer.cpp
@@ -464,8 +464,10 @@ namespace WebUI {
             return;
         }
 
-        // Download a file.  The true forces a download instead of displaying the file
-        if (myStreamFile(request, path.c_str(), true)) {
+        // Serve HTML files inline so the browser renders them; download everything else
+        auto dot = path.rfind('.');
+        bool download = (dot == std::string::npos || path.substr(dot) != ".html");
+        if (myStreamFile(request, path.c_str(), download)) {
             return;
         }
 


MitchBradley:
Just out of curiosity, what happens if you access the .html file via the WebDAV route /flash/foo.html?

I sort of hope that it does the right thing because of content type decoding, but I haven't tried it.  If it works, it should also be possible to get a page from /sd/foo.html .

jeyeager65:
This works fine.  http://fluidnc.local/flash/test.html

I'm just running this off a bare ESP32 devkit at the moment.  I could try it with an SD card on a controller tomorrow.

MitchBradley:
Given that it is possible to get the desired behavior with a different route, what do you think about the tradeoffs of making the change vs leaving the existing behavior unchanged?  I doubt that many people are depending on the existing behavior, but it is hard to know.

jeyeager65:
Since it works with the webdav address, I would say that the existing functionality without this PR is fine.  I'm not sure there is any advantage to this then, other than a shorter URL, which is irrelevant.  Updating the URL macro to flash/test.html still works too.

PR #1704, opened 2026-03-29
USB Host

APIUM:
I have been testing this with another ESP32-S3 device for my pendant and it's been working well. I've not yested tested my changes to support changing the setting in the YAML, I'll try to do this tomorrow.


FluidNC/esp32/esp32s3/Console.cpp
```diff
@@ -21,13 +21,8 @@ class UartConsole : public UartChannel {
...
         UartChannel::init(uart0);
-
-#if defined(CONFIG_TINYUSB_CDC_ENABLED) && ESP_IDF_VERSION_MAJOR >= 5
-        auto cdc_enable = new EnumSetting("USB CDC Enable", WEBSET, WG, NULL, "USBCDC/Enable", true, &onoffOptions);
-        if (cdc_enable->get()) {
-            CDCChannel.init();
-        }
-#endif
+        // USB CDC/Host initialization is handled post-config in Main.cpp
+        // (YAML config must be loaded first to determine USB mode)
```
MitchBradley: Enabling CDC via an nvs setting was a very conscious choice, because it permits controllers that do not have an external USB-Serial chip to issue startup messages during the config process.  The config file should be though of as primarily defining the machine, not the basic communications.  That is why Web settings are in NVS.  There is still some need for secondary communications - expanders and pendants - to be in the config file, but the first level communications necessary for diagnostic messages must not depend on the config.

FluidNC/src/Machine/MachineConfig.cpp
```diff
@@ -4,6 +4,17 @@
 
 #include "MachineConfig.h"
 
+#include <sdkconfig.h>
```
MitchBradley: This puts an explicit ESP32 dependency in a file that is not supposed to have knowledge of ESP32.  It needs to be restructured to not have ESP32 names, perhaps with MAX_N_USB_OTG in Platform.h and the subordinate logic delegated to some file in esp32/esp32s3/

FluidNC/src/Machine/MachineConfig.h
```diff
@@ -20,6 +20,11 @@
 #include "Stepping.h"
 #include "Stepper.h"
 #include "UartChannel.h"
+#include <soc/soc_caps.h>
```
MitchBradley: Ditto about ESP32 dependency

FluidNC/src/Main.cpp
```diff
@@ -7,6 +7,17 @@
 #    include "Main.h"
 #    include "Machine/MachineConfig.h"
 
+#    include <sdkconfig.h>
```
MitchBradley: ESP32 dependency

libraries/usb_host/usb_host_vcp/library.json
```diff
@@ -0,0 +1,13 @@
+{
```
MitchBradley: Is there a way to fetch this code from somewhere else, instead of hosting several Espressif files in our tree?  (Yes I realize that fnc_idf_uart.c somewhat goes against this philosophy.) If Espressif does not have a suitable packaged repo, perhaps you could create a package in a repo in your own github area?

platformio.ini
```diff
@@ -181,14 +190,24 @@ build_flags =
...
-lib_deps = ${common.lib_deps} ${common.wifi_deps}
+lib_deps = ${common.lib_deps} ${common.wifi_deps} ${common_esp32s3.lib_deps_usb_host}
 build_src_filter =
     ${common_esp32s3.build_src_filter}
     ${common_wifi.build_src_filter}
 build_flags =
     ${common_esp32s3.build_flags}
     ${common_wifi.build_flags}
 
+[env:bt_s3]
```
MitchBradley: How did this come in?  ESP32-S3 only supports BT-LE, which does not work with FluidNC.

MitchBradley (commented):
Okay, this idea is sort of out of the blue and I wish I had thought of it sooner.  USB VCP is functionally nearly identical to a UART, so maybe it would make sense to bury it under the Uart class as, say, uart3.  Then there would be no need for new Channel types.  A USB UART would not be usable for TMC2209 interfaces or RS485, but a uart_channel that refers to a USB UART would probably Just Work.  OS's pretty much treat a USB UART just like any other.

MitchBradley (commented):
This is a really nice patch now.  I will test it when I get a little time.  I think that the USB Host capability can open up a lot of interesting possibilities.

One thing to think about - I'm not suggesting that you implement this in this PR, only give it some thought: If there was a GPIO monitoring the state of the USB VBus line, I wonder if the USB Device / Host selection could be automated somehow?  Presence of VBus indicates that an external host is connected so the ESP32 would have to be in device mode.  Lack of VBus voltage can mean either that nothing is connected or that an external device is connected.  From what I have read, that is about the best that the ESP32-S3 hardware can do.  It would require suitable wiring on the controller board.

FluidNC/esp32/esp32s3/Console.cpp
APIUM: OK. Have fixed this now. In order to have the pendant in the config you must disable CDC now in the nvs setting. I think this puts a big barrier to breaking your install without recovery via config.

FluidNC/src/Machine/MachineConfig.cpp
APIUM: Done.

FluidNC/src/Machine/MachineConfig.h
APIUM: Done

FluidNC/src/Main.cpp
APIUM: Done

libraries/usb_host/usb_host_vcp/library.json
APIUM: Done (https://github.com/APIUM/usb_host_vcp)

platformio.ini
APIUM: Removed

APIUM:
> This is a really nice patch now. I will test it when I get a little time. I think that the USB Host capability can open up a lot of interesting possibilities.

Thanks for taking a look Mitch. I was working on some changes last night and did get to test them. Tested now and made some fixes.

> One thing to think about

I like this. The best usecase is for ease of use, particularly for users who don't want to read the documentation (most, including myself). Removes the hassle of changing, and the risk of breaking your install. If it were built into all future boards it would be good, though ideally we'd have a way of informing the SW that we have that capability so that the user doesn't have to set it up. It's still set-and-forget though, which is a definitely improvement from now (although I'd expect most users would set up the pendant and then not change?).

bdring:
I did some testing on a Doberman. So far I have not seen any issues. I need to create a connected device to test this more. 

I might try a simple sketch on an M5Dial

bdring:
I have this running on a Doberman controller.

```yaml
uart3:
  usb_host: 
    baud: 1000000

uart_channel3: 
  report_interval_ms: 0
  uart_num: 3
  message_level: Verbose
  ```
    
 I have a M5Stack M5Dial connected that outputs "G91G0X10<cr><lf>" to the USB serial port when the dial button is pushed. I have verified this works by connecting FluidTerm to the dial.
  
 When I hot connect the dial to the doberman I get this.
 
 ```
 E (40752) HUB: Root port reset failed
[MSG:INFO: USB Host: device attached (VID:0x303a PID:0x8120)]
```

When I boot with it connected I get this...

```
Grbl 4.0 [FluidNC v4.0.2 (pr/1704-d92a1238-dirty) (esp32s3-wifi) '$' for help]
[MSG:INFO: USB Host: device attached (VID:0x303a PID:0x8120)]
```

When I click the dial button, nothing happens on FluidNC. No debug text outputs when I disconnect the dial.
 
I assume the baud rate is not important.

APIUM:
> I have a M5Stack M5Dial connected that outputs "G91G0X10" to the USB serial port when the dial button is pushed. I have verified this works by connecting FluidTerm to the dial.

@bdring 

Ah, yes I was only using VCP::open() from usb_host_vcp - so this wouldn't work with USB CDC. I've added a fallback now, hopefully it will work - I didn't get a chance to test it this morning I'll see if I get time this evening.

bdring:
It works now.

Here it is when connected at boot.

```
Grbl 4.0 [FluidNC v4.0.2 (pr/1704-f3708698-dirty) (esp32s3-wifi) '$' for help]
[MSG:INFO: USB Host: device attached (VID:0x303a PID:0x8120)]
[MSG:INFO: USB Host: device connected (CDC-ACM), baud 115200]
```

Here is a disconnect and reconnect while FluidNC is running.

```
[MSG:WARN: USB Host: device event 0]
E (150333) cdc_acm: cdc_acm_host_data_tx_blocking(947): Bulk OUT transfer error
[MSG:WARN: USB Host: E (150337) USBH: Device 1 gone
TX error: ESP_ERR_INVALID_RESPONSE]
[MSG:INFO: USB Host: device disconnected]
[MSG:INFO: USB Host: waiting for reconnect...]
E (158754) HUB: Root port reset failed
[MSG:INFO: USB Host: device attached (VID:0x303a PID:0x8120)]
[MSG:INFO: USB Host: device connected (CDC-ACM), baud 115200]
```

PR #1728, opened 2026-05-09
Fix1727

MitchBradley:
GCode.cpp was parsing G59.1-3, but failing to clear the mantissa, triggering the "not integer" error.  Also it was failing to explicitly reject G59.4+, although the "not integer" error effectively rejected that anyway.

Finally, $G was incorrectly reporting G59.1-3 as G60-62

### DIFF
--- a/FluidNC/src/GCode.cpp
+++ b/FluidNC/src/GCode.cpp
@@ -1668,7 +1668,7 @@ Error gc_execute_line(const char* input_line) {
             if (new_spindle) {
                 gc_state.spindle_speed = 0.0;
             }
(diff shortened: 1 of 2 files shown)


MitchBradley:
@copilot apply changes based on the comments in [this thread](https://github.com/bdring/FluidNC/pull/1728#pullrequestreview-4256067814)

bdring:
Tested G59.1 through G59.3 All worked with setting the csys, setting values with G10, and reporting with $# and $G.

G59.4 and G54.1 fail as they should. G10 L2 P10 X5 fails as it should.

- WebUI looks good

- WebUI3 shows 54-59, but not g59.1-59.3

image

- FigUI does not show the coordinate system, but does not have issues with G59.1

PR #1732, opened 2026-05-13
Add uptime and reset reason to $ESP420 output

aaronse:
Thoughts on these edits, or something like them, to help Users self diagnose and root cause job failures faster?

Adds a few ESP32 runtime fields to `$ESP420` / `System/Stats`:

* `uptime_ms`
* `reset_reason_code`
* `reset_reason_name`
* `previous_panic`

This helps users and support tools quickly tell whether a connection loss or job failure likely involved a controller reboot, brownout, panic, watchdog reset, or a network-only disconnect.

Kept intentionally small... no new command, no event history, no persistent boot ID.  I'd like to propose/contribute more, but figured narrow scope might be low risk and helpful enough to incorporate.

### DIFF
--- a/FluidNC/src/WebUI/WebCommands.cpp
+++ b/FluidNC/src/WebUI/WebCommands.cpp
@@ -14,6 +14,8 @@
 #include "Report.h"  // git_info
 
 #include <Esp.h>
+#include <esp_system.h>
+#include <esp_timer.h>
 
 #include <sstream>
 #include <iomanip>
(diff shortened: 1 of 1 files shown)


MitchBradley:
Did you test with the two WebUIs to see if they are okay with the changes?

aaronse:
> Did you test with the two WebUIs to see if they are okay with the changes?

No.  I only tested with 3.0 WebUI.  I'll test with 2.x as well if this type of edit seems helpful?

If so, I'll attach observations/evidence for both.  I'll follow up in ~wk.

MitchBradley:
Sorry for losing the thread on this.  Please test with webui 2 and FigUI, and if they are okay with it, I will merge this change.

aaronse:
> Sorry for losing the thread on this. Please test with webui 2 and FigUI, and if they are okay with it, I will merge this change.

Same here, i didn't believe my "follow up in ~wk" comment shortly after posting.  I've been working on a fluidnc client that builds on Jason's FluidTouch work.  

I'll follow here if/when I verify these edits with the clients you mentioned.  Soon Hopefully...<sup>TM</sup>

MitchBradley:
Since we merged the portability work into main, the location of these changes is now wrong.  Since they are ESP-specific, they should be in esp32/SysStats.cpp

PR #1742, opened 2026-05-31
Prevent Telnet peer disconnects from triggering the TCP watchdog

figamore:
### Problem
A stalled or dead Telnet peer could block `TelnetClient::write()` indefinitely,
starving the FreeRTOS TCP task watchdog and causing an unintended reboot.

### Changes
- Replaced Telnet's blocking `WiFiClient::write()` with `::send(..., MSG_DONTWAIT)` so writes never hold the CPU.
- Added a line-accumulation buffer so output is flushed one complete line at a time, keeping `\r\n`
normalization correct across fragmented `write()` calls.
- When the client's TCP send buffer is full, the code waits up to `TX_IDLE_WAIT_MS` (50 ms) when the
machine is idle, but drops immediately during motion to keep jogging/running unaffected.
- Clients that stall `TX_STALL_LIMIT` (10) consecutive lines are disconnected and reaped on the next
`poll()`, freeing the slot for a healthy reconnect.

### DIFF
--- a/FluidNC/src/WebUI/TelnetClient.cpp
+++ b/FluidNC/src/WebUI/TelnetClient.cpp
@@ -4,7 +4,11 @@
 #include "TelnetClient.h"
 #include "TelnetServer.h"
 
+#include "../System.h"     // inMotionState()
+
 #include <WiFi.h>
+#include <lwip/sockets.h>  // ::send(), MSG_DONTWAIT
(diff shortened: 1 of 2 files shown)


MitchBradley:
Thanks for finding this.  The fix was rolled into a #1777 which includes some other related fixes.

PR #1748, opened 2026-06-15
Fix TCP watchdog timeout during WebUI WiFi AP scans

figamore:
Requesting a WiFi AP scan from the WebUI or [ESP410] triggers the TCP watchdog before the scan completes. This change pauses the request during the async scan and sends the AP list once the results are ready.

## Steps to reproduce:
1. Send [ESP410] through terminal or request an AP scan in the WebUI's `FluidNC Settings` page by clicking the magnifying glass next to the `Sta/SSID` field.
1. TCP watchdog gets triggered after several seconds and FluidNC restarts.


MitchBradley:
I like what you have done but it will ultimately need to target the Portability branch which segregates everything WiFi related into WifiConfig.cpp and esp32/WifiImpl.cpp in that branch.  WebUIServer is just not the right place for wifi specific details.  If there is a small footprint in WebUIServer, that is okay if it cannot be avoided, but the bulk of the logic needs to go somewhere Wifi-specific, and to the extent that it only works for esp32, should go in WifiImpl

MitchBradley:
Superseded by #1824

PR #1765, opened 2026-07-23
Sensorless homing: per-phase StallGuard threshold (stallguard_seek) and homing_amps for SPI Trinamic drivers

siganberg:
## Summary

Two opt-in homing improvements for the SPI Trinamic drivers (`tmc_2130`, `tmc_5160`), both defaulting to existing behavior when not configured:

1. **`homing_amps`** — reduced motor current during homing, ported from the TMC2209 driver where it already exists. Lower current makes the motor stall sooner and more gently against a hard stop during sensorless homing. Defaults to `run_amps`.

2. **`stallguard_seek`** — a separate StallGuard threshold for the fast (seek) approach phase, while the existing `stallguard` value is used for the slow (feed) approach. The two phases run at very different speeds, and StallGuard sensitivity is strongly speed-dependent, so a single threshold is often a compromise. Defaults to `stallguard`.

To support per-phase retuning, a new no-op `MotorDriver::set_homing_phase()` hook is called (via `Axes`, mirroring the existing `set_homing_mode` plumbing) from `Homing::nextPhase()` whenever the cycle transitions between `FastApproach` and `SlowApproach`. The Trinamic SPI drivers use it to reprogram SGT **and to recompute the TCOOLTHRS/THIGH stall-detection speed window from the rate of the active phase** — previously the window was always derived from the feed rate, which mis-tunes stall detection whenever seek and feed rates differ.

## Example

```yaml
tmc_5160:
  run_amps: 1.8
  homing_amps: 1.0       # gentler stalls while homing
  stallguard: 5          # threshold for the slow confirm pass
  stallguard_seek: 8     # threshold for the fast seek pass
  homing_mode: StallGuard
```

## Tested on hardware

ESP32 (`wifi` env) + 4x BTT TMC5160T Pro (TMC5160A) on a dual-Y-gantry CNC router with sensorless homing (DIAG1 wired to limit inputs). Verified:

- With the new keys absent, homing behavior is unchanged from `main`.
- Two-phase homing works in a single `$H` cycle: e.g. seek 1000 mm/min @ `stallguard_seek` vs feed 200 mm/min @ `stallguard`, with the speed window following each phase.
- `homing_amps` reduces current during homing and restores `run_amps` afterwards.

## Notes for reviewers / docs

- **Speed-window correction changes seek-phase behavior for existing sensorless users who run seek != feed**: previously the feed-tuned window left StallGuard partially gated during a faster seek; with the corrected window, seek-phase detection is genuinely live and thresholds may need retuning (typically to a less-sensitive value). Users running the recommended seek == feed are unaffected.
- With `homing_runs: 1`, the single approach is a `FastApproach`, so it uses `stallguard_seek`. Worth a docs mention.
- Sensorless bounce homing (`homing_runs: 2`) may need a larger `feed_scaler` (~2+): the confirm pass's travel budget is `pulloff x feed_scaler`, and StallGuard needs a few mm of commanded "grind" distance past contact before it asserts.
- `tmc_5160_pro` inherits the new config keys but its raw-register `set_registers()` path is intentionally untouched. The TMC2209 (`SGTHRS`) could receive the same `stallguard_seek` treatment in a follow-up.


MitchBradley:
I updated this PR to conform to the current code, by adding machine-readable documentation for the new config variables.  Please check to see if it still does what you want.

PR #1766, opened 2026-07-24
USB Pendrive Support added. Complies ok esp32s3 wifi

khatrihamid:
I have tried to add USB Pendrive Read /Write And Play (G code ) file support for ESP32-S3.
as this Esp Board supports it. 
as of now i dont have this board in my hand ,So i cant verify this with actual Hardware.

I will be thankfull if someone does so and update results.

As Things are upgrading, I am using FluidNc For my own project and find many issues.
like old esp32 wroom boards comes with microUsb, then for standard esp32 + CNC driver Boards are not available for Individual At cheaper price, SO i am fixing things 1 by 1. 

as of now , I have upgraded fluidnc, in featyre, i will design 1 shield for esp32 for cnc


bdring:
I tested on a Doberman controller. The PR compiles and loads fine for me. I used this config section, which parsed without errors.

```yaml
usbdrive:
  enabled: true
```

There was no message in the startup messages. I think there should be some indication that it is active.

I have no idea how to use it. Looking at the code it appears the intent was to support "localfs, sd & usbdrive", but commands like `$usbdrive/list` are not valid.

I think we should close this PR until you test and validate the code. You should be able to buy an ESP32-S3 dev kit with a native USB for less than $10.

PR #1779, opened 2026-08-08
Fix Telnet output after Portability merge

figamore:
Two issues were introduced in `TelnetClient` from the Portability branch that prevented Telnet from sending responses:

- The `connected()` check was inverted, causing writes to return early for connected clients.
- `availableForWrite()` always returns `0` for ESP32 `WiFiClient`, so the transmit queue never drained.

ESP32 now uses non-blocking socket writes again. The existing Arduino path remains in place for RP2040/RP2350.

This fixes G-code senders and FluidDial repeatedly disconnecting when using Telnet.

### DIFF
--- a/FluidNC/src/WebUI/TelnetClient.cpp
+++ b/FluidNC/src/WebUI/TelnetClient.cpp
@@ -7,6 +7,11 @@
 
 #include <WiFi.h>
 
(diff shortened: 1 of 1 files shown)


MitchBradley:
Duplicate of #1777

PR #1796, opened 2026-08-21
Draft: harden ESP32 WebUI/AsyncTCP connection lifecycle

WaldemarFech:
## Summary

This draft proposes the R22 WebUI/AsyncTCP hardening work from a real ESP32
failure investigation. Repeated WebUI/WebSocket activity could exhaust or
fragment the heap until FluidNC stopped responding during a machining
operation. That creates a safety concern because the controller can no longer
service the control path normally.

The public fork branch is:

- `WaldemarFech:debug/r22-webhardening`
- source commit: `ce05a763b5235f24e8d05abfbe95124623215515`
- BIN: [v4.0.4-webhardening-20260821-r22](https://github.com/WaldemarFech/FluidNC/releases/tag/v4.0.4-webhardening-20260821-r22)

## Scope

- bounded WebUI/WebSocket admission and cleanup;
- allocation-safe request/file/stream guards;
- stale-client and early-RST ownership cleanup;
- AsyncTCP accept-event cleanup and diagnostics;
- Channel semaphore lifetime/unwind repair;
- bounded OTA/error handling and recovery telemetry;
- host regression tests and a read-only WebUI soak harness.

New motion/recovery capabilities remain disabled in this debug branch. No
generic G-code relay or automatic resume is introduced here.

## Evidence

- 100 parallel bursts of 8 connection attempts passed;
- complete 9-stage read-only WebUI lifecycle smoke passed;
- static pressure rejected excess requests deliberately and recovered;
- no reboot beyond the deliberate OTA update in the controlled run;
- configuration/runtime/journal/backtrace hashes remained unchanged;
- BIN SHA-256:
  `C9CA9DCA4AD809DCE473601C1110D113471181A274654E50A3CF9C3BC3A16392`.

The current branch is intentionally a draft because the source diff is broad
and should be split/reviewed into upstream-appropriate units. The connection
admission policy is bounded and fail-closed; it is not presented as one
universal safe connection count for every board or workload.

## Review request

Please review the ownership/exception-safety changes first, especially the
AsyncTCP accept/RST paths and WebUI allocation guards. If the direction is
accepted, I can follow up with smaller production commits and keep diagnostic
instrumentation separate.


MitchBradley:
This PR touches so many files, and mixes so many concerns, that it is effective impossible to digest.  You asked for a review of the ownership/exception safety parts, but finding those amid all the other stuff would be difficult.  We have done a lot of work recently to harden against networking-induced crashes.  If you think, or can show with your test scripts, that there is still more to do, please provide one or more targeted PRs that are small enough for us to comprehend.

MitchBradley:
Thanks for this @WaldemarFech, and for calling out the ownership /
exception-safety changes as the part to look at first. Some notes on what
came of that review.

**Why the diff is hard to act on as-is.** This branch is a single squashed
commit on a base from around July 25, and it bundles several unrelated
bodies of work along with the WebUI hardening: `@config` doc comments
across ~50 files, a Python soak/diagnostics harness, the `DebugRecovery/`
subsystem, and an output-URL HTTP feature. Against current `main` a large
fraction of the remainder is also already handled by separately merged
changes:

- `Channel::_queue_mutex` creation/teardown + real destructor (#1806 /
  #1807), plus bounding of `_queue` growth
- `WSChannel` send path surviving a gone client / a throwing
  `_server->binary()`, and output-frame coalescing to cut oversized-pbuf
  pressure
- `WebClient` ctor throwing `std::bad_alloc` on mutex/queue allocation
  failure, with cleanup
- `begin_closing()` / `flushRx()` / `active(false)` on disconnect to kill
  the stale-buffered-command race

**What was genuinely not on `main`**, extracted as three small PRs rebased
onto current `main` (mutually independent, mergeable in any order):

- #1812 &mdash; make `WS_EVT_CONNECT` exception-safe: own the new
  `WSChannel` in a `unique_ptr` and unwind cleanly if `registration()` or
  the list `push_back` throws, instead of leaking it and leaving a
  dangling `_wsChannels` entry.
- #1813 &mdash; `AllChannels::kill()` returns whether the channel was
  queued (was `void`, silently dropped on a full `_killQueue`); queue
  depth 16&rarr;32; telnet retries on transient failure.
- #1814 &mdash; reap orphaned `WSChannel`s whose socket died without a
  `WS_EVT_DISCONNECT` (RST/no-FIN), gated on the client actually being
  gone so a live idle client is never touched.

**Why the `bad_alloc`-tolerant `client->close()` + deferred session-close
retry was not extracted.** Against the stock ESPAsyncWebServer,
`AsyncWebSocketClient::close()` does not throw &mdash; it `malloc`s the
close-frame control block and degrades quietly on failure. The
`catch (std::bad_alloc&)` in this branch only does anything with the
patched library produced by `tools/patch_espasyncwebserver_close.py`, so
upstream it would be dead code plus an unreachable deferred-close array.
That one needs an ESPAsyncWebServer change first; happy to look at it as
its own thing if you want to pursue it.

The heap-headroom admission system (`WebResourcePolicy.h`,
`try_reserve_*`), `DebugRecovery/`, and the output-URL feature are
separate discussions and aren't in any of the three PRs.

MitchBradley:
The three split-out PRs, for reference:

- https://github.com/bdring/FluidNC/pull/1812 &mdash; WebUI: make WS_EVT_CONNECT exception-safe
- https://github.com/bdring/FluidNC/pull/1813 &mdash; channels: AllChannels::kill() reports queue-full failure
- https://github.com/bdring/FluidNC/pull/1814 &mdash; WebUI: reap orphaned WebSocket channels

MitchBradley:
Addressed by separate PRs listed above

PR #1805, opened 2026-08-28
Add TS35 touchscreen display/pendant module

thcasssio:
## Hardware

An **MKS TS35-R V2.0**: a 3.5" 480x320 TFT with an ST7796 controller and an ADS7843E resistive touch controller, both on one SPI bus behind separate chip selects. The panel has no processor of its own. It plugs into the EXP1/EXP2 headers of an **MKS DLC32 V2.1 (ESP32-D0WD)**.

**That is the only combination this has run on.** No other board, no other panel, no ESP32-S3. The ST7796 init sequence comes from the public MKS/TFT_eSPI sources for this panel; other ST7796 modules may well need different gamma or inversion settings, and I have no way to check.

## How it fits

It is a `Channel`, like the OLED for output and like the UART pendant and the ESP-NOW remote for input — the shape the config docs already describe when they talk about a display/pendant that reads reports and sends commands back.

## Pages

A tab bar along the bottom switches between five pages. The header shows the machine state, coloured by state, plus the active WCS and a READ ONLY marker when the controls are off. A one-line notice strip above the tab bar carries transient messages.

- **DRO** — X/Y/Z work position in large digits, and a footer with units, feed, spindle RPM and the raw `Pn:` field.
- **JOG** — six axis buttons (X/Y/Z, minus and plus), the live position between them, and a bottom row of STEP / CANCEL / SPEED. STEP cycles 0.01 → 0.1 → 1 → 10 → 100 mm, truncated by `jog_step_max_mm`; SPEED cycles 10/25/50/80/100 % of the configured jog feeds. Z has its own cap (`jog_step_max_z_mm`) and the page says `Z CAPPED AT ... MM` when the selected step is larger than it. Jogs go out as `$J=G91 G21 ...`, always in millimetres regardless of the active G20/G21. CANCEL sends jog cancel (0x85), and sits in the middle of the row because it is the button reached for in a hurry.
- **ZERO** — ZERO X / Y / Z / XYZ, sent as `G10 L20 P0`, so it lands on the active coordinate system and shows up on the other channels at the next report.
- **CONTROL** — HOLD, RESUME, CANCEL JOG; a row of feed and rapid overrides (F-10, F100, F+10, R100), which change a cut that is already running; and HOME, UNLOCK, RESET. The page also displays the override percentages the controller reports.
- **INFO** — mapped and raw touch coordinates, the calibration currently in effect, the modal state, buffer/line/SD counters, and whether the controls are enabled. This is the page used to calibrate the panel.

An alarm banner appears over the middle of any page while the machine is in Alarm, showing the alarm number, a short hint for the common codes, and an UNLOCK button that sends a soft reset and then `$X`.

## Safety gating

`controls_enabled` is false by default and gates everything that moves the machine, because an uncalibrated resistive panel puts taps on the wrong button. Under it are six `allow_*` items: `allow_jog` and `allow_zero` default true; `allow_homing`, `allow_unlock` and `allow_soft_reset` default false. `allow_alarm_clear` is deliberately independent of `controls_enabled`, since neither the soft reset nor the `$X` produces motion, so a machine can be recovered from an alarm while the screen is otherwise read-only.

Zero, home, unlock and soft reset each need a second tap inside a three-second window; the notice line reads `TOUCH AGAIN: ...` while it waits.

## Integration details

- The panel uses the machine's shared `spi:` bus. `Machine::SPIBus` opens it long before modules are initialised; the module adds an LCD device and a touch device to that bus and logs an error and stops if there is no `spi:` section.
- The five GPIOs are `Pin` objects defaulting to `NO_PIN`, so a collision with a step, dir or limit pin is reported as a configuration error instead of being driven quietly.
- A full-screen wipe is spread over consecutive polls, one 32-row band at a time, so switching pages does not hold the shared channel polling loop for the length of a 307200-byte transfer.
- `docs/config_items.yaml` is regenerated for the 35 new items, and `tools/build_config_docs.py` gets the matching `ts35` entry in its `SECTIONS` table. `--fail-on-drift` passes.

## Minimum config

The pin numbers below are **an example** — they are the DLC32 V2.1 EXP1/EXP2 wiring. On any other board they will be different.

```yaml
spi:
  miso_pin: gpio.19
  mosi_pin: gpio.23
  sck_pin: gpio.18

ts35:
  tft_cs_pin: gpio.25
  tft_dc_pin: gpio.33
  touch_cs_pin: gpio.26
  tft_reset_pin: gpio.27
  backlight_pin: gpio.5
  controls_enabled: true
```

`tft_reset_pin` and `backlight_pin` are optional. `backlight_active_low` defaults to true, which is right for this panel; MKS sources disagree across revisions, so it is a setting rather than a constant.

Bring it up with `controls_enabled` left at its default first, read the raw touch values off the INFO page while touching the panel edges, put them in `touch_x_min` / `touch_x_max` / `touch_y_min` / `touch_y_max`, and only then turn the controls on.

## What is not here

- **No automated tests.** There is no harness in the repo for a display module: `[tests_common].build_src_filter` lists files under `FluidNC/src` one by one, and nothing under `FluidNC/esp32` is in it. `TS35Model` is deliberately free of Arduino, ESP-IDF and FluidNC dependencies, so it is the piece that could be tested on a host, but wiring that up is a separate change and I have not made it.
- **No board other than the DLC32 V2.1 tested**, and no panel other than the TS35-R V2.0. Nothing excludes the files from the ESP32-S3 build, but I have not run it there.
- On the ESP32 an SPI host has three device slots. The SD card takes one and this takes two, so those three are used up on a board that has both.
- Screen geometry is fixed at 480x320, which is what this module is.

## Out of scope

Two things running on my machine were left out of this PR: a page that runs a job from the screen, including probe routines, and a buzzer driven from the same panel header. Both are further from the "display and pendant" job than the rest, so they are not here. If either is of interest they can come as separate PRs.

### DIFF
--- a/FluidNC/docs/config_items.yaml
+++ b/FluidNC/docs/config_items.yaml
@@ -740,6 +740,294 @@ oled:
     description: |
       Delay, in milliseconds, before the display shows radio (WiFi/BT) connection info after boot -- gives the radio hardware time to come up first.
 
+# Lives under FluidNC/esp32/ for the same reason as oled just above -- the relative path deliberately escapes SRC to reach it.
+ts35:
+  report_interval_ms:
+    type: integer
+    min: 100
+    max: 5000
+    default: '200'
+    tuning: typical
+    description: |
+      How often the module asks the machine for a status report. This is the channel's own report interval, independent of what other channels use.
+
+  render_interval_ms:
+    type: integer
+    min: 100
+    max: 2000
+    default: '200'
+    tuning: typical
+    description: |
+      Lower bound between screen repaints. Only fields whose text actually changed are redrawn, so raising this costs responsiveness, not bandwidth.
+
+  touch_interval_ms:
+    type: integer
+    min: 20
+    max: 500
+    default: '40'
+    tuning: typical
+    description: |
+      How often the touch controller is sampled. Below about 20 ms the SPI traffic starts competing with the display writes for the same bus.
+
+  controls_enabled:
+    type: boolean
+    default: 'false'
+    tuning: per-machine
+    description: |
+      Master gate for every control that moves the machine or changes its coordinates. Leave it false until the touch calibration below has been verified on the bench: an uncalibrated panel will send taps to the wrong button.
+
+  allow_jog:
+    type: boolean
+    default: 'true'
+    tuning: per-machine
+    description: |
+      Allows the jog buttons, subject to controls_enabled.
+
+  allow_zero:
+    type: boolean
+    default: 'true'
+    tuning: per-machine
+    description: |
+      Allows setting work zero from the screen, subject to controls_enabled. Zeroing uses G10 L20 P0, so it updates the active coordinate system and shows up on the other channels at the next report.
+
+  allow_homing:
+    type: boolean
+    default: 'false'
+    tuning: per-machine
+    description: |
+      Allows starting a homing cycle from the screen.
+
+  allow_unlock:
+    type: boolean
+    default: 'false'
+    tuning: per-machine
+    description: |
+      Allows $X from the Control page. Not needed when allow_alarm_clear is true, which offers the same thing where the alarm is actually shown.
+
+  allow_soft_reset:
+    type: boolean
+    default: 'false'
+    tuning: per-machine
+    description: |
+      Allows sending a soft reset from the Control page.
+
+  allow_alarm_clear:
+    type: boolean
+    default: 'false'
+    tuning: per-machine
+    description: |
+      Shows an alarm banner with an unlock button on every page whenever the machine is in Alarm. The button sends a soft reset and then $X. This is deliberately independent of controls_enabled, because neither command produces motion, so a machine can be recovered from an alarm while the screen is otherwise read-only.
+
+  jog_step_mm:
+    type: float
+    min: 0.001f
+    max: 100.0f
+    default: '0.100'
+    tuning: per-machine
+    description: |
+      Starting jog increment. The STEP button cycles it through 0.01, 0.1, 1, 10 and 100 mm, capped as described under jog_step_max_mm; this value only picks where that ladder starts. The screen's own selection is not written back here, so $CD still reports what the config file asked for.
+
+  jog_feed_xy_mm_min:
+    type: float
+    min: 1.0f
+    max: 100000.0f
+    default: '1000.000'
+    tuning: per-machine
+    description: |
+      Feed rate used for jogs on X and Y.
+
+  jog_feed_z_mm_min:
+    type: float
+    min: 1.0f
+    max: 100000.0f
+    default: '500.000'
+    tuning: per-machine
+    description: |
+      Feed rate used for jogs on Z.
+
+  jog_step_max_mm:
+    type: float
+    min: 0.01f
+    max: 100.0f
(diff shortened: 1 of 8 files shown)


MitchBradley:
Are you proposing that we add and support code that is usable only on one third-party board that already costs us a lot to support?

bdring:
We only want support displays with integrated controllers using UART channels or wireless connections. This means we can support many displays without loading up the firmware with extra code. We compile all features and enable things via config files so all code gets compiled.

MitchBradley:
One alternative would be for you to publish and support your own DLC32-specific version.  We could work out some way to let WebInstaller install it, if you are interested.

thcasssio:
Fair points, both — I hadn't considered that every feature compiles into every build; with the DLC32 builds already near the partition limit, that concern is very concrete.

@MitchBradley I'm interested in the WebInstaller route. The DLC32 is commonly sold bundled with the TS35, so a maintained DLC32-specific build would serve that exact user base without costing mainline anything. I'd maintain it in my fork, tracking upstream releases. What would you need from my side to wire it into WebInstaller?

Happy to close this PR in favor of that path.

MitchBradley:
WebInstaller fetches releases from github.com/bdring/fluidnc-releases.  The release action copies it there along with a manifest describing what can be installed to which processor.  All that is automated via a release workflow.

I think what we would have to do is have webinstaller consult an additional site to find other releases.  I will look into that.

The release artifacts cannot be served from the main repo's release area because CORS.

thcasssio:
On the WebInstaller side — I had a look at `breiler/fluid-installer` before asking you to spend time on it, and the plumbing may already be there. `GitHubService` takes the source as constructor arguments:

```ts
constructor(apiURL?: string, resourcesURL?: string)
```

The `FLUIDNC_*` constants are only the defaults, and the FluidDial pages already drive the same class with their own pair. So fetching releases, reading the manifest and flashing all look source-agnostic already — what's missing is a way to point it somewhere else, not new machinery.

That leaves the parts that are yours to decide rather than mine: whether an additional source is a fixed list or free-form entry, and how a community build gets labelled so nobody mistakes it for an official one. Happy to write it to whatever shape you and @breiler settle on.

Either way I'll publish in the `fluidnc-releases` layout — files committed, served from `raw.githubusercontent.com`, manifest in your format — so there's a real source to test against when you get to it. Thanks for the CORS pointer; that explains the separate repo, which I'd been wondering about.

MitchBradley:
Yeah, I figured that the hard part would not be the plumbing - because @breiler and I already spent a lot of time making that generic - but rather the user presentation at the top level.  I wonder if this should look like another top level choice so FluidNC / FluidDial / CommunityFluidNC ?  My original thought was having it under FluidNC, merging the release list to include releases from both sources, but I see some possibilities for confusion around that.

Which brings up another point - I am thinking of having two release artifacts for S3.  One for DLC32-Max and one for other S3 controllers with PSRAM.  RAM limits have been a gating concern for some time now, and having a PSRAM-enabled build would make that perennial problem just go away.  All of Bart's S3 controllers have PSRAM, but it is currently unused because we have the one build that supports DLC32-Max too.  My life would be better if someone else were the gatekeeper for all things DLC32.  Interested?

PR #1811, opened 2026-08-31
Run informational commands during a job: two-task command dispatch

MitchBradley:
## What this does

Lets a second channel (a pendant, a WebUI/telnet console) run **informational commands while a job is streaming**, without stalling the job and without the OOM that motivated the work. The core change is a two‑task split of command dispatch: the decision of *what a line is and where it runs* now happens on the polling task, and only lines that genuinely need the protocol/planner context cross to `protocol_main_loop`.

## Background

An M5Dial / FluidDial pendant on a UART channel drove FluidNC to an out‑of‑memory abort during SD jobs:

* FluidDial's `service_config_requests()` retries an unanswered `$/...` config query every 500 ms, indefinitely.
* During a job, `polling_loop` only line‑polls `Job::channel()`. Every non‑job channel's `Channel::_queue` (a `std::deque<uint8_t>`) accumulates the retry traffic and is never drained → ~2 KB/min of heap loss → `std::bad_alloc` → terminate.

Bounding `_queue` stops the crash, but the deeper issue is that `protocol_main_loop` is the **single consumer** of the input‑line handoff and it blocks inside `mc_line`'s `plan_check_full_buffer()` spin while the planner is full. Anything behind it in the handoff — a status query, a config probe — waited *seconds* during a job (`$G` from the console took 12 s in testing). No polling‑schedule tweak fixes that; the consumer itself is the bottleneck.

## The change, in pieces

| commit | what |
|---|---|
| **Use atomic operations to make the GCode line handoff safe** | `activeChannel`/`activeLine` were shared unsynchronized between the polling task (core 0) and `protocol_main_loop` (core 1). Make the handoff `std::atomic` with release/acquire. |
| **Bound `Channel::_queue`** | Cap at `4*maxLine`. When full at a line boundary, discard the whole incoming line through its newline (a line already in progress finishes) — the queue never holds a partial line. One `log_debug` per overflow episode. **This is the actual OOM fix.** |
| **Lock the Job stack against cross‑core mutation** | `job` is pushed by `nest()` on the protocol task and popped by `unnest()`/`abort()` on the polling task with no synchronization. Add an internal mutex; `channel()`/`source()` return `nullptr` when idle instead of dereferencing `back()` on an empty vector; `leader_channel()` for a locked read. |
| **Split the command "synchronous" flag into context + buffer‑drain** | `Command::synchronous()` conflated "must run on the protocol task" with "drain the planner first" — they differ for `$J`. Replace with `needs_protocol_context()` and `drains_buffer()`. New `ReportCommand` for pure observers (`$G $$ $S $T $CD $#` …), so they no longer needlessly wait for buffered motion. Also: run the per‑command state guard **before** the buffer sync (`Command::disallowed()`), so a command that's going to be rejected doesn't drain the planner first. |
| **Make `$X` a no‑op outside Alarm state** | `disable_alarm_lock()` ran the `_after_unlock` macro unconditionally; issued during a job that nested a macro job into the middle of the program. |
| **Two‑task command dispatch** | The single‑slot handoff becomes a `cmd_queue` (`LineItem{channel,line}`, depth 1, same flow control). `execute_line()` gains `on_protocol_task`; `line_needs_protocol_context()` classifies a line once against `Command::List`. On the polling task: an interloper's line that needs context is rejected (`Error::AnotherInterfaceBusy`), one that doesn't runs inline via `run_command_inline()`, everything else defers to `cmd_queue` (`Error::Deferred`, new). `polling_loop` now also services non‑job channels every pass during a job, so their reads execute in ~1 ms regardless of planner depth. Also fixes a job‑channel `processing_ref` underflow. |
| **Allow more informational commands from run state** | Loosen `$$ $L $S $SC $CMD $# $I $GS` to `anyState` and move `$MU0/$MU1/$MC`, `$LM..$LV`, `$SA` to `ReportCommand` — all deliberately fine from an interloper. `$V` (NVS/FLASH contention) and `$Limits` (spins until feedhold) stay restricted. |
| **Classify WebCommands** | `WebCommand` now defaults `needs_protocol_context() == true` (filesystem, NVS, network, restart). New `WebReportCommand` for the read‑only status set: `ESP420`, `ESP400`, `ESP0`/`ESP`, `ESP800`, `ESP111`, `wifi/status`, `Ethernet/Status`. `line_needs_protocol_context()` looks `[ESPxxx]` up by name. |

## Runtime behavior

| line source | needs protocol context (gcode, `$H`, `$J`, writes, FS/NVS…) | no context (`$G`, `$$`, `$/axes/...`, `ESP420`, …) |
|---|---|---|
| the job's own channel | protocol task, ordered | protocol task, ordered |
| a different channel while a job runs | `error:120` | runs inline on the polling task, ~1 ms |
| no job running | protocol task (unchanged) | protocol task (unchanged) |

The per‑command state guard (`_cmdChecker`) still applies as a second layer everywhere.

## Testing

Bench + hardware (ESP32‑S3, WiFi, SD job, FluidDial pendant on UART):

* No `Channel::_queue` drops during a job — the pendant's `$/...` probes are answered on the next `polling_loop` pass instead of after the move buffer drains.
* Console informational commands, including a full `$CD` dump, stay responsive mid‑job.
* Builds clean for `wifi_s3` and `wifi`.

## Known limitations / follow‑ups

* `run_command_inline()` runs on the polling task and must not block — a `ReportCommand` streaming to a back‑pressured `WSChannel` could stall realtime‑char handling. Not observed in practice; a dedicated command task or an interloper reject for large dumps would close it.
* `polling_loop` now calls `pollChannels(buf)` every pass during a job (~1 ms) — cheap, but a throttle counter is a candidate if overhead shows.
* Config‑tree reads on the polling task still race a job‑file `$/x=NNN` write on the protocol task — accepted for numeric scalars (torn read = old‑or‑new), a `config_mutex` on just the command paths would remove it.
* `[ESP…]` from the WebUI console still runs synchronously on the webclient task (unchanged); only `[ESP…]` over serial/telnet/pendant goes through the new path.

## Behavior changes reviewers should note

* `$$ $L $S $SC $CMD $# $I $GS` now work from any channel in `Cycle`/`Hold` (were `IdleError`).
* `$X` from `Idle` no longer runs the `_after_unlock` macro.
* `ReportCommand`s no longer call `protocol_buffer_synchronize()` before running (they never needed to).
* Pasting g‑code into a non‑job channel during a job now returns `error:120` per line instead of silently buffering it.

### DIFF
--- a/FluidNC/src/Channel.cpp
+++ b/FluidNC/src/Channel.cpp
@@ -127,8 +127,13 @@ void Channel::queue_push(uint8_t byte) {
             _queue_at_line_start = true;
         }
     } else if (_queue_at_line_start && _queue.size() >= _queue_limit) {
-        _queue_discarding      = true;
-        _queue_at_line_start   = false;
+        // No room for another line.  A newline here is an empty line - drop it
+        // and stay at a line boundary so the next line gets a fresh check;
+        // otherwise discard the whole incoming line through its newline.
+        if (!is_newline) {
+            _queue_discarding    = true;
+            _queue_at_line_start = false;
+        }
         log_now                = !_queue_overflow_logged;
         _queue_overflow_logged = true;
     } else {
--- a/FluidNC/src/Channel.cpp
+++ b/FluidNC/src/Channel.cpp
@@ -106,7 +106,40 @@ void Channel::flushRx() {
     while (_queue.size()) {
         _queue.pop();
     }
+    _queue_at_line_start   = true;
+    _queue_discarding      = false;
+    _queue_overflow_logged = false;
     xSemaphoreGive(_queue_mutex);
+}
+
+// Enqueue one non-realtime byte.  Drop policy is whole-line: if a new line
+// starts while _queue is already at _queue_limit, the whole line is discarded
+// through its newline.  A line already in progress finishes, so _queue never
+// holds a partial line.  Worst-case size is _queue_limit + one maxLine line.
+void Channel::queue_push(uint8_t byte) {
+    bool is_newline = byte == '\n' || byte == '\r';
+    bool log_now    = false;
+
+    xSemaphoreTake(_queue_mutex, portMAX_DELAY);
+    if (_queue_discarding) {
+        if (is_newline) {
+            _queue_discarding    = false;  // over-limit line consumed; re-check at the next one
+            _queue_at_line_start = true;
+        }
+    } else if (_queue_at_line_start && _queue.size() >= _queue_limit) {
+        _queue_discarding      = true;
+        _queue_at_line_start   = false;
+        log_now                = !_queue_overflow_logged;
+        _queue_overflow_logged = true;
+    } else {
+        _queue.push(byte);
+        _queue_at_line_start = is_newline;
+    }
+    xSemaphoreGive(_queue_mutex);
+
+    if (log_now) {
+        log_debug("Input queue full on " << _name << "; dropping lines");
+    }
 }
 
 size_t Channel::queued_bytes() const {
@@ -124,6 +157,12 @@ bool Channel::try_pop_queued_byte(uint8_t& byte) {
     }
     byte = _queue.front();
     _queue.pop();
+    if (_queue.empty()) {
+        // Fully drained: reset framing state and re-arm the overflow warning.
+        _queue_at_line_start   = true;
+        _queue_discarding      = false;
+        _queue_overflow_logged = false;
+    }
     xSemaphoreGive(_queue_mutex);
     return true;
 }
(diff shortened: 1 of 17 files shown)


MitchBradley:
Thanks — all three are fair.

1. **Overflow + newline** — real edge case. When the byte tripping the bound is itself a newline (blank line, or a `\r\n` split across two `queue_push` calls) we armed `_queue_discarding` and ate the following line. Fixed: a newline in the overflow branch is just dropped, leaving the queue at a line boundary for a fresh check.
2. **`execute_line` double-lock** — agreed; also inconsistent with the equivalent check in `protocol_main_loop`, which already does the single-call form. Now `Channel* jc = Job::channel(); if (jc && &channel != jc)`.
3. **`ReportCommand` doc** — correct, "pure observer" stopped being accurate once `$RI`/`$SA`/`$MU*`/`$LM..$LV` moved to it. Reworded to the real contract.

Pushed as 520911a3.

PR #1816, opened 2026-09-03
ESP32: Fix guard disabling the task-watchdog layer

f1adang:
### The problem

Every function in `FluidNC/esp32/wdt.cpp` is guarded by `CONFIG_ESP_TASK_WDT_EN`.
That is the ESP-IDF **v5** name for the setting. On IDF 4.x, which the current
`framework-arduinoespressif32` package uses, the symbol is `CONFIG_ESP_TASK_WDT`,
so none of the guards are satisfied and every function compiles to an empty body.

From the linked image, before this change:

```
feed_watchdog:         entry / retw     ← feeds nothing
add_watchdog_to_task:  entry / retw     ← subscribes nothing
enable_core0_WDT:      entry / retw
disable_core0_WDT:     entry / retw
```

The watchdog itself is very much enabled — the same sdkconfig has
`CONFIG_ESP_TASK_WDT_TIMEOUT_S 5` and `CONFIG_ESP_TASK_WDT_PANIC` — so the timer
runs while FluidNC's only means of answering it does nothing. A task that
subscribes itself, such as `async_tcp`, still trips it and panics the board, and
the `feed_watchdog()` calls placed around the code to prevent exactly that have no
effect.

I ran into this chasing repeated `Task watchdog got triggered … async_tcp`
reboots during SD card access. Several rounds of adding `feed_watchdog()` calls
changed nothing, because none of them were compiled in.

### The change

Accept either spelling. One `#if defined(A) || defined(B)` and the four guards
renamed to the internal macro.

After, from the same image:

```
feed_watchdog:
    entry
    call8  esp_task_wdt_status
    call8  esp_task_wdt_reset
    retw
```

### Worth reviewing carefully

This turns supervision back on rather than adding anything new.
`add_watchdog_to_task()` is called by the polling and protocol tasks, which have
therefore not actually been supervised on IDF 4.x. They will be after this.

Both feed regularly — `polling_loop()` in its main loop, `protocol_main_loop()`
via `protocol_handle_events()` — so this should be correct. But any path in those
tasks that blocks longer than the five second timeout will now be caught where it
previously passed unnoticed, and that is a behaviour change on hardware I cannot
survey. I would rather flag it than have it surface as a surprise.

Builds on `wifi`, `noradio` and `bt`; unit tests pass.

### DIFF
--- a/FluidNC/esp32/wdt.cpp
+++ b/FluidNC/esp32/wdt.cpp
@@ -7,6 +7,15 @@
 #include "Config.h"
 #include <esp_idf_version.h>
 
+// ESP-IDF v5 names this CONFIG_ESP_TASK_WDT_EN; v4.x names it
+// CONFIG_ESP_TASK_WDT.  Testing only the v5 name leaves every function in this
+// file compiled to an empty body on v4.x, so feed_watchdog() feeds nothing and
+// add_watchdog_to_task() subscribes nothing, while the task watchdog itself is
+// enabled and will panic on expiry.  Accept either name.
+#if defined(CONFIG_ESP_TASK_WDT_EN) || defined(CONFIG_ESP_TASK_WDT)
+#    define FLUIDNC_TASK_WDT_ENABLED 1
+#endif
+
 static TaskHandle_t wdt_task_handle = nullptr;
 
 static void get_wdt_task_handle() {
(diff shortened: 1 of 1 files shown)


f1adang:
Pure vibe contribution.
Found while debugging some weird SD card behaviour on a [Brušograf](https://wiki.sgmk-ssam.ch/wiki/Brushograph).
Code and PR written by Claude Opus 4.8.

PR #1817, opened 2026-09-04
esp32: keep state_is() and the spindle ISR dispatch out of flash

f1adang:
> _Found and written by **Claude** (Anthropic's AI assistant), working on a fork at the owner's direction. Flagging that up front so this gets human scrutiny rather than being taken on trust — I have said explicitly below what was observed on hardware and what is reasoning from the code._

### The step ISR reads vtables out of flash

Running a job produced:

```
Guru Meditation Error: Core 1 panic'ed (Cache disabled but cached memory region accessed)
PC : 0x40082565  ->  Machine::Stepping::step()   Stepping.cpp
A0 : 0x400822dc  ->  Stepper::pulse_func()       Stepper.cpp
```

Any SPI flash access disables the cache on both cores. An ISR that touches flash in that window panics — which is why `$HTTP/BlockDuringMotion` exists.

**Two independent causes, both fixed here.**

**1. `platformio_override.ini` was disabling the vtable-in-DRAM mechanism.** `extra_scripts` *replaces* the value inherited through `extends` rather than adding to it, so setting it for the wifi/noradio/bt environments silently discarded `FluidNC/ld/esp32/vtable_in_dram.py` from `[common_esp32]`. Without it the project's own `sections.ld` is never on the linker's search path, `vtable_in_dram.ld` never applies, and every vtable stays in flash:

```
before:  vtables in DRAM   0,  in flash 240
after:   vtables in DRAM  30,  in flash 210
Pins::GPIOPinDetail   3f40eca8 (flash)  ->  3ffc4f50 (DRAM)
Spindles::Null        3f4.....  (flash)  ->  3ffc5618 (DRAM)
```

Those are exactly the two sets `vtable_in_dram.ld` names — and both are read by the step ISR: `Pin::write()` dispatches through `PinDetail`, and `Stepper::pulse_func()` calls `spindle->setSpeedfromISR()`. The protection was already designed; it had simply been switched off by accident.

**2. Two remaining flash accesses the linker script does not cover.** `state_is()` was a plain call into flash from the ISR — now `IRAM_ATTR`. `spindle->setSpeedfromISR()` read the Spindle vtable — the active spindle now supplies a plain function pointer, resolved once when the spindle changes and called directly. Every `setSpeedfromISR` implementation was already in IRAM; only the dispatch was not.

### Verification

Disassembled the linked image: `Stepper::pulse_func()`, `Machine::Stepping::step()` and `Stepping::unstep()` contain **no literal referencing flash and no calls into flash**. Builds on wifi, noradio, bt and rp2040_wifi; unit tests pass.

The panic above was observed on hardware and its PC decoded to the call site named. I have not reproduced the fixed build under the same conditions for long enough to call it proven in the field.

### DIFF
--- a/FluidNC/src/Spindles/10vSpindle.cpp
+++ b/FluidNC/src/Spindles/10vSpindle.cpp
@@ -98,4 +98,14 @@ namespace Spindles {
     namespace {
         SpindleFactory::InstanceBuilder<_10v> registration("10V");
     }
+
+    // Called from the step ISR in place of a virtual dispatch, which would have
+    // to read this class's vtable out of flash.  IRAM_ATTR, and the qualified
+    // call keeps it non-virtual.
+    static void IRAM_ATTR _10v_speed_thunk(Spindle* s, uint32_t dev_speed) {
+        static_cast<_10v*>(s)->_10v::setSpeedfromISR(dev_speed);
+    }
+    Spindle::IsrSpeedFn _10v::isr_speed_fn() {
+        return _10v_speed_thunk;
+    }
 }
--- a/FluidNC/src/Spindles/10vSpindle.h
+++ b/FluidNC/src/Spindles/10vSpindle.h
@@ -29,6 +29,7 @@ namespace Spindles {
         void init() override;
         void config_message() override;
         void setSpeedfromISR(uint32_t dev_speed) override;
+        IsrSpeedFn isr_speed_fn() override;
 
         void deinit() override;
 
--- a/FluidNC/src/Spindles/DacSpindle.cpp
+++ b/FluidNC/src/Spindles/DacSpindle.cpp
@@ -65,5 +65,15 @@ namespace Spindles {
     namespace {
         SpindleFactory::InstanceBuilder<Dac> registration("DAC");
     }
+
+    // Called from the step ISR in place of a virtual dispatch, which would have
+    // to read this class's vtable out of flash.  IRAM_ATTR, and the qualified
+    // call keeps it non-virtual.
+    static void IRAM_ATTR dac_speed_thunk(Spindle* s, uint32_t dev_speed) {
+        static_cast<Dac*>(s)->Dac::setSpeedfromISR(dev_speed);
+    }
+    Spindle::IsrSpeedFn Dac::isr_speed_fn() {
+        return dac_speed_thunk;
+    }
 }
 #endif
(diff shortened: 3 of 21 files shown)


f1adang:
Sorry for the AI-generated PR flood.

Claudia tells me this one is the one I'd push hardest, since the `platformio_override.ini` finding means every ESP32 build currently ships with all 240 vtables in flash, and the linker script that was meant to prevent it has been inert.

f1adang:

I have split this PR and want to correct how I originally framed it, because I had the cause and the remedy the wrong way round.

**The build-configuration commit is now #1821, and it is the actual root cause.** `platformio_override.ini` was silently dropping `vtable_in_dram.py`, because `extra_scripts` replaces the inherited value instead of appending to it. That is what put the `Spindle` vtables in flash on the image that panicked.

On a stock build that never happens. `common_esp32.extra_scripts` already runs the script, and `vtable_in_dram.ld` already names the relevant objects:

```
    **Detail.cpp.o(.rodata .rodata.*)
    **Spindle.cpp.o(.rodata .rodata.*)
```

So the spindle half of what remains here is defence in depth against a hazard the project's build configuration already handles by other means. I presented it as fixing a live bug; it is more accurate to say it removes a dependency on a linker script matching object files by filename pattern. That may still be worth having — a new spindle in a file not ending `Spindle.cpp` would quietly fall outside the pattern — but it is a design judgement about robustness, not a crash fix, and it touches nine classes. Your call entirely, and I will drop it without argument if you would rather not carry it.

**`state_is()` is different, and is the part I would still argue for.** It is not a vtable, so `vtable_in_dram.ld` does nothing for it: it is an ordinary function that lived in flash at `0x4011b63c` and was called directly from the step ISR at `Stepper.cpp:236`. Marking it `IRAM_ATTR` is a one-line fix for a genuine flash access from an IRAM-registered ISR, and it is unaffected by #1821.

If you would prefer, I am happy to narrow this PR to just the `state_is()` change and drop the spindle dispatch entirely. Say the word and I will push that.

The premise underneath all of it does hold, for the record: `esp_intr_alloc_intrstatus(..., ESP_INTR_FLAG_IRAM | ESP_INTR_FLAG_LEVEL3, ...)` in `FluidNC/esp32/StepTimer.cpp` means the step ISR continues to run while the flash cache is disabled, so any flash access on that path is a real fault rather than a latent style issue.

PR #1818, opened 2026-09-04
Fix two channel-lifetime bugs in the job path

f1adang:
> _Found and written by **Claude** (Anthropic's AI assistant), at the direction of **Claudia**, who owns the fork this comes from. Flagging that up front so it gets human scrutiny rather than being taken on trust — I have said explicitly below what was observed on hardware and what is reasoning from the code._

Two lifetime bugs in the job path. Both are cases of a channel being used after its owner released it, on opposite sides of a job: the **output** channel is freed too early by the reaper, and the **input** channel is freed too early by the job itself.

---

### 1. `Job::leader` can be freed while the job is still writing to it

`Job::leader` is the channel that launched a job; per-line output goes there rather than to the job's own input channel. `leader_channel()` returns `nullptr` once the job stack is empty, which guards against reading it *after* a job — but not against the object being freed *during* one.

A job started from WebUI keeps a `WebClient` (or a WebSocket) as its leader. When that channel dies — client disconnects, or WiFi drops — `allChannels.kill()` deregisters it and `reap_channels()` deletes it as soon as its reference counts reach zero. Between GCode lines that is immediately. `leader` then dangles and the job writes through it for every subsequent line.

`nest()` now takes a processing reference and `pop()` releases it, so the object cannot be reaped while a job still refers to it. `leader_channel()` notices a channel that has begun closing, reports it once, drops it and returns `nullptr` — **the job keeps running**, because losing the WebUI connection is not a reason to abandon a cut already underway.

Also adds the `write(const uint8_t*, size_t)` override `InputFile` was missing alongside `write(uint8_t)`. Without it, a job with no leader falls back to its own input channel and reaches `FileStream::write()`, which would write report text into the GCode file being read.

### 2. A job's channel is deleted while a queued log message still references it

Finishing a small macro aborted the controller:

```
abort() was called at PC 0x401de86b
output_loop            Protocol.cpp:129
  Channel::print_msg   Channel.cpp:428
    Print::write
      __cxa_pure_virtual -> std::terminate -> abort
```

A pure virtual call means the object's vtable is no longer its own — the channel had been destroyed while the output task still held a queued message for it.

Every other Channel is torn down through `AllChannels::kill()`, whose reaper refuses to delete anything with outstanding log or processing references. A job's channel was the exception:

```cpp
~JobSource() { delete _channel; }
```

`Channel::sendLine()` takes a log reference and queues the message with a bare pointer, so a message queued shortly before the job ends is still in `message_queue` when the channel is freed. `output_loop()` then dereferences it. Its `is_closing()` check cannot help: nothing set that flag, and the deletion races the check either way.

Short jobs make this easy to hit — the window between queuing a message and tearing the job down is small.

`begin_closing()` first, so `try_acquire_log_ref()` fails and nothing further can be queued against the channel, then wait for the outstanding references to drain as `output_loop()` processes them. In practice that is immediate; the 200 ms bound is only so a wedged output task cannot hang job teardown.

Deleting the channel here rather than deferring it to the reaper is deliberate: it holds an open file, and with only two SD descriptors, `Job::unnest()` reopening the parent file must not wait for a reaper pass to get one back.

---

### Evidence, stated plainly

**Bug 2** was observed on hardware and the backtrace above is real, decoded against a matching build.

**Correction (important).** I previously wrote here that the fix for bug 2 was confirmed on hardware. That was wrong and I have withdrawn it. A later capture shows the *same* abort, with the same signature, on a build that contains this fix:

```
[MSG:INFO: /zero.g job sent]
abort() was called at PC 0x401de8a7
  output_loop -> Channel::print_msg -> Print::write -> __cxa_pure_virtual
```

So this commit closes a real lifetime hole — a job's channel is the only Channel in the tree deleted directly rather than through the reaper's reference checks — but it is **not** the cause of that crash, or not the whole of it. The reporter's machine also produces several unrelated fault signatures in the same sessions (a null instruction fetch from `loop()`, a `LoadProhibited`, an interrupt-watchdog reset), which together look more like memory corruption than a single logic error. That is still being investigated.

**Bug 1** was found by *reading the code* while chasing reboots after WiFi loss — the dumps I decoded at the time pointed at heap exhaustion elsewhere, not at it. I can show the lifetime hole and the reaper that closes it, but I cannot point at a dump and say "this one was that".

Both changes stand on their own as correctness fixes. Neither is confirmed to fix a specific observed crash, and I would rather say so than have this merged on a claim I cannot support.

Builds on wifi, noradio, bt and rp2040_wifi, each commit individually; unit tests pass.

### DIFF
--- a/FluidNC/src/InputFile.h
+++ b/FluidNC/src/InputFile.h
@@ -48,7 +48,12 @@ public:
     Error readLine(char* line, size_t len);
 
     // Channel methods
+    // An InputFile is opened read-only, so discard anything written to it as
+    // a Channel.  Both overloads must be overridden; otherwise a job with no
+    // leader channel would send its output to FileStream::write(), which
+    // would try to write the report text into the GCode file being read.
     size_t write(uint8_t c) override { return 0; }
+    size_t write(const uint8_t* buffer, size_t length) override { return 0; }
     void   ack(Error status) override;
     Error  pollLine(char* line) override;
 
--- a/FluidNC/src/Job.cpp
+++ b/FluidNC/src/Job.cpp
@@ -2,12 +2,41 @@
 // Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
 
 #include "Job.h"
+#include "NutsBolts.h"  // delay_ms()
 #include <map>
 #include <vector>
 #include <freertos/FreeRTOS.h>
 #include <freertos/semphr.h>
 
 std::vector<JobSource*> job;
+
+// A job's channel is the one Channel in the system that is deleted directly
+// rather than handed to AllChannels::kill() and its reaper, so nothing was
+// checking whether the output task still had a queued message referencing it.
+// Channel::sendLine() takes a log reference and queues the message with a bare
+// pointer; if the channel is freed before output_loop() drains it, the task
+// calls a virtual method on freed memory.  That lands on __cxa_pure_virtual
+// and aborts:
+//
+//   output_loop -> Channel::print_msg -> Print::write -> __cxa_pure_virtual
+//
+// Short jobs make it easy to hit, because the window between queuing a message
+// and tearing the job down is small.
+//
+// begin_closing() makes try_acquire_log_ref() fail, so no further messages can
+// be queued against this channel, and the references already outstanding drain
+// as output_loop() processes them.  In practice that is immediate; the bound is
+// there so a wedged output task cannot hang job teardown.
+JobSource::~JobSource() {
+    if (_channel) {
+        _channel->begin_closing();
+        for (uint32_t waited = 0; waited < 200 && _channel->pending_log_refs(); ++waited) {
+            delay_ms(1);
+        }
+        delete _channel;
+    }
+}
+
 
 Channel* Job::leader = nullptr;
 
--- a/FluidNC/src/Job.cpp
+++ b/FluidNC/src/Job.cpp
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a GPLv3 license that can be found in the LICENSE file.
 
 #include "Job.h"
+#include "Logging.h"
 #include "Serial.h"  // allChannels
 #include <map>
 #include <vector>
--- a/FluidNC/src/Job.cpp
+++ b/FluidNC/src/Job.cpp
@@ -61,7 +61,20 @@ void Job::nest(Channel* in_channel, Channel* out_channel) {
     JobLock lock;
     auto source = new JobSource(in_channel);
     if (out_channel && job.empty()) {
-        leader = out_channel;
+        // Hold a processing reference for the duration of the job.  A leader
+        // can die while the job runs - a WebSocket or an HTTP client
+        // disappears as soon as WiFi connectivity is lost - and the channel is
+        // then deregistered and deleted as soon as its reference counts reach
+        // zero, which between GCode lines is immediately.  The reference keeps
+        // the object alive, so leader stays a valid pointer; writes to a
+        // channel that is closing are discarded by the channel itself.
+        //
+        // leader_channel() returning nullptr once the stack is empty guards
+        // against reading it after the job, but not against the object being
+        // freed during the job.
+        if (out_channel->try_acquire_processing_ref()) {
+            leader = out_channel;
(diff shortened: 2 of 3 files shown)


f1adang:
Sorry for all the vibe-generated PRs.

Let me know if I should stop.

f1adang:
> _Written by **Claude** (Anthropic's AI assistant), at the direction of **Claudia**._

Correcting myself: I edited this PR earlier today to say the second fix was "confirmed on hardware". **That was wrong and I have removed it.**

A later capture shows the same abort, same signature, on a build containing the fix:

```
[MSG:INFO: /zero.g job sent]
abort() was called at PC 0x401de8a7
  output_loop -> Channel::print_msg -> Print::write -> __cxa_pure_virtual
```

I had taken a "this fixes macro runs" report as confirmation for this commit specifically, when the build being praised was a later one and the evidence did not actually separate the two. My mistake, and the sort that matters — it would have put an unsupported claim in front of a reviewer.

The commit still closes a genuine hole: a job's channel is the only Channel in the tree deleted directly rather than through `AllChannels::kill()` and the reaper's reference checks. But it is not the cause of that crash, or not all of it. The same machine also produces a null instruction fetch from `loop()`, a `LoadProhibited`, and an interrupt-watchdog reset in the same sessions — a spread of signatures that points at memory corruption rather than one logic error. Still digging.

Both changes here stand as correctness fixes on their own reading. Neither should be merged on the strength of fixing a specific observed crash.

f1adang:
Follow-up: the fix in `917ead43` was **incomplete**, and I have pushed `3963a371` to correct it. Please review that commit rather than the earlier one.

Written by Claude, at the direction of Claudia, and this comment likewise.

## What was wrong with the previous fix

`917ead43` made `~JobSource()` wait for the channel's queued *log* references to drain, then delete the channel. A Channel is referenced in two ways, and it only accounted for one.

`protocol_main_loop()` holds a **processing** reference across `execute_line()` and releases it on the line *after* it calls `channel->ack()`:

```cpp
Error status_code = execute_line(item.line, *out_channel, ...);

// If the line was aborted, the channel could be invalid.
if (!sys.abort()) {
    channel->ack(status_code);          // <-- here
}
channel->release_processing_ref();
```

A job's channel is torn down from inside `execute_line()`, so that processing reference is outstanding at exactly the moment the job is destroyed. Deleting the channel there leaves `ack()` reading a freed vtable:

```
Guru Meditation Error: Core 1 panic'ed (LoadProhibited)
PC: 0x4010ff69   EXCVADDR: 0x0000003c
protocol_main_loop() at Protocol.cpp:418
```

`0x3c` is the method slot offset: the vtable pointer itself loaded as zero.

The bounded wait also made the failure mode worse rather than better — once the 200 ms expired it deleted the channel regardless of any outstanding reference.

## The corrected fix

`AllChannels` already has exactly the right machinery. Its reaper deletes a channel only once **both** counts reach zero:

```cpp
if (channel->pending_log_refs() || channel->pending_processing_refs()) {
    ++it;
    continue;
}
```

So `~JobSource()` now hands the channel to `AllChannels::kill()` and lets `poll()` deregister and reap it, instead of deleting it itself. No new machinery, no timeout, and both reference kinds are honoured.

Worth noting for reviewers: grepping the tree at `609b0626`, there are exactly two places a `Channel` is deleted — inside `reap_channels()`, correctly guarded, and `Job.h:40`:

```cpp
~JobSource() { delete _channel; }
```

The job's channel is the single object that bypasses the safety net the rest of the codebase already builds for this.

## Testing

Both failure paths were reproduced on hardware (ESP32, WebUI-launched macros, browser disconnecting mid-job) and each backtrace was decoded against the matching ELF, verified by `ELF file SHA256`:

| Path | Reference missed |
|---|---|
| `output_loop -> Channel::print_msg -> Print::write -> __cxa_pure_virtual` | queued log ref |
| `protocol_main_loop -> channel->ack()`, `EXCVADDR 0x3c` | processing ref |

With `3963a371` applied, the macros that previously panicked run without crashing.

One caveat I would rather state than leave implied: that build also carried an unrelated motor-driver fix, so the confirmation covers the pair. The attribution to this commit rests on the decoded backtrace, not on the passing run alone.

PR #1820, opened 2026-09-04
Stop an exception in a support task from killing the controller

f1adang:
> _Found and written by **Claude** (Anthropic's AI assistant), working on a fork at the owner's direction. Flagging that up front so this gets human scrutiny rather than being taken on trust — I have said explicitly below what was observed on hardware and what is reasoning from the code._

### An exception in a support task takes the controller down

`polling_loop()` and `output_loop()` are raw FreeRTOS tasks with no exception handling. Anything that escapes either reaches `std::terminate()`, which panics the board and takes a running job with it.

Observed on hardware, backtrace decoded against the image:

```
polling_loop -> pollChannels -> AllChannels::poll -> Channel::pollLine
  -> Channel::autoReport -> report_realtime_status
    -> LogStream::write -> std::string::push_back
      -> operator new -> __cxa_allocate_exception -> std::terminate -> abort
```

A status report could not be formatted, and a three-hour cut ended. Motion planning and stepping do not allocate, so catching this and taking another pass leaves the job running.

### The change

The body of `polling_loop()` moves into `poll_once()` so a whole pass can sit in a `try` block; its `continue` statements become `return`, which is equivalent. `output_loop()` drops the offending message but always releases the log reference, so the channel can still be reaped. `WebClients::background_task()` gets the same treatment.

Reporting the failure can itself throw when the heap is what failed, so that logging is wrapped too.

### What this does not do

**This is containment, not a cure.** In the trace above the runtime could not allocate the `std::bad_alloc` object either — that is what `__cxa_allocate_exception` calling `std::terminate` means — so no handler can intervene at that point. This helps where an allocation fails while a little headroom remains, and it removes a whole class of "any throw anywhere reboots the machine". It does not make the board resilient to genuine heap exhaustion; that needs the leak fixed, which is separate work.

Builds on wifi, noradio, bt and rp2040_wifi; unit tests pass.

### DIFF
--- a/FluidNC/src/Protocol.cpp
+++ b/FluidNC/src/Protocol.cpp
@@ -285,8 +285,12 @@ void polling_loop(void* unused) {
             poll_once();
         } catch (const std::exception& ex) {
             report_poll_exception(ex.what());
+            feed_watchdog();
+            vTaskDelay(1);
         } catch (...) {
             report_poll_exception("unknown exception");
+            feed_watchdog();
+            vTaskDelay(1);
         }
     }
 }
--- a/FluidNC/src/Protocol.cpp
+++ b/FluidNC/src/Protocol.cpp
@@ -111,18 +111,25 @@ void output_loop(void* unused) {
         // Block until a message is received
         LogMessage message;
         if (xQueueReceive(message_queue, &message, 100)) {  // Use timeout to check exit flag
-            if (!message.channel->is_closing()) {
-                if (message.isString) {
-                    std::string* s = static_cast<std::string*>(message.line);
-                    message.channel->print_msg(message.level, s->c_str());
-                    delete s;
-                } else {
-                    const char* cp = static_cast<const char*>(message.line);
-                    message.channel->print_msg(message.level, cp);
+            // Sending can throw - std::bad_alloc when the heap is exhausted,
+            // for instance.  An exception escaping a raw FreeRTOS task reaches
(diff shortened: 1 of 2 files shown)


f1adang:
From Claudia with love.

Let's hear what Co-Pilot has to say about this!

f1adang:
> _Written by **Claude** (Anthropic's AI assistant), at the direction of **Claudia**._

A correction to the scope of this PR, before anyone reads more into it than it does.

The description says this removes a class of "any throw anywhere reboots the machine". That is still accurate, but I have since hit a fatal path in `output_loop()` that the `try`/`catch` added here **cannot** catch:

```
output_loop            Protocol.cpp:129
  Channel::print_msg   Channel.cpp:428
    Print::write
      __cxa_pure_virtual -> std::terminate -> abort
```

`__cxa_pure_virtual` calls `std::terminate()` directly rather than throwing, so no handler is involved. That crash is a channel being destroyed while the output task still holds a queued message for it, and it is fixed separately in #1818 by making the job path respect the reference counts, not here.

Flagging it so this PR is not read as making `output_loop()` safe. It narrows the ways that task can kill the controller; it does not close them all.

MitchBradley:
Some other non-FluidNC tasks need my attention, so I can't work on this today.

f1adang:
No rush. Claudia felt like this needed to be upstreamed, so we did.

Feel free to review whenever you find the time and interest.

PR #1824, opened 2026-09-05
Answer ESP410 WiFi AP scans asynchronously (portable)

MitchBradley:
Reworks #1748 to keep all WiFi-specific detail behind the `wifiImpl()` portability seam instead of in `WebUIServer.cpp`, so it builds across every MCU target that compiles the WebUI (esp32, esp32s3, rp2040, rp2350).

## Problem

Requesting a WiFi AP scan from the WebUI (the magnifying glass next to `Sta/SSID`) or sending `[ESP410]` runs the multi-second scan inline on the web command task. It produces no response bytes for several seconds and trips the connection watchdog, restarting the controller.

## Approach

Pause the HTTP GET request for the duration of the scan and send the AP list once results are ready, driven from `WebUI_Server::poll()`.

- **`WifiImpl`** gains non-blocking scan primitives: `startApListScan()` / `apListScanState()` (`Running`/`Done`/`Failed`) / `apListCount()`. The blocking `beginApListScan()` (still used by the serial/telnet/websocket `[ESP410]` path) becomes a single definition in `WifiImplCommon.cpp` built on those primitives, replacing the per-platform copies.
  - esp32: real async scan (`WiFi.scanNetworks(true, …)` + `scanComplete()`).
  - rp2040/rp2350: one synchronous `scanNetworks(false)` (arduino-pico has no async mode), failed scan reported as 0 APs — same behavior as before.
- **New `WebUI/WifiScanAsync.{h,cpp}`** (no `#ifdef`s): `beginAsyncWifiScan()` pauses the request and starts the scan; `pollAsyncWifiScan()` completes it. Cross-task handoff uses a FreeRTOS semaphore (not `std::mutex`, which does not compile on arduino-pico).
- **`WebUI::encodeApList()`** in `WifiConfig.cpp` is now the single ESP410-body encoder, shared by the synchronous `listAPs` and the async path, so both return the same shape. `getSignal` folded into `apSignalPercent`.
- `WebUIServer.cpp` footprint is two `#if !HOSTED` call sites. The new module is excluded from the host/simulator build (no `wifiImpl()` there).
- Also adds an explicit empty `ARDUINO_EVENT_WIFI_SCAN_DONE` case to `wifiEventHandler()` so the scan no longer logs `WiFi event: 102`.

## Testing

Builds clean: `wifi`, `wifi_s3`, `rp2040_wifi`, `rp2350_wifi`. Verified on hardware that both the async (WebUI/HTTP GET) and synchronous (serial) `[ESP410]` paths return the AP list; the watchdog no longer fires.

Closes #1748.

### DIFF
--- a/FluidNC/esp32/WifiImpl.cpp
+++ b/FluidNC/esp32/WifiImpl.cpp
@@ -373,17 +373,27 @@ namespace WebUI {
             return result;
         }
 
-        int32_t beginApListScan() override {
-            while (true) {
-                int32_t n = WiFi.scanComplete();
-                if (n >= 0) {
-                    return n;
-                }
-                if (n == WIFI_SCAN_FAILED) {
-                    WiFi.scanNetworks(true, false, false, 1000);
-                }
-                delay(1000);
+        void startApListScan() override {
+            if (WiFi.scanComplete() == WIFI_SCAN_FAILED) {
+                // async, hidden=false, passive=false, max milliseconds per channel
+                WiFi.scanNetworks(true, false, false, 1000);
             }
+        }
+
+        ApScanState apListScanState() override {
+            switch (WiFi.scanComplete()) {
+                case WIFI_SCAN_RUNNING:
+                    return ApScanState::Running;
+                case WIFI_SCAN_FAILED:
+                    return ApScanState::Failed;
+                default:
+                    return ApScanState::Done;
+            }
+        }
+
+        int32_t apListCount() override {
+            int16_t n = WiFi.scanComplete();
+            return n > 0 ? n : 0;
         }
 
         bool isApProtected(int index) const override { return WiFi.encryptionType(index) != WIFI_AUTH_OPEN; }
--- a/FluidNC/esp32/WifiImpl.cpp
+++ b/FluidNC/esp32/WifiImpl.cpp
@@ -79,6 +79,9 @@ namespace WebUI {
                 disconnect_seen = false;
                 log_info_to(Console, "WiFi STA Connected");
                 break;
+            case ARDUINO_EVENT_WIFI_SCAN_DONE:
+                // Fired by every WiFi.scanNetworks() (e.g. ESP410 AP scan); nothing to do.
+                break;
             default:
                 log_debug_to(Console, "WiFi event: " << (int)event);
                 break;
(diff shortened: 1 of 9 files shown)


MitchBradley:
Thanks @copilot-pull-request-reviewer. Addressed in dd51f6ab:

**Premature completion on `ApScanState::Failed`** — `pollAsyncWifiScan()` now treats anything other than `Done` as "keep waiting": it (re)kicks `startApListScan()` (idempotent) on each poll and only ends the request when the scan reaches `Done` or `WIFI_SCAN_TIMEOUT_MS` (25 s) elapses. `Failed` is the state `apListScanState()` reports before a scan has started or when `scanNetworks()` fails to launch, so a transient start failure no longer returns an empty list. This also restores the retry behavior the old blocking `beginApListScan()` had, but now bounded by the timeout instead of looping forever.

**Blocking the webserver task** — the initial `startApListScan()` is moved out of `beginAsyncWifiScan()` into `pollAsyncWifiScan()`, so no WiFi work runs on the async webserver task at all now; on ESP32 the scan is fully async (`scanNetworks(true, ...)`). On RP2040/RP2350 `startApListScan()` is still one synchronous `scanNetworks(false)` because arduino-pico has no async scan mode — but that call has no retry loop and completes in a couple of seconds, and it now runs on the polling task, not the webserver task. The watchdog reset in #1748 was specific to the old ESP32 `beginApListScan()` retry loop, which this removes.

Rebuilt clean: `wifi`, `wifi_s3`, `rp2040_wifi`, `rp2350_wifi`.

MitchBradley:
Addressed in 5e8cddf7.

`pollAsyncWifiScan()` now defers the `startApListScan()` (re)kick until after the `s_pending` lock block is released. On RP2040/RP2350 that call blocks for the duration of the scan (arduino-pico has no async mode), and it was previously made while holding the mutex, so a concurrent `beginAsyncWifiScan()` on the webserver task could stall on it. The scan is only ever driven from the polling task, so `s_pending` needs no protection across that call — the mutex now guards only the pointer handoff between the two tasks, which is all it was ever meant to do. Comment on `s_mutex` updated to say so.

Rebuilt clean: `wifi_s3`, `rp2040_wifi` (and `wifi`, `rp2350_wifi` unaffected by this change).

Issue #1826, opened 2026-09-07
Problem: Executing $WiFi/ListAPs in versions 4.04 and later will result in a reboot and the inability to return to the scanned SSIDs


albertjacksoon:
### Wiki Search Terms

Executing $WiFi/ListAPs in versions 4.04 and later will result in a reboot and the inability to return to the scanned SSIDs

### Controller Board

6-pack

### Machine Description

ESP32S3 8MB

### Input Circuits

```yaml

```

### Configuration file

```yaml
none
```

### Startup Messages

```text
ELF file SHA256: d7f5b12f0

E (11672) esp_core_dump_flash: Core dump flash config is corrupted! CRC=0x7bd5c66f instead of 0x0
E (11681xx?x?x?xxx?x鴛?xxxx?xxxxxxx?x鴛?x鴢?xx鴛?x?x?x?鴛鴛xx?x鴛鴛?鴛鴛???xx?鴛鴛xx?x?x?xxx?xx?x??鴢?x鴢xx?x?x?xxx?x鴛?xxxx?xxxxxx?x鴛xx鴛鴛鴢??x鴛?xxx??xxxxx?x?x?鴛鴛xx???xx?鴛鴛xx?x?x?鴛鴛鴛?x?x?x?x鴛?x鴢xxx?x?x??xx?xx鴛鴛鴛?鴛鴛?xx?x?xx?x?ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0xc (RTC_SW_CPU_RST),boot:0x9 (SPI_FAST_FLASH_BOOT)
Saved PC:0x4209043c
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce2820,len:0x10cc
load:0x403c8700,len:0xc2c
load:0x403cb700,len:0x30c0
entry 0x403c88b8
E (463) esp_core_dump_flash: No core dump partition found!
E (463) esp_core_dump_flash: No core dump partition found!
E (28) task燋[MSG:INFO: uart_channel0 created]
```

### What happened?

Executing $WiFi/ListAPs in versions 4.04 and later will result in a reboot and the inability to return to the scanned SSIDs

### Did you use an AI?

i did not use ai

MitchBradley:
Did you execute $WiFi/ListAPs from a terminal program, or from some other UI program?

f1adang:
[...]

Issue #1775, opened 2026-08-06
Problem: Timed engine start_unstep() spins on stale _stepPulseEndTime during soft reset — task-watchdog reboot loop (all ESP32-S3 users)


drewsipher:
### Wiki Search Terms

task watchdog, task_wdt, reboot, soft reset, disconnect, esp32s3 restart, Timed engine

### Controller Board

Custom breakout board carrying an ESP32-S3-WROOM-1 (N16R8) 44-pin dual-USB devkit. Step/dir buffered through a 74AHCT541; FMD2740C external drivers. (Board design: https://github.com/drewsipher/sharkcnc/tree/main/hardware/fluidnc-bob-revD)

### Machine Description

Sherline mill conversion, 3 axes, external FMD2740C drivers, NEMA 23 steppers, relay spindle.

### Configuration file

```yaml
# fluidnc-bob rev D — bring-up config (FluidNC v4.0.4, ESP32-S3)
#
# Pin map taken from the AS-BUILT rev D board (schematic + PCB pad nets,
# verified 2026-08-06). NOTE: rev D physically swapped X<->Z step/dir and
# AUX<->spindle vs rev C — the rev D README's "map unchanged" table is wrong:
#   X step/dir = gpio.7 / gpio.6      (rev C had Z here)
#   Z step/dir = gpio.18 / gpio.17    (rev C had X here)
#   spindle relay = gpio.4, AUX = gpio.5
#   E-stop = gpio.8, shared EN (disable) = gpio.9
#
# Driver wiring assumed common-cathode as on rev C (step/dir ACTIVE HIGH).
# If rewired common-anode, add :low to every step_pin and direction_pin.
#
# Limits/probe: 10k pullups to 3V3, switches close to GND -> ACTIVE LOW.
# Unwired inputs read "not triggered" — safe for the bench.

name: "Sherline mill (revD)"
board: "fluidnc-bob-revD"

# v4.x has no RMT engine on ESP32-S3; Timed is fine at our ~15 kHz max.
stepping:
  engine: Timed
  idle_ms: 255
  pulse_us: 5
  dir_delay_us: 5
  disable_delay_us: 0

axes:
  # EN_G -> R42 -> Q2 shared EN- sink. High = motors disabled (LED2 on),
  # which is FluidNC's default disable polarity.
  shared_stepper_disable_pin: gpio.9
  x:
    # 20 TPI screw, 200 steps/rev, 1600 usteps/rev (1/8) -> 1259.84
[...]

PR #1827, opened 2026-09-08
Fix task-watchdog stall on soft reset in S3/RP2040 step engines (#1775)

MitchBradley:
## Problem

Intermittent `task_wdt` abort with `loopTask` (CPU 1) wedged in `spinUntil()`:

```
#0  spinUntil                       delay_usecs.cpp:45
#1  start_unstep()                  esp32s3/i2s_engine_dedicated.cpp:235
#2  Machine::Stepping::unstep()     Stepping.cpp:190
#3  Stepper::stop_stepping()        Stepper.cpp:176
#4  Stepper::reset()                Stepper.cpp:297
    ... execute_realtime_command -> protocol_exec_rt_system -> protocol_main_loop
```

This is #1775 again. #1778 fixed it, but only in `esp32/timed_engine.cpp` and `rp2040/timed_engine.cpp`.

## Cause

The "mark pulse-end time in `finish_step()`, spin it out later in `start_unstep()`" overlap trick relies on `_stepPulseEndTime` being a fresh deadline. But `Stepping::unstep()` is also called from `Stepper::reset() -> go_idle() -> stop_stepping()` on a soft reset / alarm, with **no preceding `finish_step()`** — so `start_unstep()` spins against a stale deadline.

On the S3, `spinUntil()`'s CCOUNT wraps every ~17.9 s at 240 MHz and the compare is signed, so a deadline 9–18 s in the past reads as far in the future → the loop busy-waits several seconds → task watchdog fires on `loopTask`. (RP2040's `spinUntil()` uses the free-running 1 MHz timer, ~35 min to wrap — same failure mode, longer stall.) Intermittent, depending on the phase between the last real step and the reset.

The same unguarded pattern (and a non-`volatile` deadline) is in three engines #1778 didn't touch:

- `FluidNC/esp32/esp32s3/i2s_engine_dedicated.cpp` — the S3 dedicated-GPIO I2S engine, the one hit in practice on esp32-s3
- `FluidNC/rp2040/pio_engine.cpp`
- `FluidNC/rp2040/i2s_engine.cpp`

## Fix

Port #1778's guard to all three: a `volatile bool` pending flag set in `finish_step()`, checked and cleared in `start_unstep()`, plus `volatile` on the deadline so the foreground reset path doesn't read a cached value.

The other engines (`esp32/i2s_engine.cpp`, both `rmt_engine*`, `rp2040/pio_engine_delayed.cpp`) return early from `start_unstep()` without spinning and need no change.

## Also in this PR: a second, related task-watchdog stall — `$WiFi/ListAPs`

A user hit a `task_wdt` abort running `$WiFi/ListAPs` (ESP410, the `$`-command / serial path) on this same build, and there are matching reports on builds without #1824. Different code, same class of bug, so it's folded in here.

`WifiConfig::listAPs()` → `WifiImpl::beginApListScan()` spins `startApListScan()` / `apListScanState()` with a `delay()` until the scan completes, synchronously on the caller's task — `loopTask` (via `protocol_main_loop()`) or the polling task, both TWDT-subscribed. The loop never fed the watchdog, so a multi-second scan trips it. #1824 moved the *HTTP* `[ESP410]` path off-task (`WifiScanAsync.cpp`) but did not touch `beginApListScan()`, so the `$` path fails with or without that commit.

Fix (`WifiImplCommon.cpp`): `feed_watchdog()` on every pass of the wait loop (a no-op when the caller isn't TWDT-subscribed), poll delay 1000 → 200 ms so the command also returns sooner, and — since feeding the watchdog removes the reboot that would otherwise end a wedged scan — a 25 s deadline that returns an empty AP list on expiry (same safety net `WifiScanAsync.cpp` uses).

On rp2040/rp2350 `startApListScan()` blocks the whole scan in one call, so there the feed only covers the gaps between passes; a fully watchdog-safe rp2040 scan would need the feed inside that port's `startApListScan()`. The reported crash is esp32-s3, whose scan is async and fully covered.

## Test

- esp32-s3: idle ~15 s, then several Ctrl-X soft resets — no more WDT abort. RP2040 engines are the same change by inspection.
- User retest confirmed `$WiFi/ListAPs` no longer trips the watchdog.
- `wifi_s3` and `rp2040_wifi` build clean.

### DIFF
--- a/FluidNC/esp32/esp32s3/i2s_engine_dedicated.cpp
+++ b/FluidNC/esp32/esp32s3/i2s_engine_dedicated.cpp
@@ -33,7 +33,12 @@ static bool i2s_out_initialized = 0;
 static uint32_t _pulse_delay_us;
 static uint32_t _dir_delay_us;
 
-static int32_t _stepPulseEndTime;
+// _stepPulseEndTime and _stepPulsePending are read/written from both the
+// stepper timer ISR (via Stepping::step()/finish_step()) and the foreground
+// reset path (Stepper::reset() -> stop_stepping() -> unstep()), so they're
+// volatile to keep the compiler from caching stale values across contexts.
+static volatile int32_t _stepPulseEndTime;
+static volatile bool    _stepPulsePending = false;
 
 static uint32_t i2s_output_ = 0;
 static uint32_t i2s_pulse_  = 0;
@@ -228,12 +233,20 @@ static void IRAM_ATTR start_step() {
 // will happen in start_unstep()
 static void IRAM_ATTR finish_step() {
     _stepPulseEndTime = usToEndTicks(_pulse_delay_us);
+    _stepPulsePending = true;
 
     i2s_out_gpio_shiftout(i2s_output_ ^ i2s_pulse_);
 }
 
 static bool IRAM_ATTR start_unstep() {
-    spinUntil(_stepPulseEndTime);
+    // Only spin out the pulse width if a step pulse is actually in flight.
+    // unstep() is also called from stop_stepping() on soft reset, where
+    // _stepPulseEndTime is stale; the CCOUNT-based spinUntil() can then
+    // spin for up to ~half the 32-bit wraparound and trip the task watchdog.
+    if (_stepPulsePending) {
+        _stepPulsePending = false;
+        spinUntil(_stepPulseEndTime);
+    }
     i2s_out_gpio_shiftout(i2s_output_);
     i2s_pulse_ = 0;
     return false;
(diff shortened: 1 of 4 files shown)


MitchBradley:
Addressed the Copilot review:

- **`beginApListScan()` unbounded loop** (dc9e0c6c) — added a 25 s deadline (the same bound `WifiScanAsync.cpp` uses for the async path); on expiry it returns an empty AP list instead of blocking the caller. The callers already call `finishApListScan()` unconditionally afterward, so a timed-out scan is still cleaned up.
- **WiFi-scan behavior not in the PR description** — updated the description with an "Also in this PR" section covering the `$WiFi/ListAPs` watchdog fix, why it's folded in here (same bug class, same build the user hit both on), and the rp2040 caveat.
- **"CCOUNT-based" wording on rp2040** — already fixed in the two "Update comments in start_unstep" commits on the branch (`spinUntil()` there is described as the free-running 1 MHz timer). The remaining "CCOUNT" mention is in `esp32s3/i2s_engine_dedicated.cpp`, where it's correct (Xtensa).

`wifi_s3` and `rp2040_wifi` build clean.

MitchBradley:
Probably fixes #1826 too

PR #1828, opened 2026-09-08
Generate the config validator schema from config_items.yaml; retire the hand JSON

MitchBradley:
## What

Replaces the hand-maintained `tools/fluidnc-config-schema.json` (~3,500 lines that had drifted from source) with a schema built at load time from `FluidNC/docs/config_items.yaml`, which is generated from the `// @config` source annotations. Types, ranges, enums, key sets, and section shapes now have one source of truth.

`FluidNC/src/Configuration/ItemDocs.md` already named this as the goal ("What this feeds → 1. Replace the hand/LLM-maintained `fluidnc-config-schema.json`").

## How

- **`tools/config_schema_adapter.py`** (new, ~260 lines) — `build_schema(config_items)` returns a Draft-2020-12 schema. Pure translation of `config_items.yaml` plus the generator's new `section_meta` / `pin_namespaces` / `vfd_named_types` blocks. The only hand-owned constants are the pin attribute-suffix grammar and the `gpio`/`no_pin`/`void` builtins (runtime parse grammar, not `@config` data).
- **`tools/fluidnc_validate_core.py`** — `load_schema()` calls the adapter; `normalize_permissive()` rewritten to walk the adapter's regular schema tree generically (was ~220 lines coupled to the old `$defs` names/shapes, now ~70); `scan_deprecated()` and all deprecation/`pinDeprecated` handling removed. Governing rule: a section/field is valid iff `config_items.yaml` lists it — to stop accepting something, remove its annotations from firmware source. `rgbled:` and `extenders:` therefore validate again.
- **`tools/build_config_docs.py`** — emits `section_meta` (repeatable + key_pattern for `<letter>`/`motorN`/`*N`-bus repeats, `child_types` for `<i2c_chip>`), `pin_namespaces` (from a new `// @pin_namespace` annotation on the `*PinDetail` subclasses), and `vfd_named_types` (grepped from `Spindles/VFD/*.cpp`), plus a CI check against `Pin.cpp`'s dispatch chain. Adds a `uartN.usb_host` section (a real `UartFactory` subsection the hand schema had but `config_items.yaml` lacked); splits the `midtbot`/`Cartesian` pseudo-entry into two real empty sections.
- **`tools/gen_config_docs.py`** — resolves symbolic `min`/`max` args (`MaxToolNumber`, `MAX_N_AXIS`, `set_mpos_only`, `0.0f`, `3e38`) to real numbers via a whole-tree constant scan; classifies a `handler.item(name, ..._pin)` whose type can't be resolved as a pin, and a numeric-bounded unresolved type as integer/float. No more `type: unknown` in `config_items.yaml`.
- **Firmware:** `FluidNC/esp32/esp32s3/UsbHostUart.cpp` gains the `@config baud` annotation; `FluidNC/src/Motors/TrinamicUartDriver.h` range-checks `addr` (0–3) in source (was only enforced by a hand-added schema bound). Three `// @pin_namespace` one-line annotations in `src/Pins/`.
- **CLI/MCP:** `--schema` now names an alternate `config_items.yaml`; the raw-JSON path is gone. `.github/instructions/*` and `ItemDocs.md` updated to the generated-from-source model.
- **CI:** the existing `config-docs` job also asserts `config_items.yaml` is committed in sync and smoke-tests that the adapter produces a valid schema for every `example_configs/*.yaml`.
- **`tools/fluidnc-config-spec.md`** — reframed as the grammar / cross-field-rules / worked-examples companion (1281 → 409 lines); the per-field tables are now `config_items.yaml`'s job. A matching wiki edit to `config/overview.md` is already on `bdring/fluidnc-wiki-content` main.

## Verification

- **Golden diff** of the adapter output vs the retired schema: 480 shared field leaves, 17 differences — every one a source-truth correction the hand file had wrong or missing (`tool_num` max `255→99999999`, `homing/cycle` max `6→9`, axis letters `xyzabc→xyzabcuvw`, `stepping/engine` gains `Simulator`/`PIO`, `usb_host/baud` range, `+ethernet`, `+rgbled`, `+i2so.oe_pin`).
- **47 real configs** (`data-wmb/`, `example_configs/`, …): identical strict verdict on 45. The 2 that change go invalid→valid because the hand schema was missing the real `ethernet:` section.
- Build passes (`noradio_s3` / `wifi_s3`).


MitchBradley:
Addressed the Copilot review in 6b230406:

1. **motorN allows multiple driver blocks** — restored the retired schema's `oneOf: [{required: [<driver>]}]` over all 14 driver-type keys. Now rejects a motor block naming two drivers, or none (`null_motor` is one of the choices).
2. **kinematics not restricted to one type** — restored `maxProperties: 1`.
3. **pinextenderN allows multiple chip keys** — restored `maxProperties: 1` on each instance.
4. **`meta` unconstrained** — now `{"not": {"type": ["object", "array"]}}`. Kept deliberately loose otherwise: real configs put a bare build date there (`meta: 2022-03-15`), which a YAML parser yields as a date object rather than a string, so `{type: string}` would reject valid configs — but a map/list is now rejected, which was the actual concern.
5. **permissive mode doesn't fix casing for patternProperties keys** — `normalize_permissive()` now lowercases a key that matches a schema `patternProperties` regex only case-insensitively (`X`→`x`, `Motor0`→`motor0`, `UART1`→`uart1`). Every such pattern in the generated schema is lowercase, so the canonical form is the key's lowercase.

Corpus check after the fixes: 42/47 real configs now have an identical strict verdict vs the retired hand schema (was 41 — the motor `oneOf` brings `rp2040_3axis.yaml` into verdict-parity); the remaining 3 are the two `ethernet:` invalid→valid flips (the hand schema omitted the real section) and one P4-experimental config.

PR #1833, opened 2026-09-09
ESP-NOW: stage heap allocation, deferring ~29K until first use

MitchBradley:
## Problem

Adding ESP-NOW to the `wifi_s3` build (#1795) cost ~29K of free heap at the initial prompt whether or not the node ever talks to a pendant — free heap at the prompt dropped from ~84K to ~55K.

Wi-Fi is already running in that build, so the cost is not Wi-Fi bring-up. It is ESP-NOW's own storage, most of it static in the `espnowChannel` global:

| Item | Size |
|---|---|
| `PairedPeer _paired_peers[8]`, each embedding a ~1.9K `FragBuf` reassembly buffer | ~15.4K `.bss` |
| `static uint8_t _rx_buf[4096]` | 4K `.bss` |
| `_packet_queue` (16 x 258) + `_pairing_queue` + `_pair_confirm_queue`, created in the constructor | ~6K heap |
| `esp_now_init()` internals | ~1.5K heap |

None of it is needed until a pendant is actually paired and connected.

## Change

Allocate on demand, in stages that follow real use. No config flag — the node self-arms based on whether a pairing exists.

- **Dark** (no saved pairing, no pairing window): nothing allocated. `init()` registers the `$ESPNow` commands and the channel, probes NVS, and stays idle. `esp_now_init()` is not called and Wi-Fi is not forced on.
- **Armed** (a saved pairing exists, or the operator runs `$ESPNow/Pair`): `arm()` does `esp_now_init()`, registers the callbacks, creates `_packet_queue`, and loads the roster. ~6K. Retried from `poll()` if Wi-Fi is not up yet at `init()` time.
- **Connected + streaming**: `_rx_buf` (4K) is `malloc`'d on the first message that reaches the line parser; a peer's `FragBuf` (~1.9K) is `calloc`'d on its first inbound data fragment and freed when it disconnects (idle timeout / jog-cancel), so it is not re-malloc'd per line.
- **Pairing window**: the two pairing queues (~1.5K) are created on the first window and then reset (not deleted) for the rest of the session.

All expansions are one-way within a boot session and happen in the poll task, so they never race the Wi-Fi-task receive callback. `_rx_buf` and `PairedPeer::frag` are now pointers; readers null-guard and treat an unallocated buffer as empty.

## Measurements (`wifi_s3`)

- Static RAM: 143648 -> 124200 bytes (`pio run` size report).
- Free heap at the initial prompt, no pendant paired: ~48K -> ~74K.
- With a paired pendant connected and streaming, peak ESP-NOW heap is roughly the pre-change figure, but only while the pendant is in use.

## Not done here

- `_pairing` (the `PairingTransaction`, ~530 B) is still in `.bss`; moving it behind a pointer touches ~60 call sites for half a K — left as a follow-up.
- `_packet_queue` is created at full depth 16 in `arm()` rather than grown on connect; growing a FreeRTOS queue means delete+recreate, which would race `onRecv`.

### DIFF
--- a/FluidNC/esp32/ESPNow/ESPNowChannel.cpp
+++ b/FluidNC/esp32/ESPNow/ESPNowChannel.cpp
@@ -325,8 +325,13 @@ bool ESPNowChannel::arm() {
         WiFi.disconnect(false, false);
     }
 
-    if (esp_now_init() != ESP_OK) {
-        log_debug("ESP-NOW: esp_now_init() deferred (Wi-Fi not ready)");
+    esp_err_t init_err = esp_now_init();
+    if (init_err != ESP_OK) {
+        // ESP_ERR_ESPNOW_INTERNAL here usually means Wi-Fi has not finished
+        // starting yet; poll() will retry.  Log at debug so a transient
+        // startup race does not spam, but name the actual error so a
+        // persistent failure is still diagnosable.
+        log_debug("ESP-NOW: esp_now_init() failed: " << esp_err_to_name(init_err) << " (will retry)");
         return false;
     }
 
@@ -524,6 +529,9 @@ bool ESPNowChannel::removeRuntimePeer(const uint8_t* mac) {
 
     _paired_count.store(last, std::memory_order_release);
     _paired.store(last > 0, std::memory_order_release);
+    if (last == 0) {
+        _want_arm = false;  // roster empty: cancel any pending arm() retry
+    }
     int active = _active_peer_index.load(std::memory_order_acquire);
     if (active == index || active >= (int)last) {
         _active_peer_index.store(last > 0 ? 0 : -1, std::memory_order_release);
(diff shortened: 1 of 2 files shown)


MitchBradley:
Thanks — both points addressed in 950e8219.

**`_want_arm` retry after pairings cleared:** `clearPairings()` and `removeRuntimePeer()` now set `_want_arm = false` when the persisted roster reaches empty. So if the operator clears pairings before Wi-Fi has finished starting (i.e. before `arm()` has ever run), the pending `poll()` retry is cancelled and the node stays dark. A node that was *already* armed stays armed — that only costs the ~6K that was already spent, and tearing `esp_now_init()` + the packet queue down would race the Wi-Fi-task recv callback, so it's deliberately left as a one-way transition within a boot session.

**Misleading `esp_now_init()` message:** `arm()` now logs the actual `esp_err_t` via `esp_err_to_name()` ("esp_now_init() failed: <name> (will retry)") instead of always attributing it to Wi-Fi not being ready. Kept at debug level so the normal startup race doesn't spam, but a persistent failure is now identifiable.

PR #1849, opened 2026-09-14
Emit the low-memory warning, without allocating to do it

f1adang:

`heap_monitor_poll()` tracks the heap low-water mark and decides when a warning is due, but the warning itself is commented out:

```cpp
if ((heapLowWater < heapLowWaterReported - 2048) || (ticksSinceReported > (uint32_t)usToCpuTicks(200000))) {
    //                log_warn("Low memory: " << heapLowWater << " bytes");
    heapLowWaterReported   = heapLowWater;
    heapLowWaterReportTime = getCpuTicks();
}
```

So the throttling, the 2 KB step and the low-water bookkeeping all run, and nothing is ever reported. On a board that is quietly running out of memory the operator gets no warning at all.

## Why it cannot simply be uncommented

`log_warn` builds a `std::string` to do its work. This fires precisely when the heap is nearly gone, so that allocation is the one most likely to fail — and when it does, the runtime cannot allocate the `std::bad_alloc` object either, so it calls `std::terminate()` outright. No catch block can intervene. The warning that you are about to run out of memory must not be the thing that finishes you off.

That is a plausible reason for it being commented out rather than removed.

## The change

A small allocation-free reporter, and the call restored:

```cpp
static void report_low_memory(uint32_t bytes) {
    char msg[48];
    int  len = snprintf(msg, sizeof(msg), "[MSG:WARN: Low memory: %u bytes]\n", static_cast<unsigned>(bytes));
    if (len > 0) {
        Console.write(reinterpret_cast<const uint8_t*>(msg), static_cast<size_t>(len));
    }
}
```

Fixed buffer, no heap. Straight to the console rather than through the channel layer for the same reason, and because whatever is consuming the heap is often the network stack — the path most likely to be unable to deliver the message.

The throttling above it is untouched, so the reporting rate is whatever `heap_monitor_poll()` already decided: only when the low-water mark has actually dropped, and at most every 200 ms or on a 2 KB fall.

## Testing

Built for `wifi` and run on an ESP32. The warning appears as the heap falls, in the sequence it is supposed to:

```
[MSG:WARN: Low memory: 14476 bytes]
...
[MSG:WARN: Low memory: 644 bytes]
```

That run was on a board being driven out of memory by a WebUI upload during a job, which is how the missing warning was noticed in the first place.


MitchBradley:
Logging to the console does not help the substantial fraction of users who are using WebUI.

f1adang:
Yeah, not much benefit to regular users here, but helps getting indications on memory consumption issues, which was often the last thing I saw before a machine reset.

Thought it might be helpful to others.

PR #1850, opened 2026-09-14
An unusable SD card should report, not reboot the controller

f1adang:

A card that cannot be read — failing, or formatted as exFAT, which this FATFS build does not support — took the board down instead of being reported. Four commits, each a separate route to the same reboot.

## The failure

`sd_mount()` takes a long time to give up on a bad card, and tries twice. The WebUI asks for a directory listing every few seconds, and each attempt runs on `async_tcp`, which is subscribed to the task watchdog. So an unreadable card rebooted the controller over and over, never reporting anything useful:

```
E (33623) task_wdt: Task watchdog got triggered ... async_tcp
```

## The four changes

**Report instead of rebooting** (`092c69e4`). `handleFileOps()` now returns `{"status":"No SD card"}` rather than letting the mount attempt run on into the watchdog.

**Stop retrying every request** (`71162549`). A failed mount is remembered for 3 s and further attempts fail fast within that window, instead of each one paying for a full card initialisation. The window is short enough that swapping in a working card is still picked up on the next attempt.

**Bound the mount lock** (`5123d3c8`). `SDLock` waited `portMAX_DELAY`. It is taken by whichever task wants the card, and a slow card holds it for as long as the mount takes, so a web request could block behind a mount running on the polling task and sit there past the watchdog timeout. The watchdog then names `async_tcp` as the task that failed to check in, while the task actually running is the one holding the lock — which sends you looking in the wrong place. Now 2 s, then report the card as unavailable.

**Leave the watchdog during card init** (`f5d5062c`). Card initialisation has no point inside it at which the watchdog could be fed, so the task steps out of supervision for the duration rather than being killed mid-init. This adds `suspend_watchdog_for_task()` / `resume_watchdog_for_task()` and a `WatchdogSuspension` RAII guard; both are no-ops where the task is not subscribed, and the RP2040 and capture builds get matching stubs.

## Testing

Reproduced and verified on an ESP32 with an exFAT-formatted card and with a failing card. Before: a reboot loop as long as the WebUI was open. After: `No SD card` in the UI, the controller stays up, and inserting a working card is picked up within a few seconds.

`suspend_watchdog_for_task()` is deliberately paired with `esp_task_wdt_status()` so resume never adds a task that was not being watched to begin with — the flag is `thread_local`, so two tasks suspending concurrently do not interfere.


f1adang:
Fair point. On it.

f1adang:

Correct, and worse than a style inconsistency. Fixed in `075687ae`.

`suspend_watchdog_for_task()` and `resume_watchdog_for_task()` tested `CONFIG_ESP_TASK_WDT_EN` while every other function in the file uses `FLUIDNC_TASK_WDT_ENABLED`. On Arduino-ESP32 2.0.17 / IDF 4.4 only `CONFIG_ESP_TASK_WDT` is defined, so both compiled to empty bodies with the task watchdog still running — meaning the headline fix of this PR did nothing at all on the target it was written for. `sdmmc_card_init()` could still reboot the controller exactly as before.

This is the same mistake #1816 fixed, in the same file, reintroduced in new code written afterwards. I should have recognised the shape of it.

Both guards now use `FLUIDNC_TASK_WDT_ENABLED`.

## Verified against the image, not the source

Reading the source is what missed this in the first place, so I checked the linked binary:

```
suspend_watchdog_for_task:
    entry  a1, 32
    movi   a10, 0
    l32r   a8, ...        ; esp_task_wdt_status
    callx8 a8
    ...
    callx8 a8             ; esp_task_wdt_delete
    retw.n

resume_watchdog_for_task:
    entry  a1, 32
    ...
    callx8 a8             ; esp_task_wdt_add
    callx8 a8             ; esp_task_wdt_reset
    retw.n
```

Before the change both were a bare `entry` / `retw.n`. That is the difference the guard was making, and it is only visible in the output.

Worth noting for anyone reading later: a wrong `#ifdef` here fails silently and looks entirely correct in review. If the project would take it, a `static_assert` or a build-time check that the watchdog layer is actually compiled in would turn this class of mistake into a build failure instead of a reboot in the field. Happy to propose one separately if that is of interest.

PR #1852, opened 2026-09-14
Do not strand the upload file descriptor when a transfer is abandoned

f1adang:

Uploading several files at once, or closing the browser mid-upload, left the upload file open. With only a couple of SD descriptors available that blocked every later upload until the board was restarted.

Four commits.

**Close the file when the client goes away** (`c80c8f43`). An upload whose browser disconnects never reaches the completion callback, so the descriptor stayed held for the rest of the session. An `onDisconnect` handler now closes it.

**Reclaim a stranded file when the next upload starts** (`cfd8d970`). Belt and braces for the case where that callback is missed for any reason — the next `uploadStart()` closes whatever was left open rather than overwriting the pointer and leaking it.

**Do not throw out of the completion callback** (`841cd4da`). An exception escaping there propagates into the async web server's task, which is not prepared for it.

**Reduce per-chunk overhead in the write path** (`628ea82f`). The upload write path did more work per chunk than it needed to, which matters on a slow card where each chunk is already close to the watchdog budget.

## Testing

On an ESP32 with the WebUI:

- **Before:** uploading 13 files at once failed partway, and every subsequent upload failed until reboot.
- **After:** the same 13 files upload cleanly.

Also verified that closing the browser mid-upload no longer strands the descriptor — the next upload succeeds where it previously needed a restart.

## A limitation this does not fix

`_uploadFile` and `_upload_status` are single statics, so the design assumes one upload at a time and does not enforce it. These commits make an *abandoned* upload recoverable; they do not make genuinely concurrent uploads correct. I have not established whether the WebUI ever issues overlapping uploads or simply queues them, so I have not built anything on that assumption — flagging it as a separate question rather than quietly widening the scope here.


f1adang:
I'll have Claude give it another bit of effort

f1adang:

All four points addressed in `c7374e71`. Every one of them was correct, including the suppressed comment about the mount — thank you, that one was subtle and I had it wrong.

## Critical: pointer reuse

Right, and my guard was worse than useless because it looked like it handled the case. I compared `_uploadFile` against the pointer the request opened, intending to mean "a later upload has taken over, leave it alone". That is not what the comparison tests. The allocator readily hands a just-freed block back for the next allocation of the same size, so the next upload's `FileStream` can land at exactly the address the finished one had. The test then passes and the callback tears down an upload that is still running — the bug the guard was written to prevent, made likelier rather than less likely by the reuse.

Now a counter, bumped per upload and captured by the handler, which does not depend on where the allocator put anything.

## Suppressed comment: FluidPath and the mount

This is the one I am glad you flagged. `FileStream::fpath()` returns a `FluidPath`, which holds a `shared_ptr<SDMountState>`; assigning it to a `std::filesystem::path` slices that away. Deleting the stream could then drop the last reference and unmount the card before `stdfs::remove()` ran, leaving behind precisely the partial file the handler exists to delete — and silently, since the error code was ignored.

Kept as a `FluidPath` now, so the mount is held until the removal is done. Your `WebDAV::write()` reference is the right precedent; I have taken the variant that holds the existing mount rather than re-mounting.

## Cleanup leaving incomplete files

Correct for the reclaim path, which closed the descriptor and left the half-written file on the card. Half a GCode file is worse than none: it runs, and stops somewhere arbitrary. It is removed now, with the same mount care as above.

## Fixed chunk count

Fair, and the reasoning is better than mine. How long a chunk takes is a property of the card, so 32 chunks is an unbounded amount of time on a slow one — which is the case this path exists to survive. Now yields when 20 ms have elapsed, so the interval is the same whatever the card does.

## One thing I did not change

`uploadStop()` and `uploadCheck()` slice `fpath()` the same way, so `uploadCheck()`'s `stdfs::remove()` can fail for the same reason. That is pre-existing rather than introduced here, so I have left it alone rather than widening this PR — happy to include it here or send it separately, whichever you prefer.

Rebuilt for `wifi`.

f1adang:
Oh dear, sorry for this "at the direction of Claudia" nonsense. I'll stop that.

Anyways, Opus 5 going:

**The critical one was worse than Copilot said**
My guard compared _uploadFile against the pointer the request opened, meaning "a later upload took over, leave it alone." That test doesn't hold — the allocator readily returns a just-freed block for the next allocation of the same size, so the next upload's FileStream can land at exactly the address the finished one had. The comparison then passes and the handler tears down a live upload.

So the guard didn't just fail to prevent the bug, it made it likelier, and it looked deliberate enough that I'd not have questioned it. Now a generation counter, which doesn't depend on where the allocator put anything.

**The suppressed comment was the subtle one**
FileStream::fpath() returns a FluidPath holding a shared_ptr<SDMountState>. Assigning it to std::filesystem::path slices that away — so deleting the stream could drop the last reference, unmount the card, and stdfs::remove() would then silently fail, leaving exactly the partial file the handler exists to delete.

That's the same SD mount-refcounting trap that cost several rounds on the resume checkpoints earlier. I'd just been bitten by it and still wrote it again.

**Also fixed**
Reclaim path now removes the stranded partial file, with the same mount care
Yield bounded by 20 ms elapsed rather than 32 chunks — Copilot's reasoning was better than mine, since chunk duration is a property of the card and this path exists to survive slow ones
One thing I deliberately left
uploadStop() and uploadCheck() slice fpath() the same way, so uploadCheck()'s removal can fail identically. That's pre-existing upstream code, not introduced here, so I flagged it in the reply and offered to fix it here or separately rather than silently widening someone else's review scope.

MitchBradley:
"Costs 4 KiB of heap per open file." 

I just spent days trying to save memory and you want to claim more?

f1adang:

`891ff8f0` cuts the memory this PR spends.

The write-path commit gave **every** `FileStream` a 4 KiB stdio buffer. The reasoning behind it is about coalescing small writes into SD multi-block writes — which says nothing about reads, and reads are where it cost most, because a job's file stays open for the entire job rather than the length of a transfer.

Now: write modes only, and 2 KiB rather than 4. Four sectors is still enough for FATFS to see runs of whole sectors.

Worst case, by scenario:

| | before | after |
|---|---|---|
| job running from SD | 4096 B | **0 B** |
| job + WebUI upload | 8192 B | 2048 B |
| job + upload + WebUI serving a page | 12288 B | 2048 B |

## On the size

The 4 KiB was chosen on host measurements — write() calls per megabyte, fourfold better than 1 KiB. Re-checking that, 2 KiB lands within a few percent of 4 KiB, and on the device neither was separable from run-to-run variation in the card itself.

I should not have shipped the larger number on evidence that thin, particularly in a PR whose point is robustness rather than throughput. When the measurement cannot tell two options apart, the cheaper one is the right one.

I have kept the buffering rather than dropping it entirely because the write path is the one place the rationale actually holds, and 2 KiB for the duration of an upload is a fair price. If you would rather this PR carried no performance change at all and stuck to the descriptor-lifetime fixes, say so and I will drop the `setvbuf` outright — that is a reasonable call and it would make the PR smaller.

PR #1855, opened 2026-09-14
A too-high stepping rate should not take the network down with it

### LINKED ISSUE #1762: Problem: Invalid stepping rate stops boot process
### Wiki Search Terms

Axes description mentions that exceeding max stepping rate may cause errors, but this does not explain why wireless stack is not initialized.

### Controller Board

MKS DLC32 MAX

### Machine Description

-

### Input Circuits

```yaml

```

### Configuration file

```yaml
-
```

### Startup Messages

```text
[MSG:INFO: uart_channel0 created]
[MSG:RST]
[MSG:INFO: FluidNC v4.x.x (feature/esp-now-02761b2) https://github.com/figamore/FluidNC]
[MSG:INFO: Local filesystem is /littlefs]
[MSG:INFO: Configuration file:config.yaml]
[MSG:WARN: PWM spindle set to tool 0]
[MSG:INFO: Machine MKC DLC32 Max Default]
[MSG:INFO: Board MKC DLC32 Max]
[MSG:INFO: UART1 Tx:gpio.17 Rx:gpio.18 RTS:NO_PIN Baud:115200]
[MSG:INFO: SPI SCK:gpio.13 MOSI:gpio.14 MISO:gpio.12]
[MSG:INFO: SD Card cs_pin:gpio.21 detect:gpio.10 freq:8000000]
[MSG:INFO: Stepping:Timed Pulse:4us Dsbl Delay:0us Dir Delay:0us Idle Delay:100ms]
[MSG:INFO: Axis count 3]
[MSG:INFO: Shared stepper disable gpio.8]
[MSG:INFO: Axis X (0.000,130.000)]
[MSG:INFO:   Motor0]
[MSG:INFO:     tmc_2209 UART1 Addr:0 CS:NO_PIN Step:gpio.16 Dir:gpio.15:low Disable:NO_PIN R:0.120]
[MSG:INFO:  X All Limit gpio.39]
[MSG:INFO: Axis Y (0.000,46.000)]
[MSG:INFO:   Motor0]
[MSG:INFO:     tmc_2209 UART1 Addr:1 CS:NO_PIN Step:gpio.7 Dir:gpio.6:low Disable:NO_PIN R:0.120]
[MSG:INFO:  Y All Limit gpio.40]
[MSG:INFO: Axis Z (-50.000,0.000)]
[MSG:ERR: Critical error in main_init: Stepping rate 213333 steps/sec exceeds the maximum rate 125000]
```

### User Interface Software

_No response_

### What happened?

Invalid stepping rate crashes boot process (including wireless connection) and forces a user to access the board through usb, delete existing configuration file using command line and then upload it via webui. Found following line in FluidNC/src/Machine/Axis.cpp:

`
Assert(stepRate <= maxRate, "Stepping rate %d steps/sec exceeds the maximum rate %d", stepRate, maxRate);
`

Wouldn't be better to clamp maximum steps per mm and display an error instead of asserting?

### Did you use an AI?

No.

### GCode File

_No response_

### Other Information

_No response_

f1adang:

Fixes #1762.

An axis whose `steps_per_mm` × `max_rate_mm_per_min` exceeds what the stepping engine can produce aborts the boot:

```
[MSG:INFO: Axis Z (-50.000,0.000)]
[MSG:ERR: Critical error in main_init: Stepping rate 213333 steps/sec exceeds the maximum rate 125000]
```

and nothing after that line runs.

## Why it costs the network

`Axes::init()` is inside `main_init()`'s try block, and the module init that brings up networking comes *after* it:

```cpp
Axes::init();                                   // <- Assert throws here
config->_control->init();
config->_kinematics->init();
limits_init();
for (auto const& module : Modules()) {          // <- WiFi never gets here
    module->init();
}
...
} catch (std::exception& ex) {
    log_config_error("Critical error in setup(): " << ex.what());
}
```

So one number too large in the config leaves the board reachable only over USB serial — to fix a config file the WebUI could otherwise have edited. The reporter had to delete it from the command line and re-upload. On a board that is awkward to reach, that is a long way to go for a typo.

## The change

Report it with `log_config_error()` rather than asserting. That is what this same function already does a few lines below for `motor1 defined without motor0`, so the pattern is established here.

`log_config_error()` raises `ConfigAlarm`, which is not a soft landing:

- `settings_execute_line()` rejects gcode with `Error::SystemGcLock`
- `protocol_main_loop()` aborts any running job

So the machine still refuses to move. What changes is that it finishes booting, brings up WiFi, and lets the config be corrected from the WebUI.

The message also names the axis now — with several axes configured, the old text gave no clue which one to fix.

## On clamping

The issue suggests clamping the rate instead. I have deliberately not done that: quietly running at a speed the user did not ask for turns a clear console error into a surprise at the tool, and the numbers that produce this are usually a units mistake rather than a small overshoot. Refusing to move, saying why, and staying reachable seemed the better of the two.

Happy to switch to clamping if you would rather — it is a judgement call about which failure is kinder, and it is yours to make.

## Testing

Built for `wifi`. Verified by reading the init order above and the two `ConfigAlarm` gates rather than by reproducing the original board's config; I do not have hardware that hits this limit, so I would value a check from the reporter.


f1adang:
The comment about hard to reach controllers..

I'll try getting my agent a bit less chatty.

MitchBradley:
Remove the overly-verbose comment leaving only the code and I will accept the PR.

f1adang:
Sure, done.

Issue #1745, opened 2026-06-09
Feature Request: UPS-Aware Power Loss Recovery and Job Resume


jahidhsn002:
Add an optional UPS-aware Power Loss Recovery feature to FluidNC that allows safe interruption and recovery of CNC jobs in environments where power outages are common.

This feature is intended for users who keep the entire CNC system (controller, stepper drivers, motors, spindle/VFD, and optionally the sender PC) powered from a UPS or battery for a short period (e.g. 1–5 minutes) after mains power is lost.

## Motivation

In many countries (e.g. India, Bangladesh, Pakistan, parts of Africa and South America), short power outages are a frequent occurrence.

Currently, an unexpected power loss immediately terminates a running job. Even when users install a UPS capable of powering the entire CNC system for several minutes, FluidNC cannot take advantage of that remaining runtime.

A UPS-aware recovery workflow would significantly reduce wasted material and machine downtime.

### During normal operation

FluidNC monitors a configurable "Power Fail" input (or another suitable mechanism).

### When power failure is detected

If UPS power is available:

1. Immediately perform Feed Hold.
2. Stop the spindle (if controllable).
3. Wait until the spindle reaches a safe state (optional/configurable).
4. Retract the Z axis to a safe height.
5. Move to a configurable parking position (optional).
6. Save the current job state to non-volatile storage.

Possible information to save:

- Current job file
- Current file position or line number
- Machine coordinates
- Active WCS (G54–G59)
- Modal state (G90/G91, G20/G21, etc.)
- Feed override
- Spindle speed
- Tool number (if applicable)

After parking:

- Continue waiting while UPS power is available.
- If mains power returns before the timeout expires, optionally resume automatically.
- Otherwise perform a clean shutdown.

## Startup Recovery

On boot, if an interrupted job is detected:

FluidNC could display something similar to:

Previous job interrupted.

File:
cabinet.nc

Saved Position:
X123.456
Y45.321
Z-6.000

Resume?

After user confirmation:

1. Home the machine (if required).
2. Verify machine state.
3. Restore modal state.
4. Return safely to the saved position.
5. Restart spindle.
6. Resume execution from the saved location.

## Configuration

Example configuration options:

power_recovery:
    enabled: true
    power_fail_pin: gpioXX
    park_on_power_fail: true
    park_x: 0
    park_y: 0
    park_z: 20
    wait_timeout: 300      # seconds
    auto_resume: false

## Notes

This feature assumes:

- The controller remains powered.
[...]

PR #1856, opened 2026-09-14
Resume a job from the SD card after a power cut

f1adang:

Implements the job-resume half of #1745. Two commits, each building on its own.

A job running from the SD card records its state to the card every 10 seconds. `$Job/Resume` then picks it up near where it stopped.

```
$Job/Resume          report what is stored - file, offset, work position, feed, spindle. Moves nothing.
$Job/Resume=go       restore and continue
$Job/Resume/Clear    discard the checkpoint
```

The split is deliberate: resuming drives the tool back to a recorded position, and doing that unannounced after a power cut is how a tool gets broken. The bare command only reports.

## The part worth reviewing carefully

The obvious implementation records `JobSource::position()`. That is wrong, and quietly so.

FluidNC reads ahead: a line is read, queued on `cmd_queue`, parsed, planned, and only then stepped. The reader runs ahead of the tool by that whole pipeline. Checkpointing the read position resumes **past work that was never cut**, leaving a gap in the material.

`plan_block_t::line_number` looked like the answer until I checked how it is populated — it is the N word, which `has_line_number` now says explicitly, and most CAM output carries none. So the first commit carries a byte offset alongside it, by the same route the line travels: `Channel` → `cmd_queue` → parser → planner block. The checkpoint then reads it off `plan_get_current_block()`, so the recorded position and the recorded file offset describe the same instant.

The offset recorded is the **start** of the interrupted line, so a resume re-runs it rather than skipping the part that never got cut. Overlap is recoverable; a gap is not.

## Crash safety

Two copies on the card, written alternately, each with a sequence number and a CRC over its own bytes. A power cut during a write can only damage the copy being written; the other still holds the previous good record, and a torn one is detected rather than resumed from.

The sequence continues from whatever is already on the card. A power cut is exactly when the counter would otherwise restart from zero, below the stored values, and `read()` would hand back the *previous* job as the newer record.

## Ordering that has to be right

Three things, all of which were wrong in my first draft:

- **Positioning moves are emitted as `G21 G90`** whatever the job was using. `mpos_to_wpos()` does no unit conversion, so a recorded position is always millimetres — emitting it under a restored `G20` is out by a factor of 25.4, and under `G91` it is read as a relative move. The job's own modal state is restored afterwards, before the file is handed back to the parser.
- **Feed rate is set while still in `G21`**, so `gc_state` receives the mm/min that was recorded.
- **Nothing retracts Z.** On a machine that cannot home there is no safe height to retract to — machine zero is wherever Z happened to be at boot, possibly down in the work. XY travel happens at whatever height the operator left the tool at, and only then does Z move to the recorded depth.

The moves go through `gc_execute_line()` rather than poking `gc_state`, so soft limits and the parser's own checks still apply to them.

## Homing

This does not assume homing. Work position is recorded as well as machine position, and resume drives to the **work** position: machine position is meaningless after a power cut on a machine that cannot home, while work position is what an operator can re-establish by jogging to the workpiece and re-zeroing. Resume expects that to have been done and says so, rather than pretending to know where the tool is.

On a machine that does home, homing first and then resuming works the same way.

## Scope

Only jobs on the card are checkpointed. A job on the local filesystem is a macro, which nobody resumes partway through, and it would cost a flash erase/write cycle every interval on a small partition shared with `config.yaml` and the WebUI — each one stopping the flash cache on both cores while the step ISR runs.

`sd_mount()` now allows four open files rather than two: a running job holds one descriptor and a checkpoint takes a second while it writes, which left nothing for a concurrent WebUI request.

## Limits, stated plainly

- Up to 10 s of work is re-run. That is the interval, and the overlap is deliberate.
- Granularity is one line of G-code; a job interrupted mid-arc restarts that arc.
- The file must not have changed — the size is compared against the checkpoint and a mismatch is refused rather than seeking into a file that is no longer the same.

## Testing

Checkpoint writing and `$Job/Resume` have been exercised on an ESP32 with jobs run from SD: checkpoints appear and alternate during a job, survive a power cut, and `$Job/Resume=go` restores state and continues the file.

Being straight about the balance: the **writing** side has had a lot of use and several rounds of bugs shaken out of it. The **restore** side has had far less — it works, but it has not run over real material for hours. I would not want it presented as more settled than it is, and the `$Job/Resume` / `=go` split exists partly for that reason.

What this does not cover from #1745 is the UPS half — power-fail input, automatic feed hold, retract and park. That is a separate piece of work and this does not preclude it; it provides the state-saving half it would need.


f1adang:
This one might require some discussions and refining, but I think the feature would be quite beneficial.

Being able to resume jobs massively helps in environments where constant power supply isn't guaranteed.

f1adang:
Before continuing on an implementation actually worthy of making `main`, I'd be thankful for input on how to best approach this, as I have little idea of:
- What kind of machine the majority of FluidNC users are running
- How to best design this to make it useful for a maximum amount of users
- What implications this feature might have on the larger community relying on this project

According to my "research", the current draft of the PR would leave a lot of loose ends an operator would need to be aware / take care of, e.g.:

- Laser/plasma: restoring M3 S… before the Z move means power coming on while positioning. For a laser this needs to be conditional on the spindle type, not unconditional.
- Manual tool change: the tool may not be in the spindle any more. We restore T but can't verify it.
- Z plunge assumes the work hasn't moved. Nothing checks, and nothing can.
- F100 fallback when the recorded feed rate is 0 is arbitrary — fine for a plotter, wrong for most other machines.
And probably another 99 issues this feature could bring when let loose onto the real world.

Also, I'm aware this could have dramatic consequences when used the wrong way on large machines.
Not sure how to best mitigate this except for some warnings in the documentation.

As for how to best integrate this into the workflow: Personally, I'm just calling it as a oneshot command like a macro, but I'm only on a [Brušograf](https://wiki.sgmk-ssam.ch/wiki/Brushograph)
To make this an actually useable feature, one should probably have the option to properly control Resume via the WebUI, showing something along the lines of "Job interrupted at 43% — Resume / Discard?", and maybe ensure some preconditions to prevent machines from destroying themselves or those around them..?

MitchBradley:
I don't like the log-to-SD approach.  It would be better to design a FluidNC board or module with battery backed RAM or backup power that can power the ESP32's internal RTC memory for awhile.

f1adang:
Agreed, but some people/machines/projects just aren't afforded that luxury.

Personally, I'm mostly worried about maxing out cheap SD card's flash write cycles, but I suppose the SD card controller would distribute those ~ 300B writes evenly to minimize wear & tear?
The ESPs Flash could be another option for the checkpoints, but not sure if that would make things any better.

For the use cases of the cult of Brušografy, i.e. multiple-hour-long paint jobs on a  running on whatever power source we can get (including power banks), the ability to resume a long-running job would certainly be desirable.

Personally, I'd be willing to spend significant amounts of time and effort on making this a usable feature for all of the FluidNC community.
However, I'd need some help towards a better design and specification of the feature, and probably a few rounds of going back and forth on implementation.

This being pretty much the last thing our Brušografist hearts desire, it would be an honour to work with you on it.
I'd also try to keep my AI horses in check while we're at it, I promise!

MitchBradley:
Look at https://github.com/bdring/FluidNC/tree/DryRun .  It adds the ability to resume an interrupted job, leveraging the existing SingleBlockMode and CheckMode features.  It adds a breakpoint option to SingleBlockMode, where you can specify one (or a list of) file/line_number pairs for entry to SingleBlockMode, like breakpoints in a debugger.  To resume an interrupted job, you set a breakpoint at the line where you want to resume, turn on CheckMode ($C) and start the job file.  FluidNC then "dry runs" in check mode until the break line is reached.  At that point, the system parameters have values that reflect the machine state that should be established.  FluidNC then runs a restart macro (see examples/restart.nc) that prompts the user to step through (using CycleStart) the stages of restoring the positions, spindle and coolant, before proceeding to run the rest of the job normally.

As for recording the state during the initial run, the idea is to have the sender / UI agent do it.  To that end, DryRun adds a $File/LineNumbers NVS variable that causes Ln: fields to be in the status report, without needing N words in the GCode file.  Browser-based UIs can store the information in a cookie; others can store it in a host file.

PR #1858, opened 2026-09-15
Fix WallPlotter out-of-bounds move on repeated identical G1 target

MitchBradley:
## Summary
- A repeated `G1` with the same target as the previous move takes the WallPlotter kinematics out of bounds.
- Root cause: `WallPlotter::cartesian_to_motors()`'s zero-distance fast path (`FluidNC/src/Kinematics/WallPlotter.cpp`) passed the raw **cartesian** `target` straight to `mc_move_motors()`, which expects **motor-space** (cable-length) coordinates. Every other path in this function converts through `xy_to_lengths()`/`transform_cartesian_to_motors()` first; the zero-distance shortcut skipped that conversion, so cartesian X/Y got sent to the motors as if they were cable lengths.
- Fix: route the zero-distance case through `transform_cartesian_to_motors()` like the rest of the function.

## Test plan
- [ ] User to test on real WallPlotter hardware: send a `G1` to a point, then repeat the identical `G1` line and confirm the machine no longer moves out of bounds.


MitchBradley:
Fixed in c90972b4: the fast path now returns `mc_move_motors()`'s result, and computes cable lengths directly (mirroring the segmented path's `cables[0]`/`cables[1]` mapping) instead of going through `transform_cartesian_to_motors()`, so both paths agree on where cable lengths land regardless of axis mapping.

MitchBradley:
Fixed in f2d9dbb9: the segmented path now writes cable lengths into `cables[_left_axis]`/`cables[_right_axis]` instead of hardcoded `cables[0]`/`cables[1]`, matching `transform_cartesian_to_motors()`'s mapping. The zero-distance fast path now simply calls `transform_cartesian_to_motors()` again, since both paths agree on axis mapping.

MitchBradley:
Pushed b2d69a64: factored the duplicated left/right cable offset math out into a single `WallPlotter::lengths_to_motors()` helper, used by both `transform_cartesian_to_motors()` and the segmented path — the duplication was exactly what let them disagree on axis mapping in the first place.

MitchBradley:
Getting tired of waiting for users to confirm the fix.

PR #1864, opened 2026-09-19
Fix after_reset/after_homing/after_unlock silently killed by stale unwind_cause

### LINKED ISSUE #1861: after_reset/after_homing/after_unlock macros can be silently killed by a stale unwind_cause before running a single line
> *Found and written by **Claude** (Anthropic's AI assistant), at the direction of **MyHomeMyData**, who owns the hardware this was reproduced on. Flagging that up front so it gets human scrutiny rather than being taken on trust — what was observed on hardware vs. what is reasoning from the code is called out explicitly below.*

---

### Symptom

An `after_reset` macro that toggles a `user_outputs` pin (`M62 P3`, turning on a cooling fan for our stepper drivers) reliably does **not** run after a Ctrl-X soft-reset, even though:
- The same command works instantly when typed manually in the console right afterward.
- The debug log shows the macro *starting*, but never executing its one line.

Debug log (`$Message/Level=Debug`) from a soft-reset where the fan stayed off:
```
<Idle|MPos:100.000,100.000,0.000|FS:0,0|Pn:P>
#SOFTRESET#
<Idle|MPos:100.000,100.000,0.000|FS:0,0|Pn:P|WCO:0.000,0.000,-31.108>
Grbl 3.9 [FluidNC v3.9.9 (wifi) '$' for help]
[MSG:DBG: Run after_reset: M62 P3]
<Idle|MPos:100.000,100.000,0.000|FS:0,0|Pn:P|WCO:0.000,0.000,-31.108>
<Idle|MPos:100.000,100.000,0.000|FS:0,0|Pn:P|Ov:100,100,100>
<Idle|MPos:100.000,100.000,0.000|FS:0,0|Pn:P>
```
Note there's no `[MSG:DBG: Macro line: M62 P3]` (compare with a macro triggered via `$Macros/Run=N`, which does log a `Macro line:` entry per line), no `ok`, and no error — the job is created and then vanishes without a trace.

Immediately after, sending `M62 P3` manually from the console works normally (`ok`, fan turns on).

### Root cause (from source, `FluidNC/src/Protocol.cpp`, confirmed present in both `v3.9.9` and current `main`)

`protocol_do_rt_reset()` (the Ctrl-X handler) does two things in sequence:
```cpp
void protocol_do_rt_reset() {
    ...
    protocol_do_late_reset();            // synchronous
    protocol_send_event(&restartEvent);  // only queues the event
}
```
`protocol_do_late_reset()` runs immediately and sets a one-shot abort flag:
```cpp
static void protocol_do_late_reset() {
    ...
    config->_userOutputs->all_off();   // turns the fan off
    sys.abort = true;
    unwind_cause = "Reset";
}
```
`restartEvent` (→ `protocol_do_soft_restart()` → `after_reset` macro → `Macro::run()` → `Job::nest()`) only runs later, once the queued event is dequeued in `protocol_handle_events()`.

Meanwhile, a separate FreeRTOS task (`polling_loop`) continuously does:
```cpp
if (!Job::active()) {
    unwind_cause = nullptr;   // cleared only when no job is active
    ...
} else {
    if (unwind_cause) {
        Job::abort();          // kills whatever job is currently on the stack
        unwind_cause = nullptr;
        continue;
    }
    ...
}
```

If `polling_loop` doesn't get a chance to clear `unwind_cause` (via the `!Job::active()` branch) *before* the queued `restartEvent` is processed and `after_reset` pushes its own job with `Job::nest()`, then the next `polling_loop` iteration sees `Job::active() == true` **and** a stale `unwind_cause` — and aborts the brand-new `after_reset` job before it ever calls `pollLine()`. The abort flag meant for whatever was running *before* the reset ends up killing a job that didn't exist yet when the flag was set.

This is a race between two concurrent tasks over shared state (`unwind_cause`, the `Job` stack) with no synchronization tying the flag's clearing to the specific job it was meant for.

### Scope

Likely affects `after_reset`, `after_homing`, and `after_unlock` equally, since all three run through the same `Macro::run()` → `Job::nest()` path shortly after a reset/homing/unlock event. `startup_line0`/`startup_line1` appear unaffected in our testing (different code path, `protocol_do_start()`, not reached via this reset flow).

### Possible fix direction

Not proposing a specific patch, but the underlying issue seems to be that `unwind_cause` isn't consumed atomically with respect to job creation. One direction: have `Job::nest()` itself clear `unwind_cause` when pushing a *new* job, since a job that didn't exist when the abort signal was raised shouldn't inherit it.

### Environment

FluidNC 3.9.9 (wifi), Jackpot3 (ESP32-WROOM-32). Confirmed the same code path is unchanged on current `main`.

MitchBradley:
## Summary

A Ctrl-X reset sets `unwind_cause = "Reset"` synchronously in `protocol_do_late_reset()`, then only later queues the `restartEvent` that runs `after_reset` via `Job::nest()`. `polling_loop()` clears `unwind_cause` when idle (`!Job::active()`), but if the queued `after_reset` macro nests its job before the next poll pass gets there, that pass sees `Job::active() == true` with the reset's own stale `unwind_cause` still set, and aborts the brand-new job before it runs a single line. The same path affects `after_homing`/`after_unlock`.

- `Job::nest()` now clears `unwind_cause` whenever it starts a fresh job stack (the stack was empty before this push): a job that didn't exist when the abort flag was raised shouldn't inherit it. A nested push (stack already non-empty) leaves the flag alone, since it belongs to the still-running job being nested into.
- `unwind_cause` is now declared `extern` in `Protocol.h` so `Job.cpp` can see it (it was previously only visible inside `Protocol.cpp`).

Fixes #1861

## Test plan
- [x] `pio run -e wifi` builds clean
- [ ] Hardware confirmation that `after_reset` (e.g. `M62 P3`) now reliably runs after Ctrl-X, per the reporter's repro


MitchBradley:
Good catch — pushed d56f1206 to close this gap. `protocol_do_late_reset()` now calls `Job::abort()` synchronously, right there on the protocol_main_loop task, before `restartEvent` is even queued. That guarantees the job stack is empty by the time `after_reset` nests, regardless of whether `polling_loop` (a separate task) has gotten a chance to notice `unwind_cause` yet — so the "reset during an active job" case you flagged is now handled the same as the idle case.

MitchBradley:
Also valid — pushed 83687d86. The check-then-abort and nest()'s clear-on-fresh-stack were two separate critical sections, so the polling task could read a stale \`unwind_cause\` in the gap between them and abort a job that \`nest()\` had already disclaimed.

Added \`Job::consumeUnwindCause()\`, which checks the flag against the job stack and aborts if warranted, all inside the *same* job-mutex critical section \`nest()\` uses to clear it. \`poll_once()\` now calls that instead of touching the raw flag. The two paths can no longer observe each other mid-transition since only one holds the mutex at a time.

MitchBradley:
Fixed in 1241830e — renamed to \`consume_unwind_cause\` to match the file's snake_case convention.

PR #1866, opened 2026-09-19
Block flash writes and other job-unsafe commands while a job is active

MitchBradley:
## Summary

`notIdleOrAlarm()` (the default state guard for many `$`/`[ESPxxx]` commands) isn't a sufficient guard on its own for commands that write NVS/flash or are otherwise unsafe while a job's motion or control flow is still live: `State::Idle` is reachable while `Job::active()` is still true — `protocol_do_cycle_start()` sets `State::Idle` whenever a resume finds nothing queued in the planner, which happens between motion segments or during a job's non-motion lines. A command gated only on state can slip through in that window.

Added `notIdleOrAlarmOrJobActive()` (`notIdleOrAlarm() || Job::active()`) and applied it in place of `notIdleOrAlarm()` for:

- **`WebUI/Set` (ESP401)**: reaches NVS writes, and is a WebUI-only hook never used from macros or other automation — direct `$name=value` lines are deliberately left alone, since those don't write flash and are a normal way to tweak a live setting mid-job.
- **`Settings/Erase` (`$NVX`), `Settings/Restore` (`$RST`)**: NVS writes.
- **`Sta/Setup` (ESP103), `Ethernet/Setup`**: NVS writes, plus risk of dropping the connection driving the job.
- **`Notification/Setup` (ESP610)**: NVS writes.
- **`Motor/Disable` (`$MD`), `System/Sleep` (`$SLP`)**: not flash writes, but lose step position if a job's motion is live.
- **`Uart/Passthrough` (`$UP`)**: may repurpose a UART a job's spindle is using.
- **`SD/Delete`, `LocalFS/Format`**: could delete/wipe the file or filesystem a job is currently reading from.
- **`SD/Rename`**: could rename the file a job is currently reading from.

`Motor/Enable` (`$ME`) and `Motors/Init` (`$MI`) were considered and left on `notIdleOrAlarm`.

Deliberately out of scope: `$RM` (`Macros/Run`) has no `cmdChecker` at all today (`nullptr`) and shares the same "asynchronous trigger nests into a live job stack" hazard already fixed for `macro0`-`macro3` pins in [#1863](https://github.com/bdring/FluidNC/pull/1863) — left for a separate follow-up.

## Test plan

- [x] `pio run -e macos` builds clean, no warnings
- [ ] Manual verification on hardware that each gated command is now rejected mid-job and still works when idle


MitchBradley:
Withdrawing this one. The Copilot findings on this PR turned out to be correct but pointed at a much bigger surface than the original small gap (only setWebSetting truly needed this): closing it properly means also touching Setting::check_state() (which every legacy NVS-backed Setting object goes through, independent of how it's invoked), several more filesystem mutators, and ESP555, with the accompanying risk of breaking legitimate nested-in-job use (as the $RM discussion here surfaced) somewhere in that wider blast radius. The fix ended up riskier than the narrow gap it was closing. Not pursuing further for now.

